PR #91, opened 2015-01-21
Disable cipher that is not in boringssl (AES-192 counter mode).

thughes:
This change makes libsrtp work with boringssl (fork of openssl: https://boringssl.googlesource.com/boringssl), which doesn't have AES-192 counter mode. The configure part needs some work (right now it just `#define`'s `BORINGSSL` if you enable openssl), but I wasn't sure how you'd like that done.

### DIFF
--- a/config_in.h
+++ b/config_in.h
@@ -1,5 +1,8 @@
 /* config_in.h.  Generated from configure.in by autoheader.  */
 
+/* Define this to enable compatibility with BoringSSL. */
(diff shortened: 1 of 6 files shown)


fancycode:
You can take a look at #93 which uses the `#define` provided by BoringSSL to check for it.

jfigus:
Closing this request since it's a duplicate of 93.

PR #93, opened 2015-02-18
Don't include AES-192 when compiling against BoringSSL.

fancycode:
libsrtp is used for WebRTC in Chromium. There are currently plans to update to newer versions of libsrtp and compile against OpenSSL (which is actually BoringSSL for Chromium). As BoringSSL doesn't support AES-192, this cipher should not be included if libsrtp is compiled against it.

This patch implements the check, so any updates to newer upstream versions of libsrtp in Chromium don't need further changes.

### DIFF
--- a/crypto/cipher/aes_icm_ossl.c
+++ b/crypto/cipher/aes_icm_ossl.c
@@ -65,7 +65,7 @@ debug_module_t mod_aes_icm = {
 };
 extern cipher_test_case_t aes_icm_test_case_0;
 extern cipher_type_t aes_icm;
-#ifndef OPENSSL_IS_BORINGSSL
+#ifndef SRTP_NO_AES192
 extern cipher_type_t aes_icm_192;
 #endif
 extern cipher_type_t aes_icm_256;
@@ -124,7 +124,7 @@ err_status_t aes_icm_openssl_alloc (cipher_t **c, int key_len, int tlen)
      * Verify the key_len is valid for one of: AES-128/192/256
      */
     if (key_len != AES_128_KEYSIZE_WSALT &&
-#ifndef OPENSSL_IS_BORINGSSL
+#ifndef SRTP_NO_AES192
         key_len != AES_192_KEYSIZE_WSALT &&
 #endif
         key_len != AES_256_KEYSIZE_WSALT) {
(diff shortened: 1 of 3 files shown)


fancycode:
Whoops, just noticed this was already reported in #91. This merge request doesn't need any changes to the `configure` script because it uses a `#define` already provided by BoringSSL.

jfigus:
It would be better if the new #define was named something like SRTP_NO_AES192.  Then this feature isn't tied to the boringssl project.  Also, there are other references to AES-192 in the library, such as AES_192_ICM.  What happens if a user configures a new SRTP session using AES_192_ICM?  My guess is the srtp layer will try to use the 192 cipher and something ugly will occur, possibly a crash.  Perhaps we need to add some defensive code in srtp.c?

fancycode:
If it was `SRTP_NO_AES192`, I agree there should probably be more checks - however I didn't verify what happens in the code in these cases.
For WebRTC in Chromium (and thus BoringSSL), AES-192 is never requested, so just disabling it with `OPENSSL_IS_BORINGSSL` seemed fine for me as a first step.

jfigus:
IMHO, this is a poor precedent to set.  That is, adding preprocessor directives based on the needs of downstream applications.  Support for OpenSSL was added a couple of years ago.  BoringSSL was never addressed.  Hence, this pull request is a new feature request.  Having said that, it seems reasonable to accept this pull request.  But it's a poor precedent to start littering the code with preprocessor directives named from upstream/downstream projects.

fancycode:
I did some review of the code regarding creation/usage of ciphers in `srtp.c` and checked what will happen if AES-192 is not available:
- Ciphers are created by [`crypto_kernel_alloc_cipher`](https://github.com/cisco/libsrtp/blob/557a7876b35260817a584434f7bc842f72a106d6/crypto/kernel/crypto_kernel.c#L482).
- If the `cipher_type_id_t` for AES-192 is passed, the corresponding `cipher_type_t *` doesn't exist, so it will fail directly with `err_status_fail`.
- If the generic id `AES_ICM` is passed with a keylength that would result in AES-192, it will fail in (the patched) `aes_icm_openssl_alloc` due to an unsupported keylength.
- All places that call `crypto_kernel_alloc_cipher` (only called from `srtp.c`) check the return values, and report the failure to the caller.

So IMHO no additional checks are required because they already exist where needed - unless I miss something.

If you would rather have the `#define` named differently, I can change it to `SRTP_NO_AES192` and automatically set that if `OPENSSL_IS_BORINGSSL` is set.

What do you think?

jfigus:
Thanks for testing this.  Yes, please rename the #define and I'll accept the request.

fancycode:
Pull request is updated.

fancycode:
I noticed you just released 1.5.1. Do you have an ETA when a release with this change will be ready? I'm currently trying to get a new version of libsrtp in Chromium (which still is using 1.4.4) and obviously would like to use the latest version without any manual patching ;-)

jfigus:
How about if we let it sit for a week, then do 1.5.2?  Then I'll plan to collapse the 2_0_0_dev branch into master.

fancycode:
Sounds good, thanks.

PR #105, opened 2015-04-15
srtp_remove_stream fix: convert SSRC to network order before comparing

ukreator:
srtp_remove_stream() always returned error err_status_no_ctx for us. The problem is in comparison logic. SSRC is saved in network byte order internally, but is being compared against SSRC passed in host byte order.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -1973,6 +1973,9 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
   /* sanity check arguments */
   if (session == NULL)
     return srtp_err_status_bad_param;
+
+  /* will be compared against values in network order in the stream list */
+  ssrc = htonl(ssrc);
   
   /* find stream in list; complain if not found */
   last_stream = stream = session->stream_list;


JonathanLennox:
This would be an API change; existing code calls htonl on the argument it passes to srtp_remove_stream, for exactly this reason.

So it's reasonable to fix this for 2.0, but changing it silently for 1.x seems like a bad idea.

jfigus:
This pull request is for master, which currently has 2.0 code.  But even if we put this into 2.0, it's going to create a lot of confusion for the users.

ukreator:
It's currently quite confusing. I don't think any new user of this API function expects that SSRC must be in network byte order while other methods accept host byte order for SSRC. Maybe it makes sense to rename srtp_remove_stream in 2.0 to make sure old users (I doubt there's a lot of them) don't accidentally make their programs buggy after upgrading to 2.0

ibc:
@jfigus may you please reopen this issue and add documentation (within srtp.h) regarding which functions require a network byte order SSRC and which not? I also find this very confusing, so documentation would really help.

jfigus:
I've updated the documentation in 68f8a888398451a3ed4ff6c5b11af2fd53f4295e

ibc:
Thanks. May you please also clarify how to read the `value` field within the `srtp_ssrc_t` struct?:

```c
/**
 * @brief An srtp_ssrc_t represents a particular SSRC value, or a `wildcard' SSRC.
 * 
 * An srtp_ssrc_t represents a particular SSRC value (if its type is
 * ssrc_specific), or a wildcard SSRC value that will match all
 * outbound SSRCs (if its type is ssrc_any_outbound) or all inbound
 * SSRCs (if its type is ssrc_any_inbound).  
 *
 */

typedef struct { 
  srtp_ssrc_type_t type;  /**< The type of this particular SSRC */
  unsigned int     value; /**< The value of this SSRC, if it is not a wildcard */
} srtp_ssrc_t;
```

PR #115, opened 2015-09-17
Perform const-correctness to global data, test cases, "virtual tables", discriptions, and pointers to them.

persmule:
Many global data (e.g. tables for aes), test case objects and their data, "virtual tables" (srtp_cipher_type_t and srtp_auth_type_t, since they contain many function pointers able to override, just as virtual tables in C++), descriptions do and should not be changed during their lifetime, so I believe it is better to qualify them, as well as any pointers to them read-only (const). If so, compilers can put them into rodata segments and prevent any writing attempt to them. Anyone wants to override default functions should never modify those existing global "virtual tables", but make their own copies of the original, and adjust pointers to them.

### DIFF
--- a/crypto/cipher/aes.c
+++ b/crypto/cipher/aes.c
@@ -64,7 +64,7 @@
 
 #ifndef WORDS_BIGENDIAN
 
-static uint32_t T0[256] = {
+static const uint32_t T0[256] = {
     0xa56363c6, 0x847c7cf8,  0x997777ee,  0x8d7b7bf6,
     0xdf2f2ff,  0xbd6b6bd6,  0xb16f6fde,  0x54c5c591,
     0x50303060, 0x3010102,   0xa96767ce,  0x7d2b2b56,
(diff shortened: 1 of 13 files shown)


jfigus:
This looks good.  Let's give it a few days to see if anyone else comments, then we can pull your commits.

jfigus:
After merging this pull request, there was a seg fault when running the unit tests.  I don't have time to look into the root cause today, so I reverted the commit.  Sorry.

persmule:
On what unit test the merged library crashed? Maybe I could help to check it.

jfigus:
If you do a "make runtest", it'll run all the unit tests.

persmule:
Those test cases works fine with trivia configured ($ ./configure )
codes of my repository. Which case will crash on which configure
options? Please tell me. I will manage to reproduce them.
在 2015年10月24日 01:10, John Foley 写道:

jfigus:
I don't recall which one crashed now, it's been over a month.  I just did the normal process:
./configure
make
make runtest
make clean
./configure --enable-openssl
make 
make runtest

What's trivia?

jfigus:
Actually, the original build failure from your pull request can be found here:
https://openssl-sanity.cisco.com:8443/view/libsrtp/job/srtp_master_ossl/237/console

persmule:
The variables related to iv should never be read-only, since operations
within aes_gcm_openssl_set_iv require iv to be writable.

I have made a new pull request, and this time I performed unit tests
successfully on both --enable-openssl and --disable-openssl configurations.



在 2015年10月24日 02:12, John Foley 写道:

PR #116, opened 2015-10-07
Prevent buffer overflow when clearing srtp_kdf_t

pabuhler:
Use size macros that where used in the declaration.


jfigus:
Thanks for the contribution.

PR #117, opened 2015-10-15
Prevent stack-buffer-overflow with gcc -fsanitize=address

pabuhler:
The value passed to srtp_aes_gcm_openssl_set_aad is not guarantied
to be >= c->tag_len. Since the call to EVP_CIPHER_CTX_ctrl is a
dummy call pass it a dummy tag that is large enough.

### DIFF
--- a/crypto/cipher/aes_gcm_ossl.c
+++ b/crypto/cipher/aes_gcm_ossl.c
@@ -255,7 +255,8 @@ static srtp_err_status_t srtp_aes_gcm_openssl_set_aad (srtp_aes_gcm_ctx_t *c, ui
      * Set dummy tag, OpenSSL requires the Tag to be set before
      * processing AAD
      */
-    EVP_CIPHER_CTX_ctrl(&c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len, aad);
+    unsigned char dummy_tag[GCM_AUTH_TAG_LEN];
+    EVP_CIPHER_CTX_ctrl(&c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len, &dummy_tag);
 
     rv = EVP_Cipher(&c->ctx, NULL, aad, aad_len);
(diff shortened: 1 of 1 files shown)


jfigus:
I'm not convinced this is needed.  What's the value in passing uninitialized data on the stack as compared to whatever aad points too?

jesup:
> I'm not convinced this is needed. What's the value in passing uninitialized data on the stack as compared to whatever aad points too?

That's (I assume) the point here; aad may not be larger than the tag_len.  This patch avoids reading/writing past the end of the allocation.  If it actually reads first, then it might be good to initialize the buffer to 0's, but I suspect (without checking) it doesn't.

pabuhler:
To be honest I did not find any code that wrote to the address, only that read from it. There is a memcpy in openssl from the address passed in to an internal variable that holds the tag. This read is of tag len which was larger than the size of the variable on the stack.  So although the read is “out of bounds” it is most likely harmless.
I was also thinking about initializing it but since it is not used for anything then I did not wish add more code for no reason.

pascal

jesup:
It's important for security reasons to have no stack garbage exposed, and the enable tools like address sanitizer to run without finding problems.  For example, no "out of bounds" read is safely harmless.  You don't know if the caller's buffer is at the very end of a page (MMU crash on OOB read).  Or if the next block of memory has credit card info.  Or an SSH private key (e.g. Heartbleed, etc).  The same applies to reading stack garbage.

So PLEASE use the temp buffer (per the patch), and PLEASE zero it.  Downstream people will be very happy you did so, so they don't have to carry patches against SRTP in their pulls.  Thanks.

jfigus:
Tried to merge this pull request today, but your branch is stale.  Can you sync the latest from master and resubmit?

pabuhler:
sure

pabuhler:
re submitted as https://github.com/cisco/libsrtp/pull/134

PR #121, opened 2015-10-23
Perform const-correctness to global data, test cases, "virtual tables", discriptions, and pointers to them.

persmule:
Many global data (e.g. tables for aes), test case objects and their data, "virtual tables" (srtp_cipher_type_t and srtp_auth_type_t, since they contain many function pointers able to override, just as virtual tables in C++), descriptions do and should not be changed during their lifetime, so I believe it is better to qualify them, as well as any pointers to them read-only (const). If so, compilers can put them into rodata segments and prevent any writing attempt to them. Anyone wants to override default functions should never modify those existing global "virtual tables", but make their own copies of the original, and adjust pointers to them.

Unlike the last pullreq, the variables related to iv are never read-only now, since operations within aes_gcm_openssl_set_iv require iv to be writable.

This time I performed unit tests successfully on both --enable-openssl and --disable-openssl configurations.


jfigus:
Giving it another whirl.  We'll see if Mr. Jenkins is happy with this attempt.

PR #123, opened 2015-11-06
Add srtp_update functions

pabuhler:
The srtp_update and srtp_update_stream functions are provided
as a way to rekey or update the policy of an existing stream while
preserving the ROC as required by
https://tools.ietf.org/html/rfc3711#section-3.3.1 .

The current implementation only preserves the ROC. It does not
preserve the replay database for RTP or RTCP, that is a possible
improvement.

I also have same patch for 1_5_X branch if interested.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -2065,8 +2065,9 @@ srtp_update(srtp_t session, const srtp_policy_t *policy) {
   srtp_err_status_t stat;
 
   /* sanity check arguments */
-  if ((session == NULL) || (policy == NULL) || (policy->key == NULL))
+  if ((session == NULL) || (policy == NULL) || (policy->key == NULL)) {
     return srtp_err_status_bad_param;
+  }
 
   while (policy != NULL) {
     stat = srtp_update_stream(session, policy);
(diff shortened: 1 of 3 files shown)


srtp/srtp.c
```diff
@@ -2074,6 +2060,154 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
...
+srtp_err_status_t
+srtp_update(srtp_t session, const srtp_policy_t *policy) {
+  srtp_err_status_t stat;
+
+  /* sanity check arguments */
+  if ((session == NULL) || (policy == NULL) || (policy->key == NULL))
+    return srtp_err_status_bad_param;
+
+  while (policy != NULL) {
+    stat = srtp_update_stream(session, policy);
```
jfigus: If we have multiple policies, isn't this logic simply going to apply the last policy in the list?  Why go through all the iterations of every policy in the list?

srtp/srtp.c
```diff
@@ -2074,6 +2060,154 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
...
+      srtp_xtd_seq_num_t old_index = stream->rtp_rdbx.index;
+
+      /* remove stream */
+      status = srtp_remove_stream(session, ssrc);
+      if (status) {
+        /* free new allocations */
+        while (new_stream_list != NULL) {
+          srtp_stream_t next = new_stream_list->next;
+          status = srtp_stream_dealloc(new_stream_list, new_stream_template);
+          if (status)
```
jfigus: You should include braces around the conditional block of code for any if statement.  The exception to this rule is when it can all be contained on a single line.  This is a coding standard that has been in place at Cisco for a long time.

srtp/srtp.c
```diff
@@ -2074,6 +2060,154 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
...
+
+    /* set policy to next item in list  */
+    policy = policy->next;
+  }
+  return srtp_err_status_ok;
+}
+
+
+srtp_err_status_t
+srtp_update_stream(srtp_t session, const srtp_policy_t *policy) {
```
jfigus: Can this be declared static and removed from srtp.h?  I don't see where the unit test code uses this entry point.

srtp/srtp.c
```diff
@@ -2074,6 +2060,154 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
...
+    srtp_stream_dealloc(session->stream_template, NULL);
+    /* set new template */
+    session->stream_template = new_stream_template;
+    /* add new list */
+    if (new_stream_list) {
+      new_stream_list->next = session->stream_list;
+      session->stream_list = new_stream_list;
+    }
+    break;
+  }
```
jfigus: It's not common to use braces around a case block.  Can you remove these?

srtp/srtp.c
```diff
@@ -2074,6 +2060,154 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
...
+    return srtp_err_status_bad_param;
+
+  switch (policy->ssrc.type) {
+  case (ssrc_any_outbound):
+  case (ssrc_any_inbound): {
+    if (session->stream_template == NULL) {
+      return srtp_err_status_bad_param;
+    }
+
+    srtp_stream_t new_stream_template;
```
jfigus: All variables should be declared at the top of the function.  This can create problems from some compilers (e.g. Microsoft)

srtp/srtp.c
```diff
@@ -2074,6 +2060,154 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
...
+
+    /* for all old templated streams */
+    for (;;) {
+      srtp_stream_t stream = session->stream_list;
+      while ((stream != NULL) && (stream->rtp_auth != session->stream_template->rtp_auth)) {
+        stream = stream->next;
+      }
+      if (stream == NULL) {
+        /* no more templated streams */
+        break;
```
jfigus: Have you checked this code for memory leaks?  Shouldn't we free new_stream_template here?

srtp/srtp.c
pabuhler: My preference is to add braces always, but I was trying hard to match the style in the rest of the file.
I will add braces every where.

srtp/srtp.c
pabuhler: srtp_update_stream is called from srtp_update which is called in the test and therefore this function is tested.
As to if it should be public or not is debatable, I was trying once again to follow the rest of the srtp api that has both the create function where a list of policies can be used or else the add_stream function that operates on just one policy. Could make it so you can only update one stream at a time therefore only having one function but it is not consistent with rest of api.

srtp/srtp.c
pabuhler: line 2078 has "policy = policy->next;"
This should loop though all, do you disagree?

srtp/srtp.c
pabuhler: sure

srtp/srtp.c
pabuhler: sure

srtp/srtp.c
pabuhler: breaking here is not an error, new_stream_template  will be used after this for loop so not sure why you think it would leak. line 2170.

All though if srtp_stream_dealloc fails on line 2136 and line 2151 fail then it will be leaked, but at that point there could be many leaks so not sure. I saw there was comment some where that if deallocating fails then memory will leaked.

srtp/srtp.c
```diff
@@ -2074,6 +2060,154 @@ srtp_remove_stream(srtp_t session, uint32_t ssrc) {
...
+        return status;
+      }
+
+      /* allocate and initialize a new stream */
+      status = srtp_stream_clone(new_stream_template, ssrc, &stream);
+      if (status) {
+        /* free new allocations */
+        while (new_stream_list != NULL) {
+          srtp_stream_t next = new_stream_list->next;
+          status = srtp_stream_dealloc(new_stream_list, new_stream_template);
```
pabuhler: perhaps ignore this status and continue trying to free all new stream and return original error?

srtp/srtp.c
jfigus: Yes, it'll apply all the policies to the session.  But isn't only one policy applicable to the session?  So the last policy to be applied would be in effect.  What's the point of applying all the other policies to the session?

pabuhler: This may be me not understanding the api but I had thought that a session can have multiple streams and that a policy could either apply to a specific stream based on SSRC or to any out bound / inbound stream.
So in this case the policy list will be iterated and each policy will be passed to srtp_update_stream, where that policy will either match a given ssrc and so update that or match the template policy and then update all streams that are associated with that template.
Why anyone would use a list of policies is beyond me but I assume some one must as that is how srtp_create is implemented…

jfigus: OK, I see what you're doing now.  Thanks for clearing up my confusion.

jfigus:
I'm seeing the following warning with your commit:

srtp/srtp.c: In function ‘srtp_update_stream’:
srtp/srtp.c:2214:21: warning: variable ‘status’ set but not used [-Wunused-but-set-variable]
   srtp_err_status_t status;
                     ^
Looking at the code, I'm not sure the error handling is correct.  The following patch should resolve the problem.  Do you concur?

diff --git a/srtp/srtp.c b/srtp/srtp.c
index db7bdd7..dd3e1b5 100644
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -2205,7 +2205,7 @@ update_stream(srtp_t session, const srtp_policy_t *policy) {
   /* restore old extended seq */
   stream->rtp_rdbx.index = old_index;
 
-  return status;
+  return srtp_err_status_ok;
 }
 
 
@@ -2230,7 +2230,7 @@ srtp_update_stream(srtp_t session, const srtp_policy_t *policy) {
   }
 
-  return srtp_err_status_ok;
+  return status;
 }

Issue #124, opened 2015-11-11
Add support for RFC 6904 encrypted header extensions


juberti:
Extend srtp_protect/unprotect to encrypt header extensions according to RFC 6904.

The extensions to be encrypted would be set by a new API, e.g. 
srtp_set_protect_extns(srtp_t session, const int* extn_ids, int extn_ids_len);

jfigus:
Which downstream project requires this?

JonathanLennox:
It's required by the RTCWeb RTP usage spec, so anything that's using LibSRTP for WebRTC.

jnoring:
https://github.com/ging/licode would be impacted, which is one of the most commonly used WebRTC media servers.

fancycode:
This can be closed (implemented in #128).

PR #128, opened 2015-11-16
Support header extensions encryption (RFC 6904).

fancycode:
Header ids to encrypt can be specified through the "srtp_policy_t".
Added validation testcase using test vectors from RFC. Implements issue #124.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -1102,6 +1102,9 @@ srtp_process_header_encryption(srtp_stream_ctx_t *stream, srtp_hdr_xtnd_t *xtn_h
       uint32_t xlen_with_header = 1+xlen;
       xtn_hdr_data++;
 
+      if (xtn_hdr_data + xlen > xtn_hdr_end)
+        return srtp_err_status_parse_err;
+
       if (xid == 15) {
         /* found header 15, stop further processing. */
         break;
@@ -1134,6 +1137,9 @@ srtp_process_header_encryption(srtp_stream_ctx_t *stream, srtp_hdr_xtnd_t *xtn_h
       unsigned int xlen = *(xtn_hdr_data+1);
       uint32_t xlen_with_header = 2+xlen;
       xtn_hdr_data += 2;
+
+      if (xtn_hdr_data + xlen > xtn_hdr_end)
+        return srtp_err_status_parse_err;
 
       status = srtp_cipher_output(stream->rtp_xtn_hdr_cipher, keystream, &xlen_with_header);
       if (status)
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -233,6 +233,41 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
    return stat;    
   }
 
+  if (p->enc_xtn_hdr && p->enc_xtn_hdr_count > 0) {
+    str->enc_xtn_hdr = (int*) srtp_crypto_alloc(p->enc_xtn_hdr_count * sizeof(p->enc_xtn_hdr[0]));
+    if (!str->enc_xtn_hdr) {
+      auth_dealloc(str->rtcp_auth);
+      srtp_cipher_dealloc(str->rtcp_cipher);
+      auth_dealloc(str->rtp_auth);
+      srtp_cipher_dealloc(str->rtp_cipher);
+      srtp_crypto_free(str->limit);
+      srtp_crypto_free(str);
+      return srtp_err_status_alloc_fail;
+    }
+    memcpy(str->enc_xtn_hdr, p->enc_xtn_hdr, p->enc_xtn_hdr_count * sizeof(p->enc_xtn_hdr[0]));
+    str->enc_xtn_hdr_count = p->enc_xtn_hdr_count;
+
+    /* allocate cipher for extensions header encryption */
+    stat = srtp_crypto_kernel_alloc_cipher(p->rtp.cipher_type,
+              &str->rtp_xtn_hdr_cipher,
+              p->rtp.cipher_key_len,
+              p->rtp.auth_tag_len);
+    if (stat) {
+      srtp_crypto_free(str->enc_xtn_hdr);
+      auth_dealloc(str->rtcp_auth);
+      srtp_cipher_dealloc(str->rtcp_cipher);
+      auth_dealloc(str->rtp_auth);
+      srtp_cipher_dealloc(str->rtp_cipher);
+      srtp_crypto_free(str->limit);
+      srtp_crypto_free(str);
+      return stat;
+    }
+  } else {
+    str->rtp_xtn_hdr_cipher = NULL;
+    str->enc_xtn_hdr = NULL;
+    str->enc_xtn_hdr_count = 0;
+  }
+
   return srtp_err_status_ok;
 }
 
@@ -274,6 +309,15 @@ srtp_stream_dealloc(srtp_t session, srtp_stream_ctx_t *stream) {
     srtp_crypto_free(stream->limit);
   }   
 
+  if (session->stream_template
+      && stream->rtp_xtn_hdr_cipher == session->stream_template->rtp_xtn_hdr_cipher) {
+    /* do nothing */
+  } else if (stream->rtp_xtn_hdr_cipher) {
+    status = srtp_cipher_dealloc(stream->rtp_xtn_hdr_cipher);
+    if (status)
+      return status;
+  }
+
   /* 
    * deallocate rtcp cipher, if it is not the same as that in
    * template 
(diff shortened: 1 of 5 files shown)


juberti:
I took a quick pass over this and this looks solid. However, I'm not a committer on this repo, so I leave it to one of @davidmcgrew @jfigus @JonathanLennox to give this an official thumbs up.

fancycode:
@jfigus did you get a chance to look at this yet?

JonathanLennox:
I don't think this does the right thing for the AES_GCM policy.  GCM is supposed to use ICM for the header extension keystream.

fancycode:
Thanks for your feedback. Currently this only encrypts header extensions for the non-GCM cipher suites (i.e. not in `srtp_protect_aead` / `srtp_unprotect_aead`), I wanted to get some initial feedback first.

Do you know if there are some test vectors for the AES_GCM policy I could verify against?

JonathanLennox:
Even if AES_GCM is being used for payload protection, you still use AES_ICM for the header extension keystream.  See draft-ietf-avtcore-srtp-aes-gcm-17 section 8.3.  So the same test vectors still apply.

fancycode:
Latest commits also support header extensions encryption if packets are encrypted using GCM ciphers as defined in draft-ietf-avtcore-srtp-aes-gcm-17 section 8.3.

fancycode:
Do you see any other issues that need to be changed before this can be merged?

JonathanLennox:
Your second commit changed an #ifndef OPENSSL to #ifdef OPENSSL in the driver, when I think you mean to remove the #ifdef entirely, no?

fancycode:
Good catch, thanks! This was probably some leftover from my testing. I double checked with and without OpenSSL and found another case where it didn't link the test without OpenSSL.

srtp/srtp.c
```diff
@@ -233,6 +233,60 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
...
+      return srtp_err_status_alloc_fail;
+    }
+    memcpy(str->enc_xtn_hdr, p->enc_xtn_hdr, p->enc_xtn_hdr_count * sizeof(p->enc_xtn_hdr[0]));
+    str->enc_xtn_hdr_count = p->enc_xtn_hdr_count;
+
+    /* For GCM ciphers, the corresponding ICM cipher is used for header extensions encryption. */
+    switch (p->rtp.cipher_type) {
+    case SRTP_AES_128_GCM:
+      enc_xtn_hdr_cipher_type = SRTP_AES_128_ICM;
+      enc_xtn_hdr_cipher_key_len = 30;
```
jfigus: Should we use (SRTP_SALT_SIZE + SRTP_AES_128_KEYSIZE) instead of 30?

srtp/srtp.c
```diff
@@ -233,6 +233,60 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
...
+      break;
+    case SRTP_AES_256_GCM:
+      enc_xtn_hdr_cipher_type = SRTP_AES_256_ICM;
+      enc_xtn_hdr_cipher_key_len = 46;
```
jfigus: Should we use (SRTP_SALT_SIZE + SRTP_AES_256_KEYSIZE) instead of 30?

srtp/srtp.c
fancycode: Both constants are in `crypto/include/aes_icm_ossl.h` which is only pulled in to this file if `OPENSSL` and `OPENSSL_KDF` are defined.

srtp/srtp.c
fancycode: Both constants are in `crypto/include/aes_icm_ossl.h` which is only pulled in to this file if `OPENSSL` and `OPENSSL_KDF` are defined.

srtp/srtp.c
```diff
@@ -903,6 +1057,113 @@ srtp_stream_init(srtp_stream_ctx_t *srtp,
...
+  int keystream_pos;
+  uint8_t* xtn_hdr_data = ((uint8_t *)xtn_hdr) + octets_in_rtp_extn_hdr;
+  uint8_t* xtn_hdr_end = xtn_hdr_data + (ntohs(xtn_hdr->length) * sizeof(uint32_t));
+
+  if (ntohs(xtn_hdr->profile_specific) == 0xbede) {
+    /* RFC 5285, section 4.2. One-Byte Header */
+    while (xtn_hdr_data < xtn_hdr_end) {
+      uint8_t xid = (*xtn_hdr_data & 0xf0) >> 4;
+      unsigned int xlen = (*xtn_hdr_data & 0x0f) + 1;
+      uint32_t xlen_with_header = 1+xlen;
```
jfigus: This code is challenging to follow.  I will admit to not being familiar with RTP header extensions.  My concern here is a possible DoS attack.  We're trusting the header length is correct.  What if an attacker provides a malformed packet with an oversized header extension length?  The encryption routine will potentially be invoked on data beyond the packet buffer.  Do we need to add some validation logic to the length calculation on each header extension?

srtp/srtp.c
```diff
@@ -1149,7 +1439,7 @@ srtp_unprotect_aead (srtp_ctx_t *ctx, srtp_stream_ctx_t *stream, int delta,
      */
     enc_start = (uint32_t*)hdr + uint32s_in_rtp_header + hdr->cc;
     if (hdr->x == 1) {
-        srtp_hdr_xtnd_t *xtn_hdr = (srtp_hdr_xtnd_t*)enc_start;
+        xtn_hdr = (srtp_hdr_xtnd_t*)enc_start;
         enc_start += (ntohs(xtn_hdr->length) + 1);
```
jfigus: I'm worrying about a buffer overflow again here.  Wondering if we need to add some logic to check the packet length from the UDP layer (*pkt_octet_len) against the implied length we derive based on uint32s_in_rtp_header, hdr->cc, and now xtn_hdr->length.

srtp/srtp.c
fancycode: The length is already checked by the callers of this function (for example here: https://github.com/fancycode/libsrtp/blob/header_extensions_encryption/srtp/srtp.c#L1685), this has been checked my change.

srtp/srtp.c
```diff
@@ -1667,7 +1988,7 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
   if (stream->rtp_services & sec_serv_conf) {
     enc_start = (uint32_t *)hdr + uint32s_in_rtp_header + hdr->cc;  
     if (hdr->x == 1) {
-      srtp_hdr_xtnd_t *xtn_hdr = (srtp_hdr_xtnd_t *)enc_start;
+      xtn_hdr = (srtp_hdr_xtnd_t *)enc_start;
       enc_start += (ntohs(xtn_hdr->length) + 1);
```
jfigus: Ditto on the previous comment.  Do we need to add a check on the implied length of the packet since we're trusting xtn_hdr->length and hdr->cc?

test/srtp_driver.c
```diff
@@ -292,9 +322,26 @@ main (int argc, char *argv[])
...
+        printf("testing srtp_protect and srtp_unprotect against "
+               "reference packets with encrypted extensions headers\n");
+        if (srtp_validate_encrypted_extensions_headers() == srtp_err_status_ok)
+            printf("passed\n\n");
+        else {
+            printf("failed\n");
+            exit(1);
+        }
+
+#ifdef OPENSSL
```
jfigus: Really happy to see that you not only added unit test cases, but you also considered the OpenSSL use cases.  If only all code contributors would do this. :-)

srtp/srtp.c
fancycode: See my comment above, this was already checked before my change (at least from what I understand).

srtp/srtp.c
jfigus: But that's only the length on the first header extension.  Isn't this code processing multiple header extensions?  What if the length in one of the subsequent extensions is malformed?

srtp/srtp.c
```diff
@@ -1482,6 +1786,16 @@ srtp_unprotect_aead (srtp_ctx_t *ctx, srtp_stream_ctx_t *stream, int delta,
     }
   }
 
+  if (xtn_hdr && stream->rtp_xtn_hdr_cipher) {
+    /*
+     * extensions header encryption RFC 6904
+     */
+    status = srtp_process_header_encryption(stream, xtn_hdr);
```
jfigus: Be aware that you will be decrypting data here prior to any authentication of the data.  With GCM mode, the authentication occurs when decrypting the payload, which hasn't occurred yet.  IMHO, this is a criticism of AEAD ciphers.  The authentication occurs after decryption.  You may want to consider decrypting the header extensions *after* the payload has been decrypted and authenticated.

srtp/srtp.c
fancycode: Oh right, that check is missing but should be easy to add.

fancycode: Added in c1fc2ba04102b14c161cd2f94869d1e6048f0b71

srtp/srtp.c
fancycode: Thanks for the feedback, changed in da33987a1c0b7179501a91ceff1c8d4cfc0ce1f3.

jfigus: This looks good.  You may need to re-submit the pull request, this commit isn't showing in the pull request.

fancycode: Strange, it shows up for me on https://github.com/cisco/libsrtp/pull/128/commits

fancycode:
Thanks!

PR #148, opened 2016-02-19
use the bigger key size when initializing kfd

aconchillo:
The KDF initialization currently uses the RTP key length size. It might be that one want to use RTP with AES 128 and RTCP with AES 256, which will cause a problem.


aconchillo:
Any comments on this one? Thanks!

aconchillo:
Ping!

paulej:
Who would do that and why? One can't even signal that mode of behavior in SDP, either. DTLS SRTP procedures wouldn't allow such mixed modes.

aconchillo:
You are right. Problem was on a library using libsrtp. Closing this pull request.

PR #149, opened 2016-02-24
Don't create a symlink if there is no $(SHAREDLIBVERSION)

sdroege:
Otherwise we create a symlink with the same name as the target, which will create a symlink pointing to itself.


jfigus:
Thanks for the contribution.

PR #155, opened 2016-04-25
Fix inline build failure on Mac OS X

pengyanhai:


pengyanhai:
It could build either one of the declaration in .h is or the implementation in .c  is not inline. It will fail if both are declared inline. It seems no necessary for this change now. So I close it here.

PR #156, opened 2016-05-02
Fix reading overflow in the case of AES_WRAP with key size 192

pengyanhai:


crypto/cipher/aes_wrap.c
```diff
@@ -190,7 +190,7 @@ static srtp_err_status_t srtp_aes_wrap_context_init (srtp_aes_wrap_ctx_t *c, con
         ) {
         debug_print(srtp_mod_aes_wrap, "Copying last 16 bytes of key: %s",
                     v128_hex_string((v128_t*)(key + SRTP_AES_128_KEYSIZE)));
-        v128_copy_octet_string(((v128_t*)(&c->key.v8)) + 1, key + SRTP_AES_128_KEYSIZE);
+        memcpy(((v128_t*)(&c->key.v8)) + 1, key + SRTP_AES_128_KEYSIZE, c->key_size - SRTP_AES_128_KEYSIZE);
```
paulej: To retain use of v128_copy_octet_string, we could do this:
if (c->key_size == SRTP_AES_192_KEYSIZE)
   v128_copy_octet_string((v128_t*)(((v64_t*)(&c->key.v8)) + 1), key + SRTP_AES_128_KEYSIZE/2);
else
  v128_copy_octet_string(((v128_t*)(&c->key.v8)) + 1, key + SRTP_AES_128_KEYSIZE);

PR #163, opened 2016-06-10
Fix potential memory leak in aes_wrap cipher

pabuhler:
Cipher tests where generating a memory leak as srtp_aes_wrap_set_iv was
called twice on same instance. Now we avoid reallocation.
Also fix potential buffer over flow issue if
srtp_aes_wrap_set_iv_len is called with a different length and
alternate_iv was already allocated.

### DIFF
--- a/crypto/cipher/aes_wrap.c
+++ b/crypto/cipher/aes_wrap.c
@@ -219,9 +219,11 @@ static srtp_err_status_t srtp_aes_wrap_set_iv (srtp_aes_wrap_ctx_t *c, const uin
     /*
      * Allocate space for the IV
      */
-    c->alternate_iv = malloc(c->alternate_iv_len);
     if (!c->alternate_iv) {
-        return srtp_err_status_alloc_fail;
+        c->alternate_iv = malloc(c->alternate_iv_len);
+        if (!c->alternate_iv) {
+            return srtp_err_status_alloc_fail;
+        }
     }
 
     /*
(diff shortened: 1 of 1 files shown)


crypto/cipher/aes_wrap.c
```diff
@@ -219,9 +219,11 @@ static srtp_err_status_t srtp_aes_wrap_set_iv (srtp_aes_wrap_ctx_t *c, const uin
     /*
      * Allocate space for the IV
      */
-    c->alternate_iv = malloc(c->alternate_iv_len);
     if (!c->alternate_iv) {
```
paulej: This looks like a valid concern.  However, rather than performing this check here, should we simply allocate the space in srtp_aes_wrap_alloc() and free it in srtp_aes_wrap_dealloc()?

pabuhler:
I did consider that, always allocate for a length of 8 as it is the max, but was little unsure if alternate_iv would all ways be set/created. Looking through the code it does seam that alternate_iv is expected to always be allocated, it is even accessed with checks in some places. So pre allocating is probable the best. Shall I resubmit with that change?
Calling it alternate_iv instead of just iv is also strange as it is always expected but that change can wait.

paulej:
Sure, please submit that change. It's called alternative IV in the RFCs, thus the name in the code.


pabuhler:
resubmitted as https://github.com/cisco/libsrtp/pull/166

PR #164, opened 2016-06-10
Store new master key when calling srtp_update

pabuhler:
The master key now needs to be stored in srtp srtp_stream_ctx_t.
The update_srtp function was not doing this so master key was
effective all zeros after srpt_update was called.
The code that manages the creation of new streams in side of
srtp_update is basically a copy of what is done in srtp_add_stream,
this code should be re factored to be the same to avoid such
inconsistency errors.

Will cherry pick test to master

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -3324,6 +3324,7 @@ update_template_streams(srtp_t session,
   srtp_err_status_t status;
   srtp_stream_t new_stream_template;
   srtp_stream_t new_stream_list = NULL;
+  int key_len;
 
   if (session->stream_template == NULL) {
     return srtp_err_status_bad_param;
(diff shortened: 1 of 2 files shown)


srtp/srtp.c
```diff
@@ -3358,6 +3359,10 @@ update_template_streams(srtp_t session,
     }
   }
 
+  /* Copy master key copy as provided by the application */
```
paulej: This would replicate the SRTP master key, but if PRIME is used, there are additional checks.  For EKT, we also have to ensure we copy over the SPI value, etc. Do you want to tackle that or would you like me to do it?

pabuhler:
Ideally I would like to see the srtp_add_stream code reused when performing an update, it needs to be changed abit for that. That change should be done in trunk first so as make the diff between these branches as little as possible. srtp_update as I understand it is to completely reset existing streams, only preserving the ROC/srtcp index, so srtp_add_stream was reused then it should take care of all other parameters??
Are there EKT tests that would help ensure refactoring doe snot break?
This changes simply ensures that plain old SRTP still works as expected after call srtp_update. I would prefer that this was first merged and then take a look at refactoring srtp_add_stream.

paulej:
I'm not fully sure what that call was intended to do. It's relatively new. It does what you're describing, I think, but not sure why one would use it.  Presently, there is no EKT or PRIME test code. Nivedita indicated she might try to work on that, but I've not talked to her in several weeks. Given the changes in srtp_add_stream I the EKT branch, I suspect the merge will be manual.  But, I definitely have no objections to doing that in master first. The same logic can then be applied to the EKT development branch.

PR #170, opened 2016-06-15
AES-192 (ICM) approach B

traud:
In the original libSRTP (without OpenSSL), support for AES-256 was added in May 2010 with commit 5df951a, but AES-192 was not implemented (see `crypto/cipher/aes.c:aes_expand_encryption_key`). Since 2013 with OpenSSL enabled (see pull request #34), libSRTP supports AES-192 as well. Or stated differently:

If you want the [crypto suite](https://www.iana.org/assignments/sdp-security-descriptions/sdp-security-descriptions.xhtml) `AES_192_CM_HMAC_SHA1_80` in your VoIP application, you have to go for
`./configure --enable-openssl`
otherwise, just AES-256 and AES-128 work. As a side-effect, you get AES-NI and AES-GCM.

Even with the current source code, AES-192 works only between VoIP applications which use the same library (for example locally on the same computer). I did not squash the two commits on purpose, hopefully to make the overall pull request easier to understand.

This is approach/alternative B to solve this issue. Because I am just a contributor and not a member of the team, I cannot decide whether the approach A or B is the correct one. If you do not like the other approach A, please, consider this simpler change-set here for inclusion, especially because several VoIP/SIP clients leverage AES-192 already (like Acrobits Groundwire and Media5-fone).

### DIFF
--- a/crypto/kernel/crypto_kernel.c
+++ b/crypto/kernel/crypto_kernel.c
@@ -76,6 +76,10 @@ extern cipher_type_t aes_icm;
 #ifndef OPENSSL
 extern cipher_type_t aes_cbc;
 #else
+#ifndef SRTP_NO_AES192
+extern cipher_type_t aes_icm_192;
+#endif
+extern cipher_type_t aes_icm_256;
 extern cipher_type_t aes_gcm_128_openssl;
 extern cipher_type_t aes_gcm_256_openssl;
 #endif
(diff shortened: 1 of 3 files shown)


pabuhler:
Prefer this approach as it is more consistent with rest of code for now.
Also would prefer it was against master, any reason for 1.5.x?

pabuhler:
Hi @traud , any chance of get this this against master? We are planing on a 2.1 release and would be good to get it in. If I don't hear back soon I might try to apply to master myself.

paulej:
It would be good if test code was added for AES-192.  There is a test for 256 (srtp_validate_aes_256()) and having a version for 192 would be good.

pabuhler:
This has been applied to master in #253 as 4a74b82, if that commit is satisfactory then this should be closed.

traud:
Fine. There should be a release note that AES-192 only works with libSRTP 2.1 or newer; and only when OpenSSL was configured. Furthermore, there should be a note with the next libSRTP 1.5.x that AES-192 does not work with that.

PR #171, opened 2016-06-15
AES-192 (ICM) approach A

traud:
In the original libSRTP (without OpenSSL), support for AES-256 was added in May 2010 with commit 5df951a, but AES-192 was not implemented (see `crypto/cipher/aes.c:aes_expand_encryption_key`). Since 2013 with OpenSSL enabled (see pull request #34), libSRTP supports AES-192 as well. Or stated differently:

If you want the [crypto suite](https://www.iana.org/assignments/sdp-security-descriptions/sdp-security-descriptions.xhtml) `AES_192_CM_HMAC_SHA1_80` in your VoIP application, you have to go for
`./configure --enable-openssl`
otherwise, just AES-256 and AES-128 work. As a side-effect, you get AES-NI and AES-GCM.

Even with the current source code, AES-192 works only between VoIP applications which use the same library (for example locally on the same computer). This is because of one single `#if` clause in `srtp/srtp.c`. However, this pull request takes another approach and replaces all (deprecated) `AES_xxx_ICM` with the new `AES_ICM` from commit a5754a6. Those changes avoid this issue and enable AES-192 as a side-effect. A subsequent change could do the same for GCM and replace all `AES_xxx_GCM` with a new symbol called `AES_GCM`. Then, a subsequent change could remove `srtp_cipher_t->algorithm`and replace that with `srtp_cipher_t->type->id` which then matches `srtp_crypto_policy_t->cipher_type`. I did not squash the three commits (of this possible series of changes), hopefully to make the overall pull request easier to understand.

This is approach/alternative A to solve this issue. Because I do not know how to bundle one pull request for different branches, I started with branch 1.5.x. Instead, if you need the pull request for master or additionally a request for the master branch, please, say so.

Currently, this pull request is in conflict with the new feature ‘RTP header-extension encryption’ ce37ef6 on the 2.x branch. If this pull request is considered for inclusion, I think about that conflict.

### DIFF
--- a/crypto/cipher/aes_icm.c
+++ b/crypto/cipher/aes_icm.c
@@ -121,15 +121,13 @@ aes_icm_alloc_ismacryp(cipher_t **c, int key_len, int forIsmacryp) {
 
   /* set pointers */
   *c = (cipher_t *)pointer;
+  (*c)->algorithm = AES_ICM;
   switch (key_len) {
   case 46:
       (*c)->algorithm = AES_256_ICM;
       break;
   case 38:
       (*c)->algorithm = AES_192_ICM;
-      break;
-  default:
-      (*c)->algorithm = AES_128_ICM;
       break;
   }
   (*c)->type = &aes_icm;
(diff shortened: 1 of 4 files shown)


pabuhler:
Prefer other approach in #170, making a define for cypher "family" is an ok idea that can be looked at after.

pabuhler:
Close this as we are going with he approach in #170

PR #172, opened 2016-06-15
convenience methods for all crypto suites

traud:
- depends on the inclusion of the request #170 or request #171
- adds the convenience methods for last remaining crypto suite AES-192 (ICM), for the sake of symmetry

Without those methods, downstream projects have to deal with the internals of `srtp_crypto_policy_t`. The `#ifdefs` are chosen to allow downstream projects to detect AES-192 capability plus (if added with #170 or #171) whether `-lsrtp` is a version of the library which allows a working AES-192 crypto suite.

I am not sure, whether I met the current coding style guides. Furthermore, the comments might need an update to link to the latest RFC rather than its draft. If I should update/change anything, please, say so. Please, consider this change for inclusion because I find it quite handy when it comes to detect the features of libSRTP automatically.

### DIFF
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -617,6 +617,97 @@ crypto_policy_set_null_cipher_hmac_sha1_80(crypto_policy_t *p);
 
 
 /**
+ * @brief crypto_policy_set_aes_cm_192_hmac_sha1_80() sets a crypto
+ * policy structure to a encryption and authentication policy using AES-192
+ * for RTP protection.
+ *
+ * @param p is a pointer to the policy structure to be set
+ *
+ * The function call crypto_policy_set_aes_cm_192_hmac_sha1_80(&p)
+ * sets the crypto_policy_t at location p to use policy
+ * AES_CM_192_HMAC_SHA1_80 as defined in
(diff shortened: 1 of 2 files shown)


pabuhler:
Will look at merging once #170 is in.
Master only ??

paulej:
Agree this should go into master only.  I assume the noted linking issues will be resolved once we merge the AES192 support without OpenSSL.  (That PR #170 or #171, whichever is chosen.)

traud:
> the noted linking issues will be resolved once we merge the AES192 support without OpenSSL

I am not aware of any source code which is going to add AES-192 without OpenSSL.

> Master [branch] only?

When AES-192 (with OpenSSL) is fixed, downstream projects must be able to detect that somehow in libSRTP. My proposed way would be this convenience method. If that method is available, libSRTP offers a working AES-192 implementation. That is better than static version-number checks, because distributions might backport that fix and this method here.

pabuhler:
This as been applied to master in #253 as 8872f71 , consider closing this if that is merged.

pabuhler:
#253 was merged

PR #178, opened 2016-06-20
Remove extra check that generates compiler error

pabuhler:
The code generated the following error:
error: comparison of unsigned expression < 0 is always false
[-Werror=type-limits]

This check is also not needed as four lines up there is
if (!((uint8_t*)enc_start <= (uint8_t*)hdr + pkt_octet_len))
which checks exactly the same thing.


pabuhler:
resubmiting

PR #183, opened 2016-07-11
fixed ./configure --enable-syslog

### LINKED ISSUE #48: ERR_REPORTING_SYSLOG is never defined
Does the `--enable-syslog` work? I read in the code:

> - Errors will be reported to ERR_REPORTING_FILE, if defined, and to
> - syslog, if ERR_REPORTING_SYSLOG is defined.

But the fact is that ERR_REPORTING_SYSLOG is never defined, so I don't know how syslog is supposed to work.

traud:
Yes, I know, this feature got removed with commit f063b90 in branch `2_0_x_throttle` and `master`. However, it was not removed in branch `1_5_x_throttle`. And released things should work. Furthermore, I needed this feature to debug an internal issue within libSRTP, while being compiled as a shared library. I could not use the API for this. I did not figure out, how logging to `stdout` or `/dev/console` works with a shared library. If this pull request gets rejected, please, someone explains how to use those alternatives instead.

Anyway, now logging to `LOG_SYSLOG` works when enabled via the configure script. Tested on Ubuntu 16.04 LTS. Please, consider this change for inclusion at least into the branch `1_5_x_throttle`. Additionally, please, consider to reintroduce this feature in the other branches again – undo commit f063b90.

resolves #48

### DIFF
--- a/config_in.h
+++ b/config_in.h
@@ -21,6 +21,9 @@
 /* Define to use logging to stdout. */
 #undef ERR_REPORTING_STDOUT
 
+/* Define to use syslog logging. */
+#undef ERR_REPORTING_SYSLOG
+
 /* Define this to use ISMAcryp code. */
 #undef GENERIC_AESICM
 
@@ -173,9 +176,6 @@
 
 /* Write errors to this file */
 #undef USE_ERR_REPORTING_FILE
-
-/* Define to use syslog logging. */
-#undef USE_SYSLOG
 
 /* Define WORDS_BIGENDIAN to 1 if your processor stores words with the most
    significant byte first (like Motorola and SPARC, unlike Intel). */
--- a/configure
+++ b/configure
@@ -5897,7 +5897,7 @@ fi
 
 if test "$enable_syslog" = "yes"; then
 
-$as_echo "#define USE_SYSLOG 1" >>confdefs.h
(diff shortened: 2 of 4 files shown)


pabuhler:
I would be inclined to reject this based on the arguments in  #48 .
The fact that it does not work indicates that it is never used..
Instead I would suggest a new api in 2.1.x that lets the application receive debug statements as suggested in #48

jfigus:
Try the --enable-stdout option to log errors to the console (stdout).  Libraries shouldn't log directly to syslog.  As pabuhler said, we should add an API that allows errors to be trapped by the application using libsrtp.  Then the application can log the error as desired, such as syslog.  Keep in mind that not all platforms support syslog.

traud:
Mhm, I tried `--enable-stdout` and it did not work for me. Is there anything else, I have to do like redirects?

Syslog was my rescue, especially because I had not that control over the project using libSRTP. Consequently, any logging API would not have helped me to track this down. Anyway, the code in libSRTP 1.5.x is broken. If it is not fixed (why, the change is simple and I do not follow the arguments in #48; a library should not log there on default – this option here is not enabled on default), is that dead code at least removed from libSRTP 1.5.x?

paulej:
Given @pabuhler reference to #230, should we close this PR?

pabuhler:
Syslog is no longer supported and there is no intention to support it in the future.

PR #184, opened 2016-07-11
AES-GCM: Key Derivation broken because of shorter Master Salt

### LINKED ISSUE #46: AES GCM mode using wrong SALT size
The Master SALT for GCM mode should be 96 bits.  Currently the GCM mode uses the same SALT_SIZE directive as ICM mode, which is set to a value of 14 (112 bits).   As a result, the values derived from the KDF are wrong.  Additionally, applications that use GCM mode are required to provide 30 bytes of key/salt material, where only 28 bytes should be required.

### LINKED ISSUE #179: AES-GCM Session Key/Salt Derivation is Incorrect
[RFC 7714 Sec 11.](https://tools.ietf.org/html/rfc7714#section-11) states that AEAD_AES_128_GCM must use the KDF as defined in [RFC 3711 Sec 4.3.1](https://tools.ietf.org/html/rfc3711#section-4.3.1).

IV generation for this KDF is defined in [RFC 3711 Sec 4.3.3](https://tools.ietf.org/html/rfc3711#section-4.3.3): `IV [is] equal to (x*2^16)` where x is the master salt.

The problem arises because the master IV for AES-GCM is 12 bytes, as opposed to 14 bytes for the default AES-CM algorithm specified in RFC 3711.  [Appendix B.3](https://tools.ietf.org/html/rfc3711#appendix-B.3) has a test vector for IV generation. A 14 byte master salt of `0EC675AD498AFEEBB6960B3AABE6` will result in an IV of `0EC675AD498AFEEBB6960B3AABE60000`.

For AES-GCM then, a 12 byte master salt, e.g. `0EC675AD498AFEEBB6960B3A` should result in an IV of `00000EC675AD498AFEEBB6960B3A0000`.  Unfortunately libSRTP uses an incorrect IV of `0EC675AD498AFEEBB6960B3A00000000`, leading to an incorrect session key and salt being derived.

traud:
With SRTP, the key exchanged with the remote party is called the Master key, because six Session keys are derived from that Master key. For this, a Key-Derivation Function (KDF) is defined so both parties derive the same Session keys from that Master key. The KDF must be the same, otherwise different keys are used and the remote party would not be able to decrypt the encrypted RTP data anymore.

Furthermore, that Master Key is [salted](https://en.wikipedia.org/wiki/Salt_%28cryptography%29#Benefits) and that part is called Master Salt. Both the Master Key and Master Salt are used in KDF. In [RFC 3711](https://tools.ietf.org/html/rfc3711#section-4.3) from 2006, the salt was 112 bits = 14 octets long. [RFC 6188](https://tools.ietf.org/html/rfc6188#section-7.2) contains more details about this, for example test cases in its section 7.2. With [RFC 7714](https://tools.ietf.org/html/rfc7714#section-11), AES-GCM was introduced which uses the same key length but a two byte shorter salt (12 octets = 96 bits). Nevertheless, RFC 7714 mandates to re-use the same KDF as before.

In commit Cisco/libSRTP@8719f95, the decision was made to pad the Master Salt with two zero bytes. Actually in the source code of libSRTP, the concatenation of the Master Key and Master Salt was simply forwarded to KDF. This led to a padding of zeros not at the start (left) but at the end (right) of the Master Salt.

This issue was found via [Acrobits Groundwire](https://www.acrobits.net/products/retail/groundwire), which offers AES-GCM since April 2016. In the Apple iOS variant, you enable AES-GCM via Settings → Preferences → Security → (SRTP) SDES. According to the Acrobits support, they do not use libSRTP for this. My personal tests revealed, their implementation aligns the Master Salt to the right, see issue #179 for more details (found independently).

I expect this pull request be closed, because merging this request would make AES-GCM incompatible with existing libSRTP implementations. Although I am not aware of that many, yet. However, the purpose of this pull request was to document this issue. The attached commit (simply add `.patch` to the URL to get a diff) allows to test existing implementations for this issue. Reading through the related RFCs (7714, 6188, and 3711), yet, I was not able to find a single statement related to this. It looks like, whether to move right or left is undefined:
- RFC 6188 test cases show a right-alignment of `x`. However, there, `x` was known in size (112 bits).
- RFC 3711 mentions a right-alignment for `x`. However, again, this was not about the Master Salt but about the `key_id`.
- In RFC 7714, 6188, or 3711, I found nothing which got left-aligned – all I found was right-alignment. Furthermore, when one day, a larger salt is introduced, should AES-ICM be padded on the right as well?

Although I found no argument for left-alignment – as it is done by libSRTP – my current position is, that this is undefined in the RFCs itself. The author of those RFCs @DavidMcGrew, or the author of that commit @JFigus, or anyone else who sees a statement which justifies the left-aligment, please, comment. Before I raise an issue in the IETF Audio/Video Transport Core Maintenance Working Group ([AVTCore](https://tools.ietf.org/wg/avtcore/)), I would like to clarify this issue here in libSRTP. Because I could bet, I missed the sentence which clarifies this.

fixes #46
resolves #179, found independently

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -539,12 +539,15 @@ srtp_stream_init_keys(srtp_stream_ctx_t *srtp, const void *key) {
   debug_print(mod_srtp, "rtp salt len: %d", rtp_salt_len);
 
   /* 
-   * Make sure the key given to us is 'zero' appended.  GCM
+   * Make sure the key given to us is 'zero' prepended. GCM
    * mode uses a shorter master SALT (96 bits), but still relies on 
    * the legacy CTR mode KDF, which uses a 112 bit master SALT.
    */
   memset(tmp_key, 0x0, MAX_SRTP_KEY_LEN);
-  memcpy(tmp_key, key, (rtp_base_key_len + rtp_salt_len));
+  memcpy(tmp_key, key, rtp_base_key_len);
+  memcpy(tmp_key + rtp_base_key_len + 14 - rtp_salt_len,
+         key + rtp_base_key_len,
+         rtp_salt_len);
 
   /* initialize KDF state     */
   stat = srtp_kdf_init(&kdf, AES_ICM, (const uint8_t *)tmp_key, kdf_keylen);


srtp/srtp.c
```diff
@@ -539,12 +539,15 @@ srtp_stream_init_keys(srtp_stream_ctx_t *srtp, const void *key) {
...
   /* 
-   * Make sure the key given to us is 'zero' appended.  GCM
+   * Make sure the key given to us is 'zero' prepended. GCM
    * mode uses a shorter master SALT (96 bits), but still relies on 
    * the legacy CTR mode KDF, which uses a 112 bit master SALT.
    */
   memset(tmp_key, 0x0, MAX_SRTP_KEY_LEN);
-  memcpy(tmp_key, key, (rtp_base_key_len + rtp_salt_len));
+  memcpy(tmp_key, key, rtp_base_key_len);
+  memcpy(tmp_key + rtp_base_key_len + SALT_SIZE - rtp_salt_len,
```
paulej: In the current code, this is SRTP_SALT_SIZE. Also, I think this is only defined when compiling against OpenSSL.

imhotepisinvisible:
As I mentioned in my issue, I think the RFC is clear (if confusing) about how the alignment should be handled.  RFC 3711 Sec 4.3.3 states that `IV [is] equal to (x*2^16) where x is the master salt.`.  So you simply have to multiply the master salt by 2^16, whether it's 14 bytes, 12 bytes, 2 bytes or 64 bytes.

traud:
In my copy of RFC 3711, I do not find that statement. `x` is not the Master Salt but a derived value already (XOR, …). Consequently, the question is: How shall the Master Salt be aligned in the calculation for `x`? The libSRTP team found this gotcha and went for left-alignment without updating the RFC in progress. I wonder why.

imhotepisinvisible:
x is defined in section 4.1.1:

```
Let x = key_id XOR master_salt, where key_id and master_salt are
aligned so that their least significant bits agree (right-
alignment).
```

The alignment is defined to be right-alignment.  It then determines the PRF:

```
The n-bit SRTP key (or salt) for this packet SHALL then be
derived from the master key, k_master as follows:

   PRF_n(k_master, x).

(The PRF may internally specify additional formatting and padding of
x, see e.g., Section 4.3.3 for the default PRF.)
```

Assuming the use of the default PRF, Section 4.3.3 shows that the IV is x*2^16.  In full:

```
The currently defined PRF, keyed by 128, 192, or 256 bit master key,
has input block size m = 128 and can produce n-bit outputs for n up
to 2^23.  PRF_n(k_master,x) SHALL be AES in Counter Mode as described
in Section 4.1.1, applied to key k_master, and IV equal to (x*2^16),
and with the output keystream truncated to the n first (left-most)
bits.  (Requiring n/128, rounded up, applications of AES.)
```

traud:
I am not sure, why you repeat the chapter about `x*2^16` because that says nothing about the alignment of the Master Salt itself, just of a derived value. I thought, I have explained that in detail. Therefore, please, which part of my rationale is unclear exactly?

Let us have a look at your other rationale: In my original post, I meant exactly this chapter 4.1.1, when I referenced RFC 3711 and its right-alignment statement – your first cite. Back when it was written, that cite dealt with a Master Salt known in size but a shorter `key_id`. Therefore, that statement is about the right-alignment of `key_id`. Now, ten years later this statement is re-interpreted to right-align a shorter Master Salt as well. Mhm, you can align only something to a reference, not the reference to itself. At least, not in my mother language. Because I am not an US American, I have a translation gap here as well. Anyway, if you continue with that statement, you go not for the Master Salt but the least significant bit as reference. In that case, the question arises: What is the least significant bit of a master salt which is shorter than 112 bits? Normally, I would go for its bit 0. Therefore, I found that issue at all, because I thought the same way as you do. However, the libSRTP team defined the least significant bit differently, because of their padding – for which I have not found any rationale.

> key_id and master_salt are aligned so that their least significant bits agree (right-alignment).

A) Is that the statement, which clarifies the alignment of the Master Salt for `x`? In that case, the reference implementation and all its derived projects are broken for AES-GCM.

B) Is there a statement unknown/missing, which clarifies what to do with longer/shorter Master Salts = what is its least significant bit? In that case, one of the RFCs (7714 or 3711) should be updated through an errata, because several people have not found that statement, yet, four weeks later.

With any decision, there are incompatible – inoperable – implementations out there already. This is a pity. Therefore, it would be interesting to hear the position of the authors (of that commit and/or the RFC). Depending on its severity, this might even qualify for new crypto-suite names, which do not face such an interoperability issue for sure. Puh.

paulej:
I think we should take this to AVT Core.  The issue as I see it is that the text in 3711 appears to have only considered a 14 octet salt length.  Guidance should have been in the GCM spec for SRTP about where to put padding.  Libsrtp padded on the right to form a 14 octet salt.  If it didn't do that, the end result following RFC3711 would have been a salt that is still 2 octets too short.

imhotepisinvisible:
> What is the least significant bit of a master salt which is shorter than 112 bits? Normally, I would go for its bit 0. Therefore, I found that issue at all, because I thought the same way as you do.

I think we agree. That is why I believe the libSRTP implementation is incorrect.

> because that says nothing about the alignment of the Master Salt itself, just of a derived value

x is the derived value, which is itself derived from the master salt.  Once you have x (i.e. Master salt XOR (label||r)) you multiply that by 2^16.  The RFC adds no further information about any additional padding.

> the end result following RFC3711 would have been a salt that is still 2 octets too short

True, hence the confusion.  However, because the salt is being treated as a number (if you're multiplying it, it has to be a number), that would suggest right-aligning this short salt, and padding with 0s on the left.  If a number is too short, you merely prepend 0s to make it fit: 10 == 0010 == 000010.  However I completely agree: as it's not explicitly specified clarification should be requested.

paulej:
Just so I don't lose track, this is also discussed as https://github.com/cisco/libsrtp/issues/179

paulej:
I updated issue #179 with latest information.  I think we can close this one.

PR #185, opened 2016-07-11

traud:


paulej:
I think it would be better to keep the current style with case statements and have an explicit return value like this:
~~~
  case srtp_profile_aes128_cm_sha1_80:
    return 16;
    break;
  case srtp_profile_aes128_cm_sha1_32:
    return 16;
    break;
~~~

As opposed to:

~~~
  case srtp_profile_aes128_cm_sha1_80:
  case srtp_profile_aes128_cm_sha1_32:
  case srtp_profile_aead_aes_128_gcm:
    return 16;
    break;
~~~

Obviously, both will work, but the latter is more error-prone as people make code changes.

Otherwise, I think the code looks good.  I'm not sure how/when the 1.5 throttle will be updated.  Perhaps @jfigus can weigh in on that.

traud:
> is more error-prone as people make code changes.

I am sorry, I do not understand that concern, yet. Please, explain in more detail like describing an example.

I did this on purpose because the value returned is equal for these cases and can never ever change again. I found this easier to read, because I was confused at the start why every case got its own return value. Perhaps it is even easier to use the const values of GCM like `SRTP_AES_xxx_GCM_KEYSIZE_WSALT - SRTP_AEAD_SALT_LEN`, `SRTP_AEAD_SALT_LEN = 12`, and move `SRTP_SALT_LEN = 14` of the OpenSSL part to an always available header.

Actually, I do not like literal constants which have Defines somewhere else in the code already. Therefore, I tried to minimize them here in this part of the code (a bit).

paulej:
The code previously did not use fall-through case statements, which is what you introduced there. Some people like them, but the original code did not have them and I don't see the point of adding them now. Yeah, it's rare that code gets changed, but ... it just did :)  So, I think it's best to use them.

This library makes extensive use of constants defined elsewhere. Personally, I think it's a good idea, because if a change is needed you have to be really careful to go make sure you find all the places in the code that use them.  Also, it makes it easier to read.  If the code was "key + key_len -14 + 2" or whatever, you would not know the purpose of the literal constants.  These kinds of things help with reducing errors and it makes it easier for people to maintain the code.

traud:
> I don't see the point of adding them now.
> You find all the places in the code that use constants.

Then, I have to explain. There is a misunderstanding. I am exactly about the same. I would be happy not to write `return 14` but return a previously defined constant. Because currently `(SRTP_)SLAT_LEN` is not reachable in every case (except when `--enable-openssl` is configured), I went for fall-through statements to couple those cases which relate and never change. That was the reason to introduce them, rather to add even more cases with literals. Although those cases do not have to be changed in future, it makes it easier to read the code that those return values happen because of a property which is shared between those cases.

I was very confused, that the library has constants for this already, but does not use them here. Therefore, I had to re-think, research, triple-check for _every_ case statement why a literal was used and where its value comes from. Coupling them, might help other maintainers. Therefore, I changed that because of my findings. I see no benefit of having separate cases. Those cases are not just equal but the same.

paulej:
Just so others are not confused by the above, the constants I'm referring to are related to #184 where a bare 14 is used.  The constants defined were used where they were needed.  Now, you found a new place where they are needed, but they're not reachable.  I think the right solution is to move the constant to some place where it is common.  And, we might even want to rename it to be more specific.  Some constants like SRTP_SALT_LENGTH are confusing since GCM has a salt length that is different.

traud:
I do not understand, yet. These constants were used here in the part of the code as literals. Consequently, a reachable `SRTP_SALT_SIZE` would have been useful here as well. Whom do you address with ‘we’: Shall I – as an external contributor – refactor the architecture to make that symbol reachable, otherwise these both changes are not accepted? Before I go for that, what about the name: `SRTP_ICM_SALT_SIZE` or `_LENGTH` or `_LEN`?

paulej:
We would be you or the others who contribute to libsrtp, including the maintenance team inside Cisco.

paulej:
As for naming, no strong preference, except for relative consistency.

pabuhler:
Will you resubmit, keeping the coding style?
Also is there any reason you would like this in 1.5.x? Preferably it should go just to master.

pabuhler:
Closing as this will not be applied to 1.5.x at this point.
Issues #260 has been created to follow up on the issue raised in this thread relating to use of literal constants.

PR #186, opened 2016-07-11

traud:


paulej:
I have the same comments as for #185.  However, I think this should go directly into master.  I'm not sure of throttle's role right now.  Again, a good question for @jfigus.

pabuhler:
agree, not the throttle.

PR #188, opened 2016-07-11

traud:


paulej:
Same comments as for #185.

paulej:
I would prefer we not change the code to have fall-through case statements.

pabuhler:
This has been reapplied to master in #258.

PR #189, opened 2016-07-11

traud:


paulej:
Looks good to me.

pabuhler:
probably not on the throttle

paulej:
Similar to #190, can we close this one?

traud:
Mhm. When I read the discussion in Pull Request #190 – which I understood there for `2_0_x_throttle` – why am I asked this question here in the `1_5_x_throttle` as well?

paulej:
It would be reasonable to comment in each PR.  We could incorporate that PR here, but usually only security fixes go into the throttle for older releases.

However, this one is pretty straightforward. Maybe the team should discuss again.

PR #190, opened 2016-07-11

traud:

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -408,7 +408,7 @@ srtp_stream_clone(const srtp_stream_ctx_t *stream_template,
   srtp_err_status_t status;
   srtp_stream_ctx_t *str;
 
-  debug_print(mod_srtp, "cloning stream (SSRC: 0x%08x)", ssrc);
+  debug_print(mod_srtp, "cloning stream (SSRC: 0x%08x)", ntohl(ssrc));
 
   /* allocate srtp stream and set str_ptr */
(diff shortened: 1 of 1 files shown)


paulej:
Looks good.

pabuhler:
Probably not on the throttle.

paulej:
Is there a reason for the PR on the throttle branch?  Can we just close this one?

traud:
> not on the throttle

Without any rationale, it is difficult to respond. By the way, are you asking me or someone else? Now, I have to make up your reason(s) how you came to that conclusion. This Pull Request is about a bug report, not a new feature. The ‘throttles’ are still maintained and especially 1_5_x_throttle is for those distributions which still rely on libSRTP 1.4.x. Therefore, I propose this change even for the other two release branches 2_0_x and 1_5_x. I know this change here is just cosmetic, anyway a bug is a bug.

Or are you stating that (came to that conclusion because) the throttle branches (release branches) only see *security* bug fixes – normal bug fixes go to the master branch only?

paulej:
This particular PR is only for 2.0, right? So putting it into master would be right. The next 2.0 release will come from master.  For bugs in 1.x, yes those would go into that throttle.

traud:
> The next 2.0 release will come from master.

Ahh. OK, then this one here can be closed. Just to reiterate what I understood: Next time, when I find a bug, I create a Pull Request against `master` *and* `1_5_x_throttle` but not `2_0_x_throttle`.

paulej:
Security issues would definitely go to both. If they're cosmetic, maybe only master.  If we out everything into 1.x then we never move forward.

PR #191, opened 2016-07-11

traud:


paulej:
Looks good.

pabuhler:
yip, lets merge

PR #192, opened 2016-07-14
Replace octet_string_is_eq with a constant-time implementation

mmdriley:
This function is used to check authentication tags in srtp_unprotect.
The current early-exit implementation might offer a timing sidechannel,
enabling attackers to brute-force a correct HMAC to a forged message.

Such attacks shouldn't be possible if replay protection is enabled, but
this is nonetheless good defense in depth.

The implementation is similar to CRYPTO_memcmp from OpenSSL/BoringSSL.

### DIFF
--- a/crypto/math/datatypes.c
+++ b/crypto/math/datatypes.c
@@ -440,14 +440,21 @@ bitvector_left_shift(bitvector_t *x, int shift) {
 
 }
 
-
 int
 octet_string_is_eq(uint8_t *a, uint8_t *b, int len) {
   uint8_t *end = b + len;
+  uint8_t accumulator = 0;
+
+  /*
+   * We use this somewhat obscure implementation to try to ensure the running
+   * time only depends on len, even accounting for compiler optimizations.
+   * The accumulator ends up zero iff the strings are equal.
+   */
   while (b < end)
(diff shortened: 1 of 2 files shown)


crypto/math/datatypes.c
```diff
@@ -440,14 +440,21 @@ bitvector_left_shift(bitvector_t *x, int shift) {
...
+   * The accumulator ends up zero iff the strings are equal.
+   */
   while (b < end)
-    if (*a++ != *b++)
-      return 1;
-  return 0;
+    accumulator |= (*a++ ^ *b++);
+
+  /* Collapse zero/nonzero to zero/one. */
+  return !!accumulator;
```
paulej: I think you wanted only one "!" here.

mmdriley: I didn't, actually -- "!!" maps 0 to 0 and everything nonzero to 1. But I agree this was a bit opaque, so I've replaced it with a more straightforward comparison.

mmdriley:
Thanks! Do you need anything from me before merging?

paulej:
No. We just need to have others discuss and then test.  Unfortunately, we have a growing backlog. Not sure of an ETA.

PR #196, opened 2016-08-04
Update sha1_driver.c

bryongloden:
```
[crypto/test/sha1_driver.c:87]: (error) Memory leak: test_case
[crypto/test/sha1_driver.c:91]: (error) Memory leak: test_case
```

Found by https://github.com/bryongloden/cppcheck

### DIFF
--- a/crypto/test/sha1_driver.c
+++ b/crypto/test/sha1_driver.c
@@ -79,12 +79,16 @@ hash_test_case_add(hash_test_case_t **list_ptr,
   unsigned tmp_len;
 
   test_case = malloc(sizeof(hash_test_case_t));
-  if (test_case == NULL)
-    return srtp_err_status_alloc_fail;
-  
-  tmp_len = hex_string_to_octet_string((char *)test_case->data, hex_data, data_len*2);
-  if (tmp_len != data_len*2)
+  if (tmp_len != data_len*2) {
+    free(test_case);
     return srtp_err_status_parse_err;
+  }
+
+  tmp_len = hex_string_to_octet_string((char *)test_case->hash, hex_hash, hash_len*2);
+  if (tmp_len != hash_len*2) {
+    free(test_case);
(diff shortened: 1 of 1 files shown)


paulej:
Much of the test code doesn't free memory an error is hit, as errors result in immediate program termination.  No objection to addressing this, but if we fix one, we make this change, we should make a comprehensive set of changes for consistency.

I'll note, though, that nothing is gained by making these changes, other than to pacify a static analysis tool.

bryongloden:
Thanks for the clever explanation and _wow_, because you have the potential of absolutely making no sense -- I believe you want to deliberately keep buggy codes instead of actually working with the open source community to patch this. That's an interesting, incredible, excellent, amazing, and awesome attitude man ♥️

paulej:
I'm not sure why my reply upset you.  I said I'm not opposed to fixing it, but if we fix it in one place, we should fix it everywhere.  Good coding practice would call for consistency, right?

Maybe it was my last comment, where I point out that fixing this really doesn't buy us a lot. That's just a statement of fact, as that code is not compiled into the SRTP library.  I'm not trying to be negative.

bryongloden:
Don’t be offended man. We just have the facts, references, and expertise that explain why building error-free computer programs outweigh the benefits of consistent coding practice.

We’ll allow some degree of opinion based your expert experience. Would like us to close this commit?

paulej:
No, don't close it.  We should have the team of maintainers look at this and decide what action to take.

bryongloden:
LOL!!! Hey there @paulej 😀 We're getting a good laugh here. **_Please_** help us understand why you'd merge our patch, then revert it 5 minutes later? And just to be perfectly clear, we're bug bounty hunters, but we're not in it for cash; we ❤️ ❤️ ❤️ the kudos 👍

paulej:
I merged it, but the test suites failed.  The change was reverted.

PR #198, opened 2016-08-11
Update sha1_driver.c

bryongloden:
Our first patch was technically incorrect because it only patched one, or half, of the errors found by cppcheck. And instead of patching the other, second error, we inadvertently inserted a line of code that called `free()` on a NULL pointer. However, we believe it shouldn't have caused the test suites to fail because if `test_case` is NULL, then by definition no action occurs. We just didn't completely mitigate the risk.

Without seeing the test output it was difficult for us to debug this. Furthermore, we don't believe we have access to run the test suite locally. @paulej do you have any more insight that could lead us to find the cause of the crash. We'd like to rule out the possibility that it was our first patch that caused the test suites to fail.

For what its worth, we patched the correct two errors in this PR. And sorry we didn't get it right the first time 😁

### DIFF
--- a/crypto/test/sha1_driver.c
+++ b/crypto/test/sha1_driver.c
@@ -79,12 +79,16 @@ hash_test_case_add(hash_test_case_t **list_ptr,
   unsigned tmp_len;
 
   test_case = malloc(sizeof(hash_test_case_t));
-  if (test_case == NULL)
-    return srtp_err_status_alloc_fail;
-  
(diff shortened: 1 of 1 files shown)


paulej:
What I did when I merged the code from your previous commit was:
$ ./configure --enable-openssl
$ make
$ makeruntest

It fails in the sha1_driver when running the test.  I did not do any work to figure out why. :)

paulej:
Seems to build and tests pass.  I'll wait for the team to review it again.

PR #199, opened 2016-08-15
Some improvements into rtp_decode

sobomax:
o New option -s for the rtp_decode to set key size and tag size based on RFC 4568 crypto suite specification

o correct example with regards to setting proper tag size

o strip out auth padding when generating pcap output it could potentially screw rtp audio decoder, which is true with both wireshark and rtpproxy's extractaudio


paulej:
There is a conflict with master now.  Not sure how that happened since we've not committed many PRs, but we're getting ready for more.  Would you mind fixing the conflict?

I also noted line 96 had an extra ; at the end.  I could not test the functionality, but will once the conflict is resolved.

sobomax:
Paul, sure, I'll take a look at it tomorrow. Thanks!

sobomax:
The updated patch has been pushed. Thanks!

paulej:
Thanks!  I've not tested it, but it looks good to me. I'll try to find some time today.

Issue #200, opened 2016-08-29
libsrtp 2.0.0 no longer exposes [srtp_]crypto_kernel_replace_cipher_type


JonathanLennox:
Rev a9ac898d removed the install of the crypto/include header files into the srtp install directory.

A consequence of this is that [srtp_]crypto_kernel_replace_cipher_type is no longer exposed.  Our system uses this API to replace srtp's crypto implementation with our own, optimized one, so this is blocking us moving up to 2.0.0.

Any thoughts on the best way to fix this?

pabuhler:
It certainly looks like this API was meant to be "public" along with srtp_crypto_kernel_replace_auth_type.
Although I don't know how much of the other functions in the file are meant to be accessed externally.
I see the debug functions have wrappers in srtp.h
We definitely should expose the api again but maybe in a cleaner fashion?
Perhaps in a sub include folder so it is clearly not part of srtp usage api or in a new include file specifically for this api.
Ideally we should only expose what is needed for this api and not the entire krypto kernel but at this point not sure on the amount of work required and if it would be worth it.

This i guess should go in to master, what will be a 2.1.x release at some point.

Other suggestion, will you work on this?

jfigus:
Part of the refactoring for 2.0 was to improve the opacity of the libsrtp data structures.  Prior to 2.0 all those header files were installed into the system, allowing the applications to reach into any private data structure within libsrtp.  With 2.0 we no longer install all all header files, some of which use to contain a mixture of both public and private entry points.  If there's some capability that needs to be exposed through the public API, then we'll need to extend this through one of the existing public header files, or create a new public header file.

IMHO, we should try to isolate the crypto from SRTP as much as possible.  Using OpenSSL for crypto helps achieve this goal.  OpenSSL already has an abstraction layer for using hardware (or kernel) crypto modules.  Why reinvent this wheel in libsrtp?  Another model to consider would be to implement the kernel crypto module as an OpenSSL engine, then configure libsrtp to use OpenSSL for crypto and enable this engine.

JonathanLennox:
I'm happy to work on this -- I can look at reorganizing the header files.

I'd rather not go for the idea of plugging our kernel into OpenSSL -- that's much more heaviweight than our previous solution.

PR #201, opened 2016-09-01
Make replace_cipher and replace_auth public again

JonathanLennox:
This set of patches resolves Issue #200 -- it makes the replace_cipher and replace_auth functions public again, while excluding srtp crypto functions which should be internal from the published headers, and making sure all the symbols which are now made public are properly prefixed with "srtp_" prefixes.

There are some other cleanups along the way as well, hopefully explained in each patch. The patches should all be self-contained and self-explanatory.

### DIFF
--- a/crypto/include/auth.h
+++ b/crypto/include/auth.h
@@ -151,13 +151,13 @@ srtp_err_status_t srtp_auth_type_test(const srtp_auth_type_t *at,
 	const srtp_auth_test_case_t *test_data);
 
 /*
- * srtp_crypto_kernel_replace_auth_type(ct, id)
+ * srtp_replace_auth_type(ct, id)
  *
- * replaces the crypto kernel's existing cipher for the auth_type id
+ * replaces srtp's kernel's auth type implementation for the auth_type id
  * with a new one passed in externally.  The new auth type must pass all the
  * existing auth_type's self tests as well as its own.
  */
-srtp_err_status_t srtp_crypto_kernel_replace_auth_type(const srtp_auth_type_t *ct, srtp_auth_type_id_t id);
+srtp_err_status_t srtp_replace_auth_type(const srtp_auth_type_t *ct, srtp_auth_type_id_t id);
 
 #ifdef __cplusplus
 }
--- a/crypto/include/auth.h
+++ b/crypto/include/auth.h
@@ -150,6 +150,15 @@ srtp_err_status_t srtp_auth_type_self_test(const srtp_auth_type_t *at);
 srtp_err_status_t srtp_auth_type_test(const srtp_auth_type_t *at, 
 	const srtp_auth_test_case_t *test_data);
 
+/*
+ * srtp_crypto_kernel_replace_auth_type(ct, id)
+ *
+ * replaces the crypto kernel's existing cipher for the auth_type id
+ * with a new one passed in externally.  The new auth type must pass all the
+ * existing auth_type's self tests as well as its own.
+ */
+srtp_err_status_t srtp_crypto_kernel_replace_auth_type(const srtp_auth_type_t *ct, srtp_auth_type_id_t id);
+
 #ifdef __cplusplus
 }
 #endif
--- a/crypto/include/auth.h
+++ b/crypto/include/auth.h
@@ -53,42 +53,42 @@
 extern "C" {
 #endif
 
-typedef const struct srtp_auth_type_t *auth_type_pointer;
-typedef struct srtp_auth_t      *auth_pointer_t;
+typedef const struct srtp_auth_type_t *srtp_auth_type_pointer;
+typedef struct srtp_auth_t      *srtp_auth_pointer_t;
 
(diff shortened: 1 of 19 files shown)


crypto/include/auth.h
```diff
@@ -150,6 +150,15 @@ srtp_err_status_t srtp_auth_type_self_test(const srtp_auth_type_t *at);
...
 	const srtp_auth_test_case_t *test_data);
 
+/*
+ * srtp_crypto_kernel_replace_auth_type(ct, id)
+ *
+ * replaces the crypto kernel's existing cipher for the auth_type id
+ * with a new one passed in externally.  The new auth type must pass all the
+ * existing auth_type's self tests as well as its own.
+ */
+srtp_err_status_t srtp_crypto_kernel_replace_auth_type(const srtp_auth_type_t *ct, srtp_auth_type_id_t id);
```
pabuhler: Might have renamed it to  srtp_auth_replace_auth_type since it is now in the auth.h file  and crypto_kernel is private ish?

crypto/include/cipher.h
```diff
@@ -206,6 +206,15 @@ srtp_err_status_t srtp_cipher_decrypt(srtp_cipher_t *c, uint8_t *buffer, uint32_
...
 srtp_err_status_t srtp_cipher_set_aad(srtp_cipher_t *c, const uint8_t *aad, uint32_t aad_len);
 
+/*
+ * srtp_crypto_kernel_replace_cipher_type(ct, id)
+ *
+ * replaces the crypto kernel's existing cipher for the cipher_type id
+ * with a new one passed in externally.  The new cipher must pass all the
+ * existing cipher_type's self tests as well as its own.
+ */
+srtp_err_status_t srtp_crypto_kernel_replace_cipher_type(const srtp_cipher_type_t *ct, srtp_cipher_type_id_t id);
```
pabuhler: Might have renamed it to  srtp_cipher_replace_cipher_type since it is now in the auth.h file  and crypto_kernel is private ish?

pabuhler:
In general looks good, not sure what others will think but for me it is ok.
Wondered why you removed the debug module from the auth & cipher types? Replacement auth & ciphers will not have access to the debugging code and will need to any config of debugging and actual output on using their own framework?

crypto/include/cipher.h
JonathanLennox: In the subsequent patch I renamed the functions to srtp_replace_cipher_type and srtp_replace_auth_type.  I can squash the two patches together if you think it'd be clearer.

JonathanLennox:
I removed the debug module since I thought it should be part of libsrtp's internals, not part of the public API.

Revisiting, though, perhaps it should be part of the public API even on its own right -- right now there's no way for an application to enable SRTP debug output at all.

pabuhler:
Dont these functions work, I think maybe you have to compile with debugging available.
srtp_err_status_t srtp_set_debug_module(char *mod_name, int v);
srtp_err_status_t srtp_list_debug_modules(void);

We had talked about improving the debugging options, perhaps adding a application defined callback instead of stdout, but that is different to your use case of allowing a replacement cipher access to debug functions.

Personally I think having replacement ciphers using the internal debugging framework would be preferable but leave that up to you for now.

crypto/include/cipher.h
pabuhler: No thats fine, guess I missed that.

JonathanLennox:
The problem is that since srtp_err_reporting_set_level() isn't exposed anywhere (not even in a private header), srtp_err_level stays at srtp_err_level_none, so srtp_err_report always returns without printing anything.

JonathanLennox:
I've rebased this pull request as JonathanLennox/public_crypto_funcs_2, after my OpenSSL changes. No substantive changes to the patch. (I'm not sure what the etiquette is here for moving a pull request to a new branch.)

I'd like to converge on the debug functions issue. And does anyone have any other comments on this change?

In principle, I'd like to be able to install the auth.h and cipher.h headers into a srtp2/crypto subdirectory rather than having them be parallel with srtp2.h.  Unfortunately, it's hard to make this work given how the libsrtp source tree is laid out, unless I'm missing something.  Any thoughts?

pabuhler:
For the debug functions, you can remove it from public api for now. I have had a plan to add a debug callback api, as part of that I will ensure that debug actually works and then maybe we can add it back?

I think srtp2/crypto would be a great idea, but it does seam like we would have to change somethings to get it to work, maybe having parallel is ok.

JonathanLennox:
@pabuhler : was that an approval?

(Now that the visual studio build patches are in, my patch also now adds the cipher and auth functions to the DLL export list.)

pabuhler:
Yes, how to mark approved with out merging?

JonathanLennox:
Just say so in the comments, I have commit rights.

I'll merge the updated branch.

JonathanLennox:
Closed because #208 is the pull request for the updated branch.

PR #203, opened 2016-09-02
OpenSSL crypto performance improvements

JonathanLennox:
The way libsrtp is using OpenSSL crypto is inefficient, causing substantial overhead.

1. For both ICM and GCM, OpenSSL key expansion is being done in srtp_crypto_set_iv, which is called once per packet, rather than in srtp_crypto_init, which is called only once. This has substantial overhead.

2. HMAC-SHA1 isn't using EVP to get the SHA-1 implementation, and so it's not using the best platform-optimized SHA1 implementations.  It also rolls its own HMAC implementation rather than using OpenSSL's.

This patch series fixes both of these issues.  It is directly on top of Master.  Note that it conflicts (syntactically) with my pull request #201 -- see my branch JonathanLennox/libsrtp/ossl_perf_improvements_over_public_crypto for these patches rebased on top of that pull request.

On my MacBook Pro (2.3 GHz Haswell Core i7) this improves the performance of OpenSSL (as reported by srtp_driver -t) as follows.

# AES-128/HMAC-SHA-1:
* rtp cipher:    AES-128 counter mode using openssl
* rtp auth:      hmac sha-1 authentication function
* rtp services:  confidentiality and authentication

```
mesg length (octets)    old throughput (mbps)           new throughput
16                      154.287505                      341.843820
32                      300.592967                      640.592548
64                      557.309241                      1090.707682
128                     985.420777                      1891.567378
256                     1735.107978                     2880.976831
512                     2543.499050                     4033.401607
1024                    3794.384385                     4903.040460
2048                    4526.216918                     5648.019194
```

# AES-GCM-128:
* rtp cipher:    AES-128 GCM using openssl
* rtp auth:      null authentication function
* rtp services:  confidentiality and authentication

```
mesg length (octets)    old throughput (mbps)           new throughput
16                      234.320653                      665.626625
32                      444.591098                      1176.903273
64                      888.580354                      2244.137629
128                     1531.260748                     4354.852428
256                     3226.111339                     7588.557878
512                     5245.297033                     9858.714227
1024                    8371.655732                     15520.442575
2048                    12410.146871                    19804.900456
```

### DIFF
--- a/crypto/hash/hmac_ossl.c
+++ b/crypto/hash/hmac_ossl.c
@@ -102,6 +102,10 @@ static srtp_err_status_t srtp_hmac_dealloc (srtp_auth_t *a)
 
     HMAC_CTX_cleanup(hmac_ctx);
 
+    /* zeroize entire state*/
+    octet_string_set_to_zero((uint8_t*)a,
+                             sizeof(HMAC_CTX) + sizeof(srtp_auth_t));
+
     /* free memory */
     srtp_crypto_free(a);
 
--- a/crypto/hash/hmac_ossl.c
+++ b/crypto/hash/hmac_ossl.c
@@ -46,11 +46,12 @@
     #include <config.h>
 #endif
 
-#include "hmac.h"
+#include "auth.h"
 #include "alloc.h"
 #include <openssl/evp.h>
+#include <openssl/hmac.h>
 
-#define HMAC_KEYLEN_MAX		20
+#define SHA1_DIGEST_SIZE		20
 
 /* the debug module for authentiation */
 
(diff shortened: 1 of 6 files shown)


paulej:
Not having looked at the code, those numbers definitely look good. I agree in principle with minimizing the number of times we do key expansion.

crypto/hash/hmac_ossl.c
```diff
@@ -95,146 +88,90 @@ static srtp_err_status_t srtp_hmac_alloc (srtp_auth_t **a, int key_len, int out_
     (*a)->out_len = out_len;
     (*a)->key_len = key_len;
     (*a)->prefix_len = 0;
-    new_hmac_ctx = (srtp_hmac_ctx_t*)((*a)->state);
-    memset(new_hmac_ctx, 0, sizeof(srtp_hmac_ctx_t));
+    new_hmac_ctx = (HMAC_CTX*)((*a)->state);
```
pabuhler: Since state is no longer a srtp_hmac_ctx_t, can we remove the openssl values from the srtp_hmac_ctx_t struct?

pabuhler:
Looks good, please merge when ready

crypto/hash/hmac_ossl.c
JonathanLennox: Yes.

crypto/hash/hmac_ossl.c
```diff
@@ -95,146 +88,94 @@ static srtp_err_status_t srtp_hmac_alloc (srtp_auth_t **a, int key_len, int out_
...
 }
 
 static srtp_err_status_t srtp_hmac_dealloc (srtp_auth_t *a)
 {
-    srtp_hmac_ctx_t *hmac_ctx;
+    HMAC_CTX *hmac_ctx;
 
-    hmac_ctx = (srtp_hmac_ctx_t*)a->state;
-    if (hmac_ctx->ctx_initialized) {
-        EVP_MD_CTX_cleanup(&hmac_ctx->ctx);
```
jfigus: One of the reasons we introduced OpenSSL support for libsrtp was for FIPS compliance.  I may be wrong about this, but typically the OpenSSL EVP interface needs to be used for FIPS compliance.  Does the HMAC interface in OpenSSL work when FIPS is enabled?

JonathanLennox: In OpenSSL's FIPS User Guide https://www.openssl.org/docs/fips/UserGuide-2.0.pdf, the sample application using FIPS (page 136) uses HMAC_CTX.  So this should be fine.

PR #205, opened 2016-09-12
Update Windows build files to Visual Studio 2015

JonathanLennox:
The Visual Studio project files in git are ancient (Visual Studio 2008 maybe?)

These patches update to Visual Studio 2015, add the x64 platform, and adjust for the libsrtp 2.0.x changes (removed source files, changed list of exported symbols, etc.)

These project files should work on older Visual Studio versions back to Visual Studio 2010 with minor adjustments.


paulej:
That was needed. How did you handle integration with OpenSSL?  (No time to dig in right now; apologies if it's obvious.)

JonathanLennox:
I didn't handle OpenSSL integration at all, yet -- these project files still only use LibSRTP's internal crypto. (Accordingly, GCM isn't supported.)

Building OpenSSL on Windows is seriously non-trivial, so I was reluctant to try to get it to work.

My inclination is that on Windows it'd be better to integrate with Windows's native crypto implementation than try to get OpenSSL working.  (The primary difficulty with this would be that although Windows crypto has HMAC-SHA1 and AES-GCM implementations, it doesn't have AES-CTR (!), so we'd have to roll our own using ECB.)

paulej:
Ok, understood.  Want

paulej:
Accidentally closed the PR. Phones can be dangerous. I reopened it.

JonathanLennox:
Speaking of dangerous phones, your previous comment seems truncated, too.

paulej:
Built the code.  Works, so I'll merge the code. Would be nice if the test code would build.

Issue #150, opened 2016-02-25
Potential bug in SRTCP Key Derivation - clarification required


pabuhler:
It would seam that the KDF in libsrtp will always set the label in the same position regardless if it is RTP or RTCP. 

Reading the section 4.3.2 of the RFC it is possible to conclude that for RTCP the "index DIV kdr" should be a 32 bit value not the same 48 bit value used for RTP. If that was the case then for RTCP the label should be set in the IV at position 9 not a position 7 as is currently done libsrtp.

For me the text is not entirely clear in the RFC and I would have thought that the label was required to be in the same position always for both RTP and RTCP when using the same master key and salt to ensure unique session keys.

So is it a bug in libsrtp or is the text not clear?

jfigus:
RFC 3711 section 4.3.2 states:
```
Replace the SRTP index by the 32-bit quantity: 0 || SRTCP index
   (i.e., excluding the E-bit, replacing it with a fixed 0-bit), and use
   <label> = 0x03 for the SRTCP encryption key, <label> = 0x04 for the
   SRTCP authentication key, and, <label> = 0x05 for the SRTCP salting
   key.

```
Clearly the SRTCP index is 32 bits.  Section 4.3.1 in RFC 3711 states the following:

```
We also make the
   convention of treating "a DIV t" as a bit string of the same length
   as a, and thus "a DIV t" will in general have leading zeros.
```
Therefore, calculating r for SRTCP would result in a 32-bit value.  The KDF algorithm then concatenates r with the label, forming a 40-bit value (compared to a 56-bit values used by SRTP as seen in appendix B.3).

This looks to be a bug that's probably been in libsrtp since inception.  Fixing this will be a significant backwards compatibility hassle.  It would be good to hear from the libsrtp historical maintainers or RFC authors.  Is submitting an errata to RFC 3711 an easier way to address this?

jesup:
I'm a historical maintainer; breaking compatibility with existing uses of SRTP would be a nightmare IMHO.

What are our options?  We could somehow negotiate this, to allow for backwards compatibility.  I doubt we can make it compatible "by default".  Negotiation would be quite a massive pain, so if we revise/clarify the text to match the existing implementation(s), what would be the impact, especially security-wise?

jfigus:
[...]

PR #217, opened 2016-11-04
Added workaroud for bug #150

Bendr0id:
Added workaroud for "Potential bug in SRTCP Key Derivation - clarification required #150". 
With this workaround the srtp_kdf_generate method will calculcate the correct keys for SRTCP.


pabuhler:
Turns out the behavior reported in #150 has already been addressed as an Errata an so therefore requires no changes.
https://www.rfc-editor.org/errata_search.php?rfc=3711&eid=3712

PR #223, opened 2016-12-13
Limited MKI support

darkangel-ua:
This is basically Per Cederqvist patch that adds MKI support. I just adopted it to current master state and removed user visible changes to policy objects. It adds limited MKI support - only one key used from array. When someone will add fully functional MKI support than API changes will be ok, but right no they are not even required.


pabuhler:
How does this relate to #224 ? Can we close this and instead focus on that pull request?

darkangel-ua:
Yes, I think we can close this. Anyway #224 has support for multiple keys.

PR #224, opened 2016-12-15
Apply MKI Support Patch

rsith71:
This PR provides support of Multiple Master Keys via MKI.  I only included up to four master keys since that is what the SRTP stack I was coming from supported.  The unit test are passing with these changes.


ulfolsson:
If you build with openssl enabled i.e ./configure --enable-openssl the following test fails

testing srtp_protect_rtp and srtp_unprotect_rtp with MKI index set to 0
# SSRC:          any outbound
# rtp cipher:    AES-128 GCM using openssl
# rtp auth:      null authentication function
# rtp services:  confidentiality and authentication
# rtcp cipher:   AES-128 GCM using openssl
# rtcp auth:     null authentication function
# rtcp services: authentication
# window size:   128
# tx rtx allowed:false
# Encrypted extension headers: none
testing that ciphertext is distinct from plaintext...passed
error: unexpected srtp failure (code 25)

rsith71:
Thank you for pointing that out.  I missed a line of code when I patched it manually.

ulfolsson:
I have done some more testing and I found a minor memory leak (attached patch)
[patch.zip](https://github.com/cisco/libsrtp/files/698180/patch.zip)

The I have a question about the API. Lets say I am a receiver and

1.  When the stream starts I have knowledge about one key/MKI i.e policy.num_master_keys = 1

2. Later on I get signaled to use a new key/MKI i.e now I'm supposed to use two keys/MKI:s

How do you think this should work?

pabuhler:
@rsith71  could you update this to include the patch from the previous comment?

rsith71:
@ulfolsson Thank you for the patch and finding the memory leak.

So currently the application that was using this code was tearing down the sessions and bringing new ones up.  This though leaves the time of the teardown and new session establishment vulnerable to a replay attack.  I have been trying to think about a possible solution and haven't really came up with one.  

I am completely open to suggestions here for updating.  Maybe some new APIs are needed to update/add new master keys to a session.

pabuhler:
in master there is now a srtp_update() function that is meant to be used to change master key on existing streams, it will preserve ROC and replay index. Can this also be used for updating mki keys?

rsith71:
@pabuhler I try out the srtp_update later today.  Thank you for pointing it out.

rsith71:
@ulfolsson I took a look at srtp_update.  It looks like this would recreate the keys for each stream.   I think this is the correct approach to handle when you receive a new master key.  You will need to add the new master key to the list of master keys in your policy.  Then you can call srtp_update to create the new session keys for that new master key.

ulfolsson:
Yes, it seems to do the trick. Thank's for your input

srtp/srtp.c
```diff
@@ -193,56 +214,74 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
...
 
-  /* allocate auth function */
-  stat = srtp_crypto_kernel_alloc_auth(p->rtp.auth_type, 
-				  &str->rtp_auth,
-				  p->rtp.auth_key_len, 
-				  p->rtp.auth_tag_len); 
-  if (stat) {
-    srtp_stream_free(str);
-    return stat;
+  str->session_keys = (srtp_session_keys_t *)srtp_crypto_alloc(sizeof(srtp_session_keys_t) * str->num_master_keys);
```
paulej: This function does not set memory to zero.  I would suggest doing a memset() here so that, in case there is a failure, the data structure will be zerod and the srtp_stream_free() will not misinterpret random values are valid values.

include/srtp_priv.h
```diff
@@ -107,21 +135,16 @@ typedef enum direction_t {
...
-  srtp_auth_t    *rtp_auth;
+  srtp_session_keys_t *session_keys;
+  unsigned int num_master_keys;
   srtp_rdbx_t     rtp_rdbx;
   srtp_sec_serv_t rtp_services;
-  srtp_cipher_t  *rtcp_cipher;
-  srtp_auth_t    *rtcp_auth;
   srtp_rdb_t      rtcp_rdb;
   srtp_sec_serv_t rtcp_services;
   srtp_key_limit_ctx_t *limit;
```
paulej: Curious as to why this wasn't also moved to srtp_session_keys_t.  Is the thinking that once the key limit is reached, then we consider ALL keys invalid?

srtp/srtp.c
```diff
@@ -3473,24 +3785,37 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
       return srtp_err_status_no_ctx;
     } 
   }
-  
+
+  /*
+   * Determine if MKI is being used and what the MKI index is
```
paulej: This returns the session keys given the MKI found in the packet.  It does not determine an MKI index, as the comment states.

srtp/srtp.c
```diff
@@ -3529,8 +3854,14 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
    * (note that srtcp *always* uses authentication, unlike srtp)
    */
   auth_start = (uint32_t *)hdr;
-  auth_len = *pkt_octet_len - tag_len;
-  auth_tag = (uint8_t *)hdr + auth_len;
+
+  /*
+   * The location of the auth tag in the packet needs to know MKI ID 
+   * could be present.  The data needed to calculate the Auth tag
+   * must not include the MKI ID
```
paulej: ID after MKI is redundant :)

paulej:
BTW, I noted that there are a number of changes here that will conflict with the changes for AES-192.  It might be good to allow us to commit the other code first, if you don't mind.  Either way, somebody has to resolve conflicts. :)

rsith71:
@paulej I can resolve the conflicts with AES-192.  I probably wont have time to address your comments until Wed, Thur this week.

paulej:
@rsith71 No worries.  This is perhaps the biggest commit, so take whatever time you need to ensure it's right.

srtp/srtp.c
```diff
@@ -193,56 +214,74 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
...
-  /* allocate cipher */
-  stat = srtp_crypto_kernel_alloc_cipher(p->rtp.cipher_type, 
-				    &str->rtp_cipher, 
-				    p->rtp.cipher_key_len,
-				    p->rtp.auth_tag_len); 
-  if (stat) {
-    srtp_stream_free(str);
-    return stat;
+  /* To keep backwards API compatible if someone is using multiple master
+   * keys then key should be set to NULL
```
rsith71: This is referring to the Policy key passed into SRTP.  The end user of SRPT if he/she is using MKI then they must set the Policy Key to NULL.

srtp/srtp.c
rsith71: Thanks.  Done should see fix in next Push.

srtp/srtp.c
rsith71: Fixed.  Should see fix in next push

srtp/srtp.c
rsith71: Fixed. Should see fix in next push.

include/srtp_priv.h
rsith71: This was something I missed.  I imagine we would want to keep track of a limit per Master Key.  Probably does need to be moved to the session context.  This may take a bit longer then the other comments.  I'll start working on it tonight.

rsith71: I think I have the changes.  Is there a specific test for the limit code?  I ran all the test and everything passes.  I am going to push the changes but if you know of a specific limit test that is being run that will make me have greater confidence in the quick change I just made.

paulej: Not sure if there is a test in the test/ directory for this or not.

paulej:
@rsith71 Thanks for the update.  Everything looks good and tests pass.

I did have one last question.  Is there a reason for putting the mki_length with the srtp_session_keys_t and srtp_master_key_t structures, versus the stream context (srtp_stream_ctx_t_)?  As I understand it, the way the code is currently written, it assumes that every single key will have the same MKI length repeated over and over.  In srtp_get_session_keys(), for example, it does not inspect the SSRC to determine which stream context to use to then determine an MKI length.

Per RFC 3711, "the value of the MKI indicator and length MUST be kept fixed for the lifetime of the context".  This suggests that, in theory, the MKI length could differ based on the SSRC (stream), just as the cipher varies.  I think the way it's coded, that can be accommodated, but with a lot of repetition.

So while it's not wrong, I think it could open the door to abuse.  One could create 4 keys, each with a different MKI length.  That would be a user error.  But, if we force the MKI length on the stream, that might prevent such abuse.

I have no strong feeling one way or the other, but I would like to understand your view on this.

include/srtp.h
```diff
@@ -1484,6 +1697,18 @@ srtp_err_status_t srtp_set_debug_module(char *mod_name, int v);
...
+/**
+ * @brief srtp_get_protect_data_added_length(policy, use_mki, mki_index, is_rtcp, length)
+ *
+ * Determines the length of the amount of data Lib SRTP will add to the 
+ * packet during the protect process. The length is returned in the length parameter
+ *
+ * returns err_status_ok on success, err_status_bad_mki if the MKI index is invalid
+ *
+ */
+srtp_err_status_t srtp_get_protect_data_added_length(const srtp_policy_t *policy, uint32_t use_mki,
```
pabuhler: I can see why having a more dynamic way of determining the required size of the buffer could be useful especially with a possible mki size of 128 bytes but is it really needed? Why not just change the definition of SRTP_MAX_TRAILER_LEN to be SRTP_MAX_TAG_LEN + SRTP_MAX_MKI_LEN?  If you look at the current documentation for SRTP_MAX_TRAILER_LEN, it already states that it will include the auth tag and the mki size. 

If it is decided that this functions is far more useful then perhaps remove the SRTP_MAX_TRAILER_LEN define altogether and instead rename this to srtp_get_protect_trailer_length, in which case I would also remove the is_rtcp argument and have a second function called srtp_get_protect_rtcp_trailer_length() keeping it in line with the rest of the api.  Also why operate on a policy rather than a session?

pabuhler:
I have a general question about the need for the new apis:
```
srtp_protect_mki()
srtp_protect_rtcp_mki()
srtp_unprotect_mki()
srtp_unprotect_rtcp_mki()
``` 
are these really needed? As I understand a session/stream is either configured to use MKI or it is not, in which case the `unsigned int use_mki` argument to these functions is redundant or did I miss something?

I see the need to know the current mki_index when protecting a packet, but is that something that is likely to change often, per packet? Could an alternative be to have a srtp_set_current_mki_index/id(session/ssrc) function that was called when needed.

If there is a valid use case for adding these function then please explain a bit.

rsith71:
@pabuhler You need to also specify which MKI key you are using.  Since you can support multiple keys and each packet could potential being used by a different Master key.  I guess you could say you could remove the unprotect functions since you can have the use_mki variable inside the context.  I guess you could also move an field inside the context about which Master key you would use, but you still need an API to update that field when you want to use another master key.

include/srtp.h
rsith71: This is more of a coding style.  I don't want to allocate memory that isn't needed.  If you are an appliance (non endpoint) you may have thousands of calls going through you and being terminated.  I know memory is cheap today, but still this is my coding preference.

I understand the push to a session.  That would probably make more sense then a policy.

pabuhler:
@rsith71 do you see that uses would change the MKI index on a per packet bases? Having to call a separate function to set the MKI index before protecting each packet would be a annoying so it is really up to the frequency of MKI index changes.

rsith71:
@pabuhler  :) Since this is a library for applications to use I see a need to be flexible.  I agree it seems not probably that a master key would change very often.  Then again I'm only looking at one application that we are trying to use libsrtp instead of the existing SRTP stack which had MKI support.  Thus libSRTP needed MKI support hence this patch.  The application's master key doesn't change very often if at all.  The existing code always used the same master key for encryption.  It was really the decryption part where you need the flexibility to handle different master keys since you do not know what the other end is doing.

That being said if an application using MKI wants to switch between master keys for every packet that is there choice.  The library should be flexible to handle that choice.

rsith71:
@pabuhler All being said I am find changing the code to be what you guys want it to be.

pabuhler:
Flexibility is good, I just have a feeling it doesn't scale well as an api. In the EKT dev branch they have introduced a srtp_protect_with_flags() function maybe it wold be possible to reuse that concept. Can see if others have an opinion.

paulej:
I think having the srtp_protect_mki() (and family of APIs) is OK.  Existing applications are not impacted, since srtp_protect() is still callable.  The srtp_protect_with_flags() does offer a more flexible way to handle these kinds of things.  Here's a link to the code for that:
https://github.com/cisco/libsrtp/blob/2_0_0_ekt_dev/include/srtp.h#L212
https://github.com/cisco/libsrtp/blob/2_0_0_ekt_dev/include/srtp.h#L630

However, there is no way to pass in the MKI value.  A possible way to avoid passing this in per call to srtp_protect_mki() is to select the MKI index via a separate API.  For example: srtp_select_mki().  This might 
then store that MKI value in the crypto context and remain the MKI value to use until explicitly set again.  Is it likely that we would change from one MKI value to another frequently?  I would assume a single MKI will be used most of the time, changing very rarely.

On receipt of a packet, the MKI value is looked up in the packet, though the SSRC is not considered in that lookup.  Would we ever see MKI values of different lengths in the same SRTP session?  If so, we should perhaps associate the MKI values with the stream (and specific SSRC).   If not, then why replicate the length over-and-over separately for each key?

srtp/srtp.c
```diff
@@ -1759,9 +2018,9 @@ srtp_unprotect_aead (srtp_ctx_t *ctx, srtp_stream_ctx_t *stream, int delta,
 #else
      iv.v64[1] = be64_to_cpu(est << 16);
 #endif
-     status = srtp_cipher_set_iv(stream->rtp_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
-     if (!status && stream->rtp_xtn_hdr_cipher) {
-       status = srtp_cipher_set_iv(stream->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
+     status = srtp_cipher_set_iv(session_keys->rtp_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
+     if (!status && session_keys->rtp_xtn_hdr_cipher) {
+       status = srtp_cipher_set_iv(session_keys->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1759,9 +2018,9 @@ srtp_unprotect_aead (srtp_ctx_t *ctx, srtp_stream_ctx_t *stream, int delta,
 #else
      iv.v64[1] = be64_to_cpu(est << 16);
 #endif
-     status = srtp_cipher_set_iv(stream->rtp_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
-     if (!status && stream->rtp_xtn_hdr_cipher) {
-       status = srtp_cipher_set_iv(stream->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
+     status = srtp_cipher_set_iv(session_keys->rtp_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
```
thisisG: Nit: Linebreak

srtp/srtp.c
```diff
@@ -1774,9 +2033,9 @@ srtp_unprotect_aead (srtp_ctx_t *ctx, srtp_stream_ctx_t *stream, int delta,
      iv.v64[0] = 0;
 #endif
      iv.v64[1] = be64_to_cpu(est);
-     status = srtp_cipher_set_iv(stream->rtp_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
-     if (!status && stream->rtp_xtn_hdr_cipher) {
-       status = srtp_cipher_set_iv(stream->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
+     status = srtp_cipher_set_iv(session_keys->rtp_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1774,9 +2033,9 @@ srtp_unprotect_aead (srtp_ctx_t *ctx, srtp_stream_ctx_t *stream, int delta,
...
+     if (!status && session_keys->rtp_xtn_hdr_cipher) {
+       status = srtp_cipher_set_iv(session_keys->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_encrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1940,23 +2211,35 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
 #endif
 
   /*
+   * Determine if MKI is being used and what session keys should be used
+   */
+  if (use_mki) {
+      session_keys = srtp_get_session_keys(stream, (uint8_t *)hdr, (const unsigned int*)pkt_octet_len, &mki_size);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1856,12 +2115,22 @@ srtp_unprotect_aead (srtp_ctx_t *ctx, srtp_stream_ctx_t *stream, int delta,
...
 }
 
 
 srtp_err_status_t
 srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
+    return srtp_unprotect_mki(ctx, srtp_hdr, pkt_octet_len, 0);
+}
+
+srtp_err_status_t
+srtp_unprotect_mki(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len, unsigned int use_mki) {
```
thisisG: 'size_t' or 'uint32_t' for length - whichever makes sense in this context.

srtp/srtp.c
```diff
@@ -1875,6 +2144,8 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
   uint8_t tmp_tag[SRTP_MAX_TAG_LEN];
   uint32_t tag_len, prefix_len;
   srtp_hdr_xtnd_t *xtn_hdr = NULL;
+  unsigned int mki_size = 0;
```
thisisG: 'size_t' or 'uint32_t' for length - whichever makes sense in this context.

srtp/srtp.c
```diff
@@ -1940,23 +2211,35 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
...
+  /*
    * Check if this is an AEAD stream (GCM mode).  If so, then dispatch
    * the request to our AEAD handler.
    */
-  if (stream->rtp_cipher->algorithm == SRTP_AES_128_GCM ||
-      stream->rtp_cipher->algorithm == SRTP_AES_256_GCM) {
-      return srtp_unprotect_aead(ctx, stream, delta, est, srtp_hdr, (unsigned int*)pkt_octet_len);
+  if (session_keys->rtp_cipher->algorithm == SRTP_AES_128_GCM ||
+      session_keys->rtp_cipher->algorithm == SRTP_AES_256_GCM) {
+      return srtp_unprotect_aead(ctx, stream, delta, est, srtp_hdr, (unsigned int*)pkt_octet_len, session_keys, mki_size);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1967,9 +2250,9 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
 #else
     iv.v64[1] = be64_to_cpu(est << 16);
 #endif
-    status = srtp_cipher_set_iv(stream->rtp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
-    if (!status && stream->rtp_xtn_hdr_cipher) {
-      status = srtp_cipher_set_iv(stream->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
+    status = srtp_cipher_set_iv(session_keys->rtp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1967,9 +2250,9 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
...
+    if (!status && session_keys->rtp_xtn_hdr_cipher) {
+      status = srtp_cipher_set_iv(session_keys->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1981,9 +2264,9 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
     iv.v64[0] = 0;
 #endif
     iv.v64[1] = be64_to_cpu(est);
-    status = srtp_cipher_set_iv(stream->rtp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
-    if (!status && stream->rtp_xtn_hdr_cipher) {
-      status = srtp_cipher_set_iv(stream->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
+    status = srtp_cipher_set_iv(session_keys->rtp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -1981,9 +2264,9 @@ srtp_unprotect(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len) {
...
+    if (!status && session_keys->rtp_xtn_hdr_cipher) {
+      status = srtp_cipher_set_iv(session_keys->rtp_xtn_hdr_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -2467,7 +2753,7 @@ update_template_streams(srtp_t session, const srtp_policy_t *policy) {
     srtp_rdb_t old_rtcp_rdb;
 
     stream = session->stream_list;
-    while ((stream != NULL) && (stream->rtp_auth != session->stream_template->rtp_auth)) {
+    while ((stream != NULL) && (stream->session_keys[0].rtp_auth != session->stream_template->session_keys[0].rtp_auth)) {
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -2286,7 +2572,7 @@ srtp_add_stream(srtp_t session,
   srtp_stream_t tmp;
 
   /* sanity check arguments */
-  if ((session == NULL) || (policy == NULL) || (policy->key == NULL))
+  if ((session == NULL) || (policy == NULL) || (!srtp_validate_policy_master_keys(policy)))
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -2419,7 +2705,7 @@ srtp_update(srtp_t session, const srtp_policy_t *policy) {
   srtp_err_status_t stat;
 
   /* sanity check arguments */
-  if ((session == NULL) || (policy == NULL) || (policy->key == NULL)) {
+  if ((session == NULL) || (policy == NULL) || (!srtp_validate_policy_master_keys(policy))) {
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -2933,6 +3221,8 @@ srtp_protect_rtcp_aead (srtp_t ctx, srtp_stream_ctx_t *stream,
         *trailer = 0x00000000; /* set encrypt bit */
     }
 
+    mki_size = srtp_inject_mki((uint8_t *)hdr + *pkt_octet_len + tag_len + sizeof(srtcp_trailer_t), session_keys, use_mki);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -2895,7 +3181,8 @@ static void srtp_calc_aead_iv_srtcp(srtp_stream_ctx_t *stream, v128_t *iv,
  */
 static srtp_err_status_t
 srtp_protect_rtcp_aead (srtp_t ctx, srtp_stream_ctx_t *stream, 
-                        void *rtcp_hdr, unsigned int *pkt_octet_len)
+                        void *rtcp_hdr, unsigned int *pkt_octet_len,
```
thisisG: `size_t` or `uint32_t` for length, whichever makes sense in this context.

srtp/srtp.c
```diff
@@ -2907,9 +3194,10 @@ srtp_protect_rtcp_aead (srtp_t ctx, srtp_stream_ctx_t *stream,
     uint32_t seq_num;
     v128_t iv;
     uint32_t tseq;
+    unsigned int mki_size = 0;
```
thisisG: `size_t` or `uint32_t` for length, whichever makes sense in this context.

srtp/srtp.c
```diff
@@ -3043,7 +3336,8 @@ srtp_protect_rtcp_aead (srtp_t ctx, srtp_stream_ctx_t *stream,
  */
 static srtp_err_status_t
 srtp_unprotect_rtcp_aead (srtp_t ctx, srtp_stream_ctx_t *stream, 
-                          void *srtcp_hdr, unsigned int *pkt_octet_len)
+                          void *srtcp_hdr, unsigned int *pkt_octet_len,
```
thisisG: `size_t` for length?

srtp/srtp.c
```diff
@@ -3056,9 +3350,14 @@ srtp_unprotect_rtcp_aead (srtp_t ctx, srtp_stream_ctx_t *stream,
     uint32_t seq_num;
     v128_t iv;
     uint32_t tseq;
+    unsigned int mki_size = 0;
```
thisisG: `uint32_t`?

srtp/srtp.c
```diff
@@ -3215,6 +3514,11 @@ srtp_unprotect_rtcp_aead (srtp_t ctx, srtp_stream_ctx_t *stream,
 
 srtp_err_status_t 
 srtp_protect_rtcp(srtp_t ctx, void *rtcp_hdr, int *pkt_octet_len) {
+  return srtp_protect_rtcp_mki(ctx, rtcp_hdr, pkt_octet_len, 0, 0);
+}
+
+srtp_err_status_t 
+srtp_protect_rtcp_mki(srtp_t ctx, void *rtcp_hdr, int *pkt_octet_len, unsigned int use_mki, unsigned int mki_index) {
```
thisisG: `uint32_t` for index?
Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3226,6 +3530,8 @@ srtp_protect_rtcp(srtp_t ctx, void *rtcp_hdr, int *pkt_octet_len) {
   srtp_stream_ctx_t *stream;
   uint32_t prefix_len;
   uint32_t seq_num;
+  unsigned int mki_size = 0;
```
thisisG: `uint32_t`?

srtp/srtp.c
```diff
@@ -3277,17 +3583,19 @@ srtp_protect_rtcp(srtp_t ctx, void *rtcp_hdr, int *pkt_octet_len) {
...
   /*
    * Check if this is an AEAD stream (GCM mode).  If so, then dispatch
    * the request to our AEAD handler.
    */
-  if (stream->rtp_cipher->algorithm == SRTP_AES_128_GCM ||
-      stream->rtp_cipher->algorithm == SRTP_AES_256_GCM) {
-      return srtp_protect_rtcp_aead(ctx, stream, rtcp_hdr, (unsigned int*)pkt_octet_len);
+  if (session_keys->rtp_cipher->algorithm == SRTP_AES_128_GCM ||
+      session_keys->rtp_cipher->algorithm == SRTP_AES_256_GCM) {
+      return srtp_protect_rtcp_aead(ctx, stream, rtcp_hdr, (unsigned int*)pkt_octet_len, session_keys, use_mki);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3312,13 +3620,15 @@ srtp_protect_rtcp(srtp_t ctx, void *rtcp_hdr, int *pkt_octet_len) {
     *trailer = 0x00000000;     /* set encrypt bit */    
   }
 
+  mki_size = srtp_inject_mki((uint8_t *)hdr + *pkt_octet_len + sizeof(srtcp_trailer_t), session_keys, use_mki);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3405,13 +3715,21 @@ srtp_protect_rtcp(srtp_t ctx, void *rtcp_hdr, int *pkt_octet_len) {
...
 }
 
 
 srtp_err_status_t 
 srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
+    return srtp_unprotect_rtcp_mki(ctx, srtcp_hdr, pkt_octet_len, 0);
+}
+
+srtp_err_status_t 
+srtp_unprotect_rtcp_mki(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len, unsigned int use_mki) {
```
thisisG: `size_t` for length?

srtp/srtp.c
```diff
@@ -3428,6 +3746,8 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
   uint32_t seq_num;
   int e_bit_in_packet;     /* whether the E-bit was found in the packet */
   int sec_serv_confidentiality; /* whether confidentiality was requested */
+  unsigned int mki_size = 0;
```
thisisG: `uint32_t`?

srtp/srtp.c
```diff
@@ -3472,24 +3792,37 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
       return srtp_err_status_no_ctx;
     } 
   }
-  
+
+  /*
+   * Determine if MKI is being used and what session keys should be used
+   */
+  if (use_mki) {
+      session_keys = srtp_get_session_keys(stream, (uint8_t *)hdr, (const unsigned int*)pkt_octet_len, &mki_size);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3472,24 +3792,37 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
...
   /*
    * Check if this is an AEAD stream (GCM mode).  If so, then dispatch
    * the request to our AEAD handler.
    */
-  if (stream->rtp_cipher->algorithm == SRTP_AES_128_GCM ||
-      stream->rtp_cipher->algorithm == SRTP_AES_256_GCM) {
-      return srtp_unprotect_rtcp_aead(ctx, stream, srtcp_hdr, (unsigned int*)pkt_octet_len);
+  if (session_keys->rtp_cipher->algorithm == SRTP_AES_128_GCM ||
+      session_keys->rtp_cipher->algorithm == SRTP_AES_256_GCM) {
+      return srtp_unprotect_rtcp_aead(ctx, stream, srtcp_hdr, (unsigned int*)pkt_octet_len, session_keys, mki_size);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3472,24 +3792,37 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
...
+
   /* get tag length from stream context */
-  tag_len = srtp_auth_get_tag_length(stream->rtcp_auth);
+  tag_len = srtp_auth_get_tag_length(session_keys->rtcp_auth);
 
   /* check the packet length - it must contain at least a full RTCP
      header, an auth tag (if applicable), and the SRTCP encrypted flag
      and 31-bit index value */
-  if (*pkt_octet_len < (int) (octets_in_rtcp_header + tag_len + sizeof(srtcp_trailer_t))) {
+  if (*pkt_octet_len < (int) (octets_in_rtcp_header + tag_len + mki_size + sizeof(srtcp_trailer_t))) {
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3559,14 +3898,14 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
...
-  if (stream->rtcp_cipher->type->id == SRTP_AES_ICM) {
+  if (session_keys->rtcp_cipher->type->id == SRTP_AES_ICM) {
     v128_t iv;
 
     iv.v32[0] = 0;
     iv.v32[1] = hdr->ssrc; /* still in network order! */
     iv.v32[2] = htonl(seq_num >> 16);
     iv.v32[3] = htonl(seq_num << 16);
-    status = srtp_cipher_set_iv(stream->rtcp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
+    status = srtp_cipher_set_iv(session_keys->rtcp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3576,17 +3915,17 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
     iv.v32[1] = 0;
     iv.v32[2] = 0;
     iv.v32[3] = htonl(seq_num);
-    status = srtp_cipher_set_iv(stream->rtcp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
+    status = srtp_cipher_set_iv(session_keys->rtcp_cipher, (uint8_t*)&iv, srtp_direction_decrypt);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3614,14 +3953,17 @@ srtp_unprotect_rtcp(srtp_t ctx, void *srtcp_hdr, int *pkt_octet_len) {
 
   /* if we're decrypting, exor keystream into the message */
   if (enc_start) {
-    status = srtp_cipher_decrypt(stream->rtcp_cipher, (uint8_t *)enc_start, &enc_octet_len);
+    status = srtp_cipher_decrypt(session_keys->rtcp_cipher, (uint8_t *)enc_start, &enc_octet_len);
```
thisisG: Nit:
Linebreak

srtp/srtp.c
```diff
@@ -3820,6 +4162,33 @@ srtp_profile_get_master_salt_length(srtp_profile_t profile) {
   }
 }
 
+srtp_err_status_t
+srtp_get_protect_data_added_length(const srtp_policy_t *policy,
+                                   uint32_t use_mki,
+                                   uint32_t mki_index,
```
thisisG: Here it is `uint32_t` :) Other places it is `unsigned int`. I think the C99(ish) types are generally preferred.

test/util.c
```diff
@@ -182,4 +182,3 @@ int base64_string_to_octet_string (char *out, int *pad, char *in, int len)
     *pad = j;
     return i;
 }
```
thisisG: Nitty nit:
\n!

pabuhler:
@paulej & @rsith71  ok lets stay with the xxx_mki() functions . I still think it would be better to try and combine the functions but since this is a new feature keeping it isolated is perhaps a good idea. Maybe it is better to wait for ekt to be completed and can look at making a consolidated api in a v3.0 .

include/srtp.h
pabuhler: Sounds fair.
I would still prefer SRTP_MAX_TRAILER_LEN  to be defined as SRTP_MAX_TAG_LEN + SRTP_MAX_MKI_LEN since it was previously documented to do so.

Also having two functions that operate on session to match the rest of the API.
srtp_get_protect_trailer_length()
srtp_get_protect_rtcp_trailer_length()

rsith71:
@thiisG If you want to push the changes you can go ahead and do so.  If not I can take a look at fixing up the issues.

rsith71:
@paulej The question about the MKI length being kept at the session keys.  The reason I did that is when written the code thought it was possible for each master key index to have different sizes.  Thus when trying to determine the MKI of the encrypted packet I needed the length.  See function srtp_get_session_keys.  Since I was planning on each MKI to have it's own size I figured it was better to associate it with the session keys instead of having another array to manage.

If you want to restrict it to be one MKI length per Context/stream then I can easily change the code.

paulej:
@rsith71 I think keeping a didn't length per key is OK. You're right that it allows for more flexibility, and that might be needed.  RFC 3711 specifies that the MKI length for a stream is constant for the duration of a stream, though. It's a transform independent parameter like the ROC.

Allowing for varying MKI length in the code is OK, but I don't recall seeing any code that looks up that length.  When a packet with SSRC 1 arrives, does the code look for the MKI length for SSRC 1? And if the next packet in the session is SSRC 2 (having a different MKI length than SSRC 1), is that length looked up?

Perhaps it's there and I just overlooked it.

rsith71:
@paulej So I have Master keys which are used for multiple SSRCs if the context is setup that way.  Since that is the case they will have the same MKI information between multiple SSRCs.  Therefore I do not have any code that checks per an SSRC bases.  If there should be MKI info per SSRC then this code needs to be reworked.  As I understand a context can have multiple SSRCs.  It was my understanding that the MKI info is associated with the context and not the SSRCs.  If that is incorrect please let me know.

paulej:
@rsith71 One key can be used for multiple SSRC values.  However, you also want to allow varying MKI length values that are used to select this master key.  That's where I'm confused.  There isn't a single master key with MKI, but an index into an set of keys.  If the MKI length was always constant, no problem.  But, since I can set one MKI length to be 1 octet and the next to be 500, don't you need to know this before digging too far into the packet?

paulej:
That said, I guess one could argue that correlating an SSRC to an MKI length might just create more confusion since, as you noted, multiple SSRCs might exist on a flow, each of which using the same master key (and MKI).

I think that might be why the designers originally put restrictions in place to say the MKI length within a crypto context is fixed in size. So, we could leave the code as-is and allow users to hang themselves. Or, we could just put the MKI length inside the crypto context somewhere. (Perhaps have a default length of 0, meaning MKI is not used. Then, as each key / MKI value is introduced, the first one sets the MKI value and the others must be the same size. A different size just gets rejected.)

I have no strong preference. Perhaps I should yield and allow others to provide an opinion. :)

thisisG:
@rsith71 - if you could do a fixup the ones I marked for clearing secrets and checking for NULL that would be great since you've a more intimate knowledge about the changes and might easier spot what return values and such might be more appropriate in case of errors.

I will then do a drive-by fixup for formatting before a merge, it is pretty much automated so no problem for me to do it! Waiting until the API is finalised and the PR is accepted otherwise. :-)

srtp/srtp.c
```diff
@@ -193,55 +215,73 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
...
-    return stat;
-  }
-  
-  /* allocate key limit structure */
-  str->limit = (srtp_key_limit_ctx_t*) srtp_crypto_alloc(sizeof(srtp_key_limit_ctx_t));
-  if (str->limit == NULL) {
-    srtp_stream_free(str);
-    return srtp_err_status_alloc_fail;
-  }
+  str->session_keys = (srtp_session_keys_t *)srtp_crypto_alloc(sizeof(srtp_session_keys_t) * str->num_master_keys);
```
rsith71: Done Will be fixed in next Push

srtp/srtp.c
```diff
@@ -299,74 +347,106 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
...
+      status = srtp_auth_dealloc(session_keys->rtcp_auth);
+      if (status)
+        return status;
+    }
+
+    /*
+     * zeroize the salt value
+     */
+    memset(session_keys->salt, 0, SRTP_AEAD_SALT_LEN);
+    memset(session_keys->c_salt, 0, SRTP_AEAD_SALT_LEN);
```
rsith71: Done will be fixed in next pushed

srtp/srtp.c
```diff
@@ -299,74 +347,106 @@ srtp_stream_alloc(srtp_stream_ctx_t **str_ptr,
...
+
+    if (session_keys->mki_id) {
+      memset(session_keys->mki_id, 0, session_keys->mki_size);
```
rsith71: Done Will be there in next push.

srtp/srtp.c
```diff
@@ -419,21 +495,46 @@ srtp_stream_clone(const srtp_stream_ctx_t *stream_template,
...
   *str_ptr = str;  
 
-  /* set cipher and auth pointers to those of the template */
-  str->rtp_cipher  = stream_template->rtp_cipher;
-  str->rtp_auth    = stream_template->rtp_auth;
-  str->rtp_xtn_hdr_cipher = stream_template->rtp_xtn_hdr_cipher;
-  str->rtcp_cipher = stream_template->rtcp_cipher;
-  str->rtcp_auth   = stream_template->rtcp_auth;
+  str->num_master_keys = stream_template->num_master_keys;
+  str->session_keys = (srtp_session_keys_t *)srtp_crypto_alloc(sizeof(srtp_session_keys_t) * str->num_master_keys);
```
rsith71: Done will be in next push

srtp/srtp.c
```diff
@@ -1253,14 +1451,49 @@ static void srtp_calc_aead_iv(srtp_stream_ctx_t *stream, v128_t *iv,
...
+
+  // Determine the authentication tag size
+  if (stream->session_keys[0].rtp_cipher->algorithm == SRTP_AES_128_GCM ||
+      stream->session_keys[0].rtp_cipher->algorithm == SRTP_AES_256_GCM) {
+      tag_len = 0;
+  } else {
+      tag_len = srtp_auth_get_tag_length(stream->session_keys[0].rtp_auth);
+  }
+
+  base_mki_start_location -= tag_len;
```
rsith71: Done in next Push

thisisG:
Hey @rsith71 ! Any update on the final push? Not wanting to nag, just want the merge to come through. :-)

rsith71:
@thisisG Sorry I thought I pushed it a week ago.  I have the commit in my local.  I guess I forgot the push command.  Let me go ahead and push the code review requests.

JonathanLennox:
srtp.h still has a comment that says:
```
/* 
 * nota bene: since libSRTP doesn't support the use of the MKI, the
 * SRTP_MAX_TRAILER_LEN value is just the maximum tag length
 */
```

ulfolsson:
<Now that #224 is merge, can this issue be closed?
Fine by me. Will start to use and test the new implementation in a new project. Thank's for your cooperation

mjerris:
I'm working on a commit to add MKI support to FreeSWITCH.  Other than juggling the new functions for protect/unprotect when we do MKI, what else do I need to do with keys for this to work right?

rsith71:
Sorry was on PTO for 2 weeks.  Besides the new functions (protect/unprotect) you have to setup the policy (srtp_policy_t) correctly.  You will need to setup the policy to use keys not key and therefore key should be NULL.  Keys will be an array of keys with the MKI Index, mki size and the key.  Also will need to update the policies number of master keys (num_master_keys).  Test/srtp_driver.c has test using MKI with two keys and should be a good place to start.

Thank you,

mjerris:
Looking at key lifetime, i can't find any examples of how to do this on the internet anywhere or in any test programs.  It looks like you use srtp_key_limit_set but I don't see any details to confirm how this would be expected to work with MKI.. Is there any examples of this.

paulej:
I might be wrong, but I think the code that is in librsrtp isn't presently doing anything with this "limit" field.  There is logic there, but effectively useless since the lifetime is set to 2^64-1, basically.  I don't think this is something new related to MKI.  Rather, I think it's been that way for a very long time.  If there is a desired to allow setting this, perhaps we need to explore introducing a new public-facing API.  Maybe this one will work, but not sure without spending time on it.

Issue #67, opened 2014-08-21
What is the purpose of srtp_event_data_t->stream if srtp_stream_t is opaque?


ibc:
The event emitter provides a single argument of type `srtp_event_data_t` which contains a `srtp_stream_t` member. Why? It is 100% useless given that it is an opaque struct:

```
typedef struct srtp_stream_ctx_t *srtp_stream_t;
```

and `srtp_stream_ctx_t` is defined in the private file `include/srtp_priv.h`. So being in an event callback function I cannot even get the SSRC value of the affected stream:

```
void on_srtp_event(srtp_event_data_t* data) {
    uint32_t ssrc = data->stream->ssrc;
}


=> it produces:

error: member access into incomplete type 'struct srtp_stream_ctx_t'
        uint32_t ssrc = data->stream->ssrc;
                                    ^
/usr/local/include/srtp/srtp.h:275:16: note: forward declaration of 'srtp_stream_ctx_t'
typedef struct srtp_stream_ctx_t *srtp_stream_t;
```

so what is the purpose of the `stream` field in the event data? is there any way to get the SSRC value of the affected stream?

jfigus:
Closing issue due to inactivity

ibc:
Inactivity? :)
I think the bug was properly reported and no more action was needed (but just a proper fix/implementation). I expected this to be done in the announced future new API design.

jfigus:
This issue is tagged as a "question", not a "bug".  Since the most recent discussion was October 9, it looked inactive to me.

pabuhler:
Looks like this was missed in the move to 2.0, this is a shame I think it will be a long time to 3.0 .
One option would be make a new api in 2.1 and depreciate the old api, removing it in 3.0
The fix would be to swap the srtp_stream_t argument with uint32_t SSRC. 
I wounder how many users are actually using this API.

ibc:
> I wounder how many users are actually using this API.

Probably no one (other than for logging purposes) as such a current API is of very little use.

pabuhler:
So then the question is if we can just change the srtp_stream_t argument to be a uintptr_t, and set it to the SSRC in version 2.1  Changing it to be uint32_t in 3.0 .

ibc:
Same SSRC may happen many times in different SRTP contexts/sessions. I expect the associated session should be given within the callback.

pabuhler:
in 2.0 the event data signature is
```
typedef struct srtp_event_data_t {
[...]

PR #234, opened 2017-01-25
Report ssrc instead of srtp_stream_t in srtp_event_data_t

pabuhler:
An intermediate fix for #67.

srtp_stream_t is an opaque type and unusable from the public api,
this made the entire event framework almost useless.
This change swaps the stream point with an ssrc, which is used as a key
in the public api. A uintptr_t is used instead of a uint32_t to ensure
the size of the srtp_event_data_t struct remains unchanged.

This change could potentially break existing code but the assumption
is that the number of uses of this api that are doing anything with this
value after the version 2.0 release approaches zero.

### DIFF
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -1408,7 +1408,7 @@ typedef enum {
 
 typedef struct srtp_event_data_t {
   srtp_t        session;  /**< The session in which the event happend. */
-  uintptr_t     ssrc;     /**< The ssrc in host order of the stream in which the event happend */
+  uint32_t      ssrc;     /**< The ssrc in host order of the stream in which the event happend */
   srtp_event_t  event;    /**< An enum indicating the type of event.   */
 } srtp_event_data_t;
 
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -270,7 +270,7 @@ typedef enum {
   srtp_err_status_pfkey_err    = 24  /**< error while using pfkey                 */
 } srtp_err_status_t;
 
-typedef struct srtp_stream_ctx_t_ srtp_stream_ctx_t;
+
 typedef struct srtp_ctx_t_ srtp_ctx_t;
 
 /* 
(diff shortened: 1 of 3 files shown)


include/srtp.h
```diff
@@ -1424,7 +1408,7 @@ typedef enum {
 
 typedef struct srtp_event_data_t {
   srtp_t        session;  /**< The session in which the event happend. */
-  srtp_stream_t stream;   /**< The stream in which the event happend.  */
+  uintptr_t     ssrc;     /**< The ssrc in host order of the stream in which the event happend */
```
thisisG: I think the use of `uintptr_t` requires some additional checks and defines configure and integers.h akin to what we do in order to support other C99ish types.
According to C99 7.18.1.4 the type is also optional so I think we need some guards, even for C99 compliant compilers.

pabuhler: ok, but If the check is added to configure.in and a conditional define set in interges.h, is the change otherwise sane given the potential breakage.

paulej: Would anyone really be relying on the length of this structure, but not looking at the data?  I personally think it's a bad idea to assume the length of any data structure doesn't change.

thisisG: Yeah I have no objections to the overall change - seems to make sense to me from a coding point of view at least, not too familiar with actually using the interface... 

Could leave the PR open a bit to see if others have inputs regarding the usability (terminology?), but pretty sure it is better to not have to depend on the private API in order to use the public API. :-)

pabuhler: @paulej , I'm not sure was just trying to be carefully. Someone could possibly be printing the value ... Could change to use a uint32_t now and be done with it.

paulej:
This is where the mailing list would be handy, I guess. Still, I can't imagine people operating on an assumed fixed length.  I could believe somebody imported srtp_priv.h so they could access that field.

pabuhler:
@paulej so you are advocating not to do this change until v3.0 or that it might as well be changed to uint32_t now?

paulej:
My opinion is to just use uint32_t to represent the ssrc.  I just doubt anyone is assuming a length for that data structure.

srtp/srtp.c
```diff
@@ -1026,7 +1026,7 @@ srtp_stream_init(srtp_stream_ctx_t *srtp,
  srtp_event_reporter(srtp_event_data_t *data) {
 
    srtp_err_report(srtp_err_level_warning, "srtp: in stream 0x%x: ", 
-	      data->stream->ssrc);
+                   data->ssrc);
```
thisisG: Just to make sure - this was previously incorrect and would print the information in Network Order (since there was no nothl() in the old code)? Making it look... odd on little endian systems. If so this PR is sortof a secondary bugfix as well, maybe mention it in the final commit/merge message. :)

LGTM if the above assumption was correct!

paulej:
Two days too late, I guess.  I was going to say that it looks good.  :)

Issue #158, opened 2016-05-05
Condition has identical expressions


davidsilveiro:
Within /test/rtp_decoder.c:444 the conditions arises with the exact same expressions on either side:

**if (dcdr->start_tv.tv_sec == 0 && dcdr->start_tv.tv_sec == 0**

PR #236, opened 2017-01-26
test/rtp_decoder.c: Removed superfluous conditional

thisisG:
No need to check the same thing twice! Re issue #158


JonathanLennox:
I don't think this was the right fix -- I suspect the second clause was actually supposed to be a test for tv_usec == 0 (i.e. the timeval actually encodes a 0 interval, not just a < 1s interval).

thisisG:
I will have another look @JonathanLennox!

PR #239, opened 2017-01-26
octet_string_set_to_zero() now delegates to OPENSSL_cleanse() if available, if not it will use srtp_cleanse() to zero memory.

thisisG:
Please see the first commit for a more in depth explanation.

Posting this since I would like to get some input on how to proceed before a merge happens.

The problem is that there is some code smell:
/crypto/hash/hmac.c:
`srtp_burn(a, sizeof(srtp_hmac_ctx_t) + sizeof(srtp_auth_t));`
/crypto/hash/hmac_ossl.c:
`srtp_burn(a, sizeof(HMAC_CTX) + sizeof(srtp_auth_t));`
/crypto/hash/null_auth.c:
`srtp_burn(a, sizeof(srtp_null_auth_ctx_t) + sizeof(srtp_auth_t));`

I'd like some input on if this is safe or not considering that we theoretically can exceed the limits of `size_t` by adding the result of two `sizeof()`s together, however this is not likely (?) since we are in control of the structures in question, as well as performing the memory allocation in one go in `srtp_crypto_alloc()` which would not succeed in allocating if we exceed the maximum value for 'size_t'.

Other than that please review with some scrutiny. :)

### DIFF
--- a/crypto/cipher/aes_gcm_ossl.c
+++ b/crypto/cipher/aes_gcm_ossl.c
@@ -163,7 +163,7 @@ static srtp_err_status_t srtp_aes_gcm_openssl_dealloc (srtp_cipher_t *c)
     if (ctx) {
         EVP_CIPHER_CTX_free(ctx->ctx);
 	/* zeroize the key material */
-	octet_string_set_to_zero((uint8_t*)ctx, sizeof(srtp_aes_gcm_ctx_t));
+    octet_string_set_to_zero(ctx, sizeof(srtp_aes_gcm_ctx_t));
 	srtp_crypto_free(ctx);
     }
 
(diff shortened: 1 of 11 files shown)


paulej:
The cosmetic comment I have is just the name "srtp_burn()".  The name doesn't tell me "zero memory".  I'd prefer a function name that tells me exactly what it does.

I'm not worried about adding sizeof() things together.  It could be worse if we used a constant.  As you said, we're in control of what the length is and the buffer being set to zero.

The important question is why this is needed.  Are there platforms where this zeroing function does not actually zero?  Perhaps the concern is that the call is optimized out and memory actually doesn't get zeroed on exit?

thisisG:
I can change the name back to the original name. Used 'srtp_burn()' because of the naming in the cryptocoding source and the other Cisco library.

But yes - the chief concern is that the call can be optimised away due it having no side effects in certain situations. You could argue that this change gives a false sense of security due to the lack of guaranteed execution, but for portable code this seems like the best alternative.

JonathanLennox:
I agree we should either leave the function name as it is, or else just add "secure" to it.  I don't like "burn" either.

When libsrtp is built with OpenSSL, the function could just call OPENSSL_cleanse() rather than write our own implementation.

thisisG:
OK! I'll change the original function to be a conditional wrapper to OPENSSL_cleanse() if compiled with openssl, if not use the burn function (and change the name to srtp_cleanse()) and stick to the void pointer and size_t length for consistent interface.
Thanks!

PR #240, opened 2017-01-27
ISSUE-235 SIGSEGV in rdb_add_index() when index >= rdb->window_start+256

timfuze:
@pabuhler 
Here is the patch. Thanks for looking into it.

### DIFF
--- a/crypto/replay/rdb.c
+++ b/crypto/replay/rdb.c
@@ -117,10 +117,17 @@ srtp_err_status_t srtp_rdb_add_index (srtp_rdb_t *rdb, uint32_t p_index)
 
         delta -= rdb_bits_in_bitmask - 1;
 
-        /* shift the window forward by delta bits*/
-        v128_left_shift(&rdb->bitmask, delta);
-        v128_set_bit(&rdb->bitmask, rdb_bits_in_bitmask - 1);
-        rdb->window_start += delta;
+        if (delta >= rdb_bits_in_bitmask) {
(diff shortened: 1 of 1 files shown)


timfuze:
hmm.. strange but it shouldn't be compiling as index name was changed to p_index... not sure how it was able to compile for me.

timfuze:
there must be some extern defined as "index" else where.  I caught this only when I was compiling it for Android.  Revert the old and commit the new one.

pabuhler:
So I have been trying out this patch, and it is possible to get the test to pass if I massage it abit, but in the end there is not much of a patch left. Whats more interesting is that it does not seam possible to recreate the crash with the existing code, it should be straight forward given the explanation.
also looking though the code , as long as srtp_rdb_check() is called before srtp_rdb_add_index() i don't see how a crash can happen. I would appreciate an other opinion, but as it stands I am inclined to reject this patch without a reproducible test.

JonathanLennox:
index() is the BSD function equivalent to the standard C strchr().  Many C libraries provide it for compatibility.

timfuze:
@pabuhler, that's fine. Thanks for looking into it.  This is only seen in our server which I haven't dug enough to know what caused it.  I only applied the patch and it seems not to have the issue anymore. 

@JonathanLennox , thanks for pointing it out for me.  Interesting..

pabuhler:
@timfuze , you are welcome to keep using the patch in you own code, but the fact that it fails the tests would make me think tiwice.

Issue #230, opened 2017-01-19
Clean up the debug api and configure options


pabuhler:
The debug options are not complete as it stands today.
The idea would be to create an application call back for debug and to enable an option of debugging to std out based on compile time option.
It should be possible through the api to enable debug level for each module.
By default no debug should be enabled.
There should be no out put to stdout/stderr, all out put should go through the debug api.  
This is a follow on from #183 #48 #110

pabuhler:
After looking through the code I have the following proposal:
- configure option "ENABLE_DEBUGGING" this will enable debugging for all modules regardless of weather they have been enabled though the api, this is very verbose, to be used only when debugging and would be default off.
- configure option "ERR_REPORTING_STDOUT" & "ERR_REPORTING_FILE", this would cause error reporting to go to either stdout or given file name, by default neither would be defined
- new api to set callback that receiver error reports in the application

This simplifies it some what but still offers full flexibility, would be great to get some feedback before finish implementing.

paulej:
Would those be supported with --enable-debugging in "configure"?  Are you proposing to allow the file to be specified or would this be a pre-defined filename?

pabuhler:
The filename should be passed as an argument, I don't see how a default would make sense.

pabuhler:
Closing this now as configure has been updated and a public logging api added.

PR #241, opened 2017-01-30
Simplify logging of errors and debug

pabuhler:
First commit for #230 , a second commit containing a callback API for application to receive logging will come if this pull is merged.

Previously logging of debug and errors did not always work as expected.
Now there are three configure options to control output. The idea is not to
be overly complicated but to give some simple options that work.

--enable-log-stdout
This will out put all debug and error logging to stdout

--with-log-file=<filename>
This will output all debug and error logging to the given filename

--enable-debug-logging
This enable debug logging for all internal debug modules. This is very
verbose and should only be used for debugging and development.


pabuhler:
Will update documentation in #248 if accepted

Issue #114, opened 2015-08-30
reserved identifier violation


elfring:
I would like to point out that identifiers like "[`_AES_H`](https://github.com/cisco/libsrtp/blob/ed2212973b3248dde97849e38f3d50a752e98d41/crypto/include/aes.h#L46)" and "[`_DATATYPES_H`](https://github.com/cisco/libsrtp/blob/ed2212973b3248dde97849e38f3d50a752e98d41/crypto/include/datatypes.h#L47)" [do eventually not fit](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28HeaderGuard%29) to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

jfigus:
Good point.  Fortunately the standard C library doesn't have aes.h or datatypes.h.  Otherwise this probably would have been an issue much earlier.  I'll leave this issue open and address this sometime in the future.

PR #242, opened 2017-01-30
Don't use underscore as leading character in header guards.

pabuhler:
This address the concern in #114 , It still compiles both with and without opeenssl for me. Not sure if there is a platform where this will be a problem.

### DIFF
--- a/crypto/include/aes.h
+++ b/crypto/include/aes.h
@@ -78,4 +78,4 @@ void srtp_aes_decrypt(v128_t *plaintext, const srtp_aes_expanded_key_t *exp_key)
 }
 #endif
 
-#endif /* _AES_H */
+#endif /* AES_H */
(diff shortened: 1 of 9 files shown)


JonathanLennox:
Might be better to use SRTP_AES_H, etc, instead?  Just to avoid the possibility of conflicts with other headers?

pabuhler:
I was thinking the same but then should we change all includes? I see that openssl has HEADER_AES_H

JonathanLennox:
The public headers are probably higher priority -- particularly RTP_H seems dangerously likely to conflict with other code.  But I'd be inclined to change all of them, on the same principle that led us to add srtp_ to all the public symbols.

elfring:
Thanks for [your improvement of affected include guards](https://github.com/cisco/libsrtp/pull/242/commits/6a832d70822238af53ab04ffbf17c03fbc74d954 "Prefix public header include guards with SRTP_").

Issue #219, opened 2016-11-17
ekt.h includes srtp_priv.h, which is not installed


Lastique:
In libsrtp 2.0.0, the ekt.h header includes srtp_priv.h, which is not installed as part of the `install` target. Either srtp_priv.h and its dependencies should be installed as well, or ekt.h should not depend on it.

pabuhler:
@paulej should the ekt.h header even be there? Is the implementation of ekt in master something that is complete and supported?

paulej:
The ekt stuff that is currently in here is really non-functional.  We have a separate ekt_dev branch with a complete implementation of the avtcore draft.  That's going to be very challenging to merge with MKI changes going in.  Plus, the IETF PRIME WG is changing the EKT spec such that it might make sense to just re-write that code entirely.

Anyway, I think we can and probably should just remove all the EKT stuff right now, because it has no value.

PR #243, opened 2017-01-30
ekt is not really supported yet, remove from install

pabuhler:
The ekt support in master is not currently usable so remove it from the install. there is a ekt dev branch that will eventually be merged back to master, at which point the ekt header can be installed again.

This indirectly address #219 by not installing ekt.h .


pabuhler:
The idea with leaving it commented out was that hopefully it will generate a conflict when merging with the ekt branch and so get noticed.

Issue #194, opened 2016-07-26
Missing error checking in srtp_cipher_bits_per_second


trivee:
srtp_cipher_bits_per_second() does not check the status values returned by srtp_cipher_set_iv() and srtp_cipher_encrypt().

PR #244, opened 2017-01-31
Add simple error checking in timing test to avoid false results.

pabuhler:
This way the though put would be 0 instead of some amazingly
fast number.

Address #194

### DIFF
--- a/crypto/cipher/cipher.c
+++ b/crypto/cipher/cipher.c
@@ -620,10 +620,14 @@ uint64_t srtp_cipher_bits_per_second (srtp_cipher_t *c, int octets_in_buffer, in
     v128_set_to_zero(&nonce);
     timer = clock();
     for (i = 0; i < num_trials; i++, nonce.v32[3] = i) {
-        if (srtp_cipher_set_iv(c, (uint8_t*)&nonce, srtp_direction_encrypt) != srtp_err_status_ok)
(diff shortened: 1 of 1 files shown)


crypto/cipher/cipher.c
```diff
@@ -620,8 +620,10 @@ uint64_t srtp_cipher_bits_per_second (srtp_cipher_t *c, int octets_in_buffer, in
     v128_set_to_zero(&nonce);
     timer = clock();
     for (i = 0; i < num_trials; i++, nonce.v32[3] = i) {
-        srtp_cipher_set_iv(c, (uint8_t*)&nonce, srtp_direction_encrypt);
-        srtp_cipher_encrypt(c, enc_buf, &len);
+        if (srtp_cipher_set_iv(c, (uint8_t*)&nonce, srtp_direction_encrypt) != srtp_err_status_ok)
+            return 0;
```
thisisG: Should probably deallocate to avoid memory leaks!

crypto/cipher/cipher.c
```diff
@@ -620,8 +620,10 @@ uint64_t srtp_cipher_bits_per_second (srtp_cipher_t *c, int octets_in_buffer, in
...
+        if (srtp_cipher_encrypt(c, enc_buf, &len) != srtp_err_status_ok)
+            return 0;
```
thisisG: Deallocate here as well.

pabuhler:
your are right, it is only a test but it is the correct thing to do.

PR #246, opened 2017-02-03
test/rtp_decoder: Add missing conditional

thisisG:
This conditional was incorrectly removed in PR #236 where it was just
incorrectly set to be a duplicate of `(dcdr->start_tv.tv_sec == 0)`.

It makes sense to check that the microseconds are 0 as well as these are
used later in the test in `timersub()` in order to calculate and display the 
ΔT for the operations.


paulej:
LGTM

Issue #247, opened 2017-02-06
Evaluate removing the ISMAcryp options from configure


pabuhler:
Discovered while reviewing #238.
Check if this is still needed, seams to be dead and can be removed.

paulej:
The only reference I could find to this was the fact that some very old MPEG code was using the library (https://github.com/stsquad/mpeg4ip/blob/master/lib/ismacryp/README), but even it does not use this switch.  There was some discussion of this years ago: https://sourceforge.net/p/srtp/mailman/message/1810415/.

Projects like https://github.com/gpac/gpac do not rely on libsrtp.  So, it's unclear to me if any out there really do. I doubt it, given the option appears to have no effect.  (Or so it appears to me.)

My guess is that it would be safe to remove it.

PR #249, opened 2017-02-10
remove support for generic aesicm from configure.in

pabuhler:
This was removed form the code along time ago as referenced in #247.
Will removed references from documentation in #248.


crypto/cipher/aes_icm.c
```diff
@@ -92,7 +92,7 @@ srtp_debug_module_t srtp_mod_aes_icm = {
  *
  */
 
-static srtp_err_status_t srtp_aes_icm_alloc_ismacryp (srtp_cipher_t **c, int key_len, int forIsmacryp)
+static srtp_err_status_t srtp_aes_icm_alloc (srtp_cipher_t **c, int key_len)
```
thisisG: This function signature change may cause problems.

In `aes_icm.c:494` it is used like so:

~~~.c
const srtp_cipher_type_t srtp_aes_icm = {
    srtp_aes_icm_alloc, // throws error
    srtp_aes_icm_dealloc,
    srtp_aes_icm_context_init,
    0,                          /* set_aad */
    srtp_aes_icm_encrypt, // throws error
    srtp_aes_icm_encrypt, // throws error
    srtp_aes_icm_set_iv,
    0,                          /* get_tag */
    srtp_aes_icm_description,
    &srtp_aes_icm_test_case_1,
    SRTP_AES_ICM
};
~~~

In 'cipher.h:142' the `srtp_cipher_type_t` is defined as:

~~~.c
typedef struct srtp_cipher_type_t {
    srtp_cipher_alloc_func_t alloc;
    srtp_cipher_dealloc_func_t dealloc;
    srtp_cipher_init_func_t init;
    srtp_cipher_set_aad_func_t set_aad;
    srtp_cipher_encrypt_func_t encrypt;
    srtp_cipher_encrypt_func_t decrypt;
    srtp_cipher_set_iv_func_t set_iv;
    srtp_cipher_get_tag_func_t get_tag;
    const char                       *description;
    const srtp_cipher_test_case_t         *test_data;
    srtp_cipher_type_id_t id;
} srtp_cipher_type_t;
~~~

Where the signature of `srtp_cipher_alloc_func_t` is:
~~~.c
typedef srtp_err_status_t (*srtp_cipher_alloc_func_t)
    (srtp_cipher_pointer_t *cp, int key_len, int tag_len);
~~~

Since `tag_len` does not seem to make sense in the context of `srtp_aes_icm_alloc_ismacryp(...)` I suggest adding a dummy `int` in order to get the signature to match? Not sure what the correct approach here is.

thisisG: Btw - it is gcc 6.3.1 that complains about incompatible pointers... I am unsure if it will trigger any potential issues.

crypto/cipher/aes_icm.c
```diff
@@ -311,16 +292,15 @@ static void srtp_aes_icm_advance_ismacryp (srtp_aes_icm_ctx_t *c, uint8_t forIsm
  *  - fill buffer then add in remaining (< 16) bytes of keystream
  */
 
-static srtp_err_status_t srtp_aes_icm_encrypt_ismacryp (srtp_aes_icm_ctx_t *c,
-                                                 unsigned char *buf, unsigned int *enc_len,
-                                                 int forIsmacryp)
+static srtp_err_status_t srtp_aes_icm_encrypt (srtp_aes_icm_ctx_t *c,
```
thisisG: This function also throws an error about incompatible pointer. The signature to match is:

~~~.c
typedef srtp_err_status_t (*srtp_cipher_encrypt_func_t)
    (void *state, uint8_t *buffer, unsigned int *octets_to_encrypt);
~~~

Just changing '*c' to void did _not_ help :-) This worked and is used elsewhere:

~~~.c
static srtp_err_status_t srtp_aes_icm_encrypt (void *cv,
                                               uint8_t *buf, unsigned int *enc_len)
{
    (srtp_aes_icm_ctx_t*)c = (srtp_aes_icm_ctx_t*) cv;
    // ...
~~~

thisisG:
Hmm I see Travis complains a lot. Will dig a bit more - the final version seemed to compile fine for me locally.

crypto/cipher/aes_icm.c
pabuhler: You are right, the signature should match. If you see other chnages in this commit you will see that the previous version of the function was `srtp_aes_icm_alloc (srtp_cipher_t **c, int key_len, int forIsmacryp)` not quit sure how tag_len should have mapped to forIsmacryp, but it was ignored anyway. I will add an extra argument.

PR #209, opened 2016-09-14
Restore AES-192 under BoringSSL

mmdriley:
PR #93 added the SRTP_NO_AES192 macro and removed all references to AES-192
if it was set. The goal was to get libsrtp to compile with BoringSSL, which
didn't support AES-192 at the time (Feb 2015).

In the time since (Apr 2015), BoringSSL has restored support for AES-192:
https://boringssl.googlesource.com/boringssl/+/1049e26f
https://boringssl.googlesource.com/boringssl/+/5dca031c

Remove the SRTP_NO_AES192 macro, since it's no longer needed.
Effectively reverts #93.

PR #253, opened 2017-02-21
Cipher type cleanup for aes

pabuhler:
This applies #170 & #172  on to master as well as making more consistent use of the srtp AES_XXX_ICM defines.

### DIFF
--- a/crypto/cipher/aes_icm.c
+++ b/crypto/cipher/aes_icm.c
@@ -130,11 +130,11 @@ static srtp_err_status_t srtp_aes_icm_alloc (srtp_cipher_t **c, int key_len, int
 
     switch (key_len) {
     case 46:
-        (*c)->algorithm = SRTP_AES_256_ICM;
+        (*c)->algorithm = SRTP_AES_ICM_256;
         (*c)->type = &srtp_aes_icm_256;
         break;
     default:
-        (*c)->algorithm = SRTP_AES_128_ICM;
+        (*c)->algorithm = SRTP_AES_ICM_128;
         (*c)->type = &srtp_aes_icm_128;
         break;
     }
@@ -503,7 +503,7 @@ const srtp_cipher_type_t srtp_aes_icm_128 = {
     0,                          /* get_tag */
     srtp_aes_icm_128_description,
     &srtp_aes_icm_128_test_case_0,
-    SRTP_AES_128_ICM
+    SRTP_AES_ICM_128
 };
 
 const srtp_cipher_type_t srtp_aes_icm_256 = {
@@ -517,5 +517,5 @@ const srtp_cipher_type_t srtp_aes_icm_256 = {
     0,                          /* get_tag */
     srtp_aes_icm_256_description,
     &srtp_aes_icm_256_test_case_0,
-    SRTP_AES_256_ICM
+    SRTP_AES_ICM_256
 };
--- a/crypto/cipher/aes_icm.c
+++ b/crypto/cipher/aes_icm.c
@@ -106,7 +106,7 @@ static srtp_err_status_t srtp_aes_icm_alloc (srtp_cipher_t **c, int key_len, int
      * has not broken anything. Don't know what would be the
      * effect of skipping this check for srtp in general.
      */
-    if (key_len != 30 && key_len != 38 && key_len != 46) {
+    if (key_len != 30 && key_len != 46) {
         return srtp_err_status_bad_param;
     }
 
@@ -131,9 +131,6 @@ static srtp_err_status_t srtp_aes_icm_alloc (srtp_cipher_t **c, int key_len, int
     switch (key_len) {
     case 46:
         (*c)->algorithm = SRTP_AES_256_ICM;
-        break;
-    case 38:
-        (*c)->algorithm = SRTP_AES_192_ICM;
         break;
     default:
         (*c)->algorithm = SRTP_AES_128_ICM;
--- a/crypto/cipher/aes_icm.c
+++ b/crypto/cipher/aes_icm.c
@@ -404,47 +404,47 @@ static srtp_err_status_t srtp_aes_icm_encrypt (void *cv,
 static const char srtp_aes_icm_128_description[] = "AES-128 integer counter mode";
 static const char srtp_aes_icm_256_description[] = "AES-256 integer counter mode";
 
-static const uint8_t srtp_aes_icm_test_case_0_key[30] = {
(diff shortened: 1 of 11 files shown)


include/srtp.h
```diff
@@ -1046,6 +1046,95 @@ void srtp_crypto_policy_set_aes_cm_256_hmac_sha1_32(srtp_crypto_policy_t *p);
  */
 void srtp_crypto_policy_set_aes_cm_256_null_auth(srtp_crypto_policy_t *p);
```
paulej: I'm all for keeping the API consistent, but I will note that here we have aes_cm_xxx and the variable names inside are aes_icm_xxx or AES_XXX_ICM.  Would be nice if those were all consistent.

crypto/cipher/aes_icm.c
```diff
@@ -498,8 +501,21 @@ const srtp_cipher_type_t srtp_aes_icm = {
...
+const srtp_cipher_type_t srtp_aes_icm_256 = {
+    srtp_aes_icm_alloc,
+    srtp_aes_icm_dealloc,
+    srtp_aes_icm_context_init,
+    0,                          /* set_aad */
+    srtp_aes_icm_encrypt,
+    srtp_aes_icm_encrypt,
+    srtp_aes_icm_set_iv,
+    0,                          /* get_tag */
+    srtp_aes_icm_256_description,
```
JonathanLennox: Suggest renaming srtp_aes_icm_test_case_0 as srtp_aes_icm_128_test_case_0, and srtp_aes_icm_test_case_1 as srtp_aes_icm_256_test_case_0.

JonathanLennox: (This would match the names in aes_icm_ossl.c.)

srtp/srtp.c
```diff
@@ -3081,6 +3098,53 @@ srtp_crypto_policy_set_aes_cm_256_null_auth (srtp_crypto_policy_t *p)
 }
 
 #ifdef OPENSSL
+void
+srtp_crypto_policy_set_aes_cm_192_hmac_sha1_80(srtp_crypto_policy_t *p) {
```
paulej: Any reason this function isn't sitting near the other functions with similar names?

crypto/cipher/aes_icm.c
```diff
@@ -483,7 +483,7 @@ static const srtp_cipher_test_case_t srtp_aes_icm_test_case_1 = {
     0,
     NULL,
     0,
-    &srtp_aes_icm_test_case_0                 /* pointer to next testcase */
+    NULL,                 /* pointer to next testcase */
```
JonathanLennox: Separating the aes_128 and aes_256 test cases.

crypto/cipher/aes_icm.c
```diff
@@ -402,7 +401,8 @@ static srtp_err_status_t srtp_aes_icm_encrypt (void *cv,
     return srtp_err_status_ok;
 }
 
-static const char srtp_aes_icm_description[] = "aes integer counter mode";
+static const char srtp_aes_icm_128_description[] = "AES-128 integer counter mode";
+static const char srtp_aes_icm_256_description[] = "AES-256 integer counter mode";
```
JonathanLennox: libsrtp's built-in aes implementation doesn't implement aes-192.

crypto/cipher/aes_icm_ossl.c
```diff
@@ -460,7 +460,7 @@ static const srtp_cipher_test_case_t srtp_aes_icm_256_test_case_2 = {
  * This is the function table for this crypto engine.
  * note: the encrypt function is identical to the decrypt function
  */
-const srtp_cipher_type_t srtp_aes_icm = {
+const srtp_cipher_type_t srtp_aes_icm_128 = {
```
JonathanLennox: They were already defined for the aes_icm_ossl version.

crypto/cipher/aes_icm.c
paulej: Oh, well that's a good reason. I had it in my head that we had only disabled 192-bit keys for BoringSSL, but we now have that via #209.  While I don't particularly see the value in 192-bit key lengths, myself, I sure do like consistency.

crypto/cipher/aes_icm.c
pabuhler: :+1:

crypto/cipher/aes_icm.c
pabuhler: @paulej well you could always implement it ...

crypto/cipher/aes_icm.c
```diff
@@ -501,8 +501,21 @@ const srtp_cipher_type_t srtp_aes_icm = {
     srtp_aes_icm_encrypt,
     srtp_aes_icm_set_iv,
     0,                          /* get_tag */
-    srtp_aes_icm_description,
-    &srtp_aes_icm_test_case_1,
-    SRTP_AES_ICM
+    srtp_aes_icm_128_description,
+    &srtp_aes_icm_test_case_0,
+    SRTP_AES_128_ICM
```
pabuhler: I was in general just following the existing naming conventions, will look at consolidating one way or the other.

srtp/srtp.c
pabuhler: I would guess it was placed there due to the opeenssl block

crypto/cipher/aes_icm.c
paulej: If I had time, it would be fun to do.

include/srtp.h
```diff
@@ -1046,6 +1046,95 @@ void srtp_crypto_policy_set_aes_cm_256_hmac_sha1_32(srtp_crypto_policy_t *p);
...
+ * @brief srtp_crypto_policy_set_aes_cm_192_hmac_sha1_32() sets a crypto
+ * policy structure to a short-authentication tag policy using AES-192
+ * encryption.
+ *
+ * @param p is a pointer to the policy structure to be set
+ *
+ * The function call srtp_crypto_policy_set_aes_cm_192_hmac_sha1_32(&p)
+ * sets the crypto_policy_t at location p to use policy
+ * AES_CM_192_HMAC_SHA1_32 as defined in
+ * draft-ietf-avt-srtp-big-aes-03.txt.  This policy uses AES-192
```
paulej: This is now RFC 6188.  Not sure how many such references like this exist in the code.

pabuhler: we should try to do a review of references in general, I think there are a few that are incorrect

PR #258, opened 2017-02-22
add AES-GCM to DTLS-SRTP Protection Profiles v2

pabuhler:
This is a reapplication of #188 on to master taking into account comments from both #188 & #185

### DIFF
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -1217,8 +1217,6 @@ typedef enum {
   srtp_profile_reserved           = 0,
   srtp_profile_aes128_cm_sha1_80  = 1,
   srtp_profile_aes128_cm_sha1_32  = 2,
-  srtp_profile_aes256_cm_sha1_80  = 3,
-  srtp_profile_aes256_cm_sha1_32  = 4,
   srtp_profile_null_sha1_80       = 5,
   srtp_profile_null_sha1_32       = 6,
   srtp_profile_aead_aes_128_gcm   = 7,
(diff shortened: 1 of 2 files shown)


JonathanLennox:
While we're in this space, the srtp_profile_aes256_* profiles should probably get removed -- those aren't in the IANA registry, since they didn't make it into either RFC 5764 or RFC 6188.  (Not sure why.)

paulej:
Oh, one question on removal of srtp_profile_aes256* entries.  Why is this claim made about not being in RFC 6188.  They appear to be there, or am I misreading what you wrote @JonathanLennox?

JonathanLennox:
@pauej: As far as I can tell, RFC 6188 has codepoints for security descriptions (section 5), but none for DTLS/SRTP, which is what the srtp_profile values encode.  Or am I missing something?

There's no mention of "DTLS" in the RFC, it doesn't cite RFC 5764, and the IANA registrations for DTLS profiles https://www.iana.org/assignments/srtp-protection/srtp-protection.xhtml has values 0x0003 and 0x0004 unassigned.

paulej:
@JonathanLennox No, you're right (as usual).  I didn't read the comment above the function and just made the assumption that it was a nice short-hand way of setting specific "profiles" (not the be confused with the DTLS "Protection Profiles".  The word "profile" is a bit over-used since it had meaning in libsrtp before dtls-srtp came along.  This type should have been more explicitly been called "srtp_dtls_protection_profile_t" or something.  That way, my pee brain would get less confused.

PR #261, opened 2017-02-24
Add log handler api to receive log messages from libSRTP

pabuhler:
This address the second half of #230 

This can be tested ruining the test/srtp_driver test application with the new -o option.

### DIFF
--- a/crypto/kernel/err.c
+++ b/crypto/kernel/err.c
@@ -67,12 +67,32 @@ srtp_err_status_t srtp_err_reporting_init ()
     return srtp_err_status_ok;
 }
 
+static srtp_err_report_handler_func_t * srtp_err_report_handler = NULL;
+
+srtp_err_status_t srtp_install_err_report_handler(srtp_err_report_handler_func_t func)
+{
+    srtp_err_report_handler = func;
+}
+
 void srtp_err_report (srtp_err_reporting_level_t level, const char *format, ...)
 {
     va_list args;
-    va_start(args, format);
     if (srtp_err_file != NULL) {
+        va_start(args, format);
         vfprintf(srtp_err_file, format, args);
+        va_end(args);
     }
-    va_end(args);
+    if (srtp_err_report_handler != NULL) {
+        va_start(args, format);
+        char msg[512];
+        if (vsnprintf(msg, sizeof(msg), format, args) > 0) {
+            /* strip trailing \n, callback should not have one */
+            size_t l = strlen(msg);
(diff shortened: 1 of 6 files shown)


JonathanLennox:
Suggestion: should the log handler callback take a user data pointer, for applications where the caller's log object isn't a global?

crypto/kernel/err.c
```diff
@@ -67,12 +67,32 @@ srtp_err_status_t srtp_err_reporting_init ()
...
+        va_start(args, format);
         vfprintf(srtp_err_file, format, args);
+        va_end(args);
+    }
+    if (srtp_err_report_handler != NULL) {
+        va_start(args, format);
+        char msg[512];
+        if (vsnprintf(msg, sizeof(msg), format, args) > 0) {
+            /* strip trailing \n, callback should not have one */
+            size_t l = strlen(msg);
```
thisisG: Requires a `#include <string.h>` for strlen methinks.

crypto/kernel/err.c
```diff
@@ -67,12 +67,32 @@ srtp_err_status_t srtp_err_reporting_init ()
     return srtp_err_status_ok;
 }
 
+static srtp_err_report_handler_func_t * srtp_err_report_handler = NULL;
+
+srtp_err_status_t srtp_install_err_report_handler(srtp_err_report_handler_func_t func)
+{
+    srtp_err_report_handler = func;
+}
```
thisisG: Missing some sort of return here. Probably just an `srtp_err_status_ok` since there is little that can go wrong? Or make it void. Not sure. :-)

pabuhler:
@JonathanLennox I can see how a user data pointer would be useful. Added it in 9876231.

crypto/kernel/err.c
```diff
@@ -67,12 +68,33 @@ srtp_err_status_t srtp_err_reporting_init ()
...
     va_list args;
-    va_start(args, format);
     if (srtp_err_file != NULL) {
+        va_start(args, format);
         vfprintf(srtp_err_file, format, args);
+        va_end(args);
+    }
+    if (srtp_err_report_handler != NULL) {
+        va_start(args, format);
+        char msg[512];
```
thisisG: Should this msg buffer be considered a storage of secrets which should be cleared using `octet_string_set_to_zero()` after being used? Honestly not sure. :-)

pabuhler: hmm, you may have a point. Will think on it. Certainly would not hurt.

Issue #260, opened 2017-02-24
Reduce the use of numeric literal constants used in the code 


pabuhler:
This is a follow on from a discussion in #185. there are far to meany places in the code where the same literal numerical constant is reused. Instead use constant defines. There are already some examples of doing this in the code ie from srtp.h
```
#define SRTP_AEAD_SALT_LEN	12
#define SRTP_AES_128_GCM_KEYSIZE_WSALT   SRTP_AEAD_SALT_LEN + 16
#define SRTP_AES_192_GCM_KEYSIZE_WSALT   SRTP_AEAD_SALT_LEN + 24
#define SRTP_AES_256_GCM_KEYSIZE_WSALT   SRTP_AEAD_SALT_LEN + 32
```
There are equivalent values for AES-ICM but they are locked away in the aes_icm_ossl.h header, they should be made public and used.

PR #262, opened 2017-03-03
Return an error in srtp_calc_aead_iv_srtcp() if seq_num is invalid

paulej:
This is to address the comment in #259 to return an error from srtp_calc_aead_iv_srtcp() if the seq_num parameter is invalid.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -3189,9 +3189,13 @@ srtp_crypto_policy_set_aes_gcm_256_16_auth(srtp_crypto_policy_t *p) {
  *         seq_num       - The SEQ value to use for the IV calculation.
  *         *hdr          - The RTP header, used to get the SSRC value
  *
+ * Returns: srtp_err_status_ok if no error or srtp_err_status_bad_param
+ *          if seq_num is invalid
+ *
  */
-static void srtp_calc_aead_iv_srtcp(srtp_session_keys_t *session_keys, v128_t *iv, 
-                                    uint32_t seq_num, srtcp_hdr_t *hdr)
+static srtp_err_status_t
+srtp_calc_aead_iv_srtcp(srtp_session_keys_t *session_keys, v128_t *iv,
+                        uint32_t seq_num, srtcp_hdr_t *hdr)
 {
     v128_t	in;
     v128_t	salt;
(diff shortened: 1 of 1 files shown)


srtp/srtp.c
```diff
@@ -3202,6 +3206,9 @@ static void srtp_calc_aead_iv_srtcp(srtp_session_keys_t *session_keys, v128_t *i
     in.v16[0] = 0;
     memcpy(&in.v16[1], &hdr->ssrc, 4); /* still in network order! */
     in.v16[3] = 0;
+    if (seq_num & 0x80000000) { /* the most significant bit should be zero */
```
thisisG: Maybe use a `#define MOST_SIG_BIT_32_BIT_WORD (0x80000000)` re #260 ?
(Terrible name made up on the fly :-) Not sure if it applies here though, since this is sort of a one-off value of which name would perhaps not make sense in other contexts such as 

~~~.c
crypto/math/datatypes.c=148=v128_bit_string(v128_t *x) {
crypto/math/datatypes.c-149-  int j, i;
crypto/math/datatypes.c-150-  uint32_t mask;
crypto/math/datatypes.c-151-  
crypto/math/datatypes.c-152-  for (j=i=0; j < 4; j++) {
crypto/math/datatypes.c:153:    for (mask=0x80000000; mask > 0; mask >>= 1) {
crypto/math/datatypes.c-154-      if (x->v32[j] & mask)
crypto/math/datatypes.c-155-    bit_string[i] = '1';
crypto/math/datatypes.c-156-      else
crypto/math/datatypes.c-157-    bit_string[i] = '0';
crypto/math/datatypes.c-158-      ++i;
crypto/math/datatypes.c-159-    }
crypto/math/datatypes.c-160-  }
crypto/math/datatypes.c-161-  bit_string[128] = 0; /* null terminate string */
crypto/math/datatypes.c-162-
crypto/math/datatypes.c-163-  return bit_string;
crypto/math/datatypes.c-164-}
~~~

paulej: Given this is a constant we're not using elsewhere, I'm not terribly favorable to #defining it.  Either way, it's a line of code that somebody could break.  But, looking at it again, I do wonder if we should tack on an 'L' on the end of that.  And, perhaps we should do the same in datatypes.c?

thisisG: Tacking on 'UL' I think would be a good thing. In that case we get compilation warnings if we try to assign it to a 'int32_t' or something else that might overflow. 

Would not produce a warning for `&` though, but it is an almost free low effort piece of defensive programming - normally a good thing!

paulej: I changed it

PR #265, opened 2017-03-06
Suggestion for unit tests framework

thisisG:
This is a very simple header only framework without bells and whistles. See commit introducing the framework for more information.

This is in response to requests for more tests in #259. A commented out test is included which depends on #262 to ensure we get an error if we exceed a sequence number of `0x7FFFFFFF` in `srtp_calc_aead_iv_srtcp()`.

No work has been done to start moving tests from test drivers. The reasons are:

  1) Want to ensure we settle on a framework before porting any tests.
  2) One could argue the test drivers are integration/module tests which can be seen as orthogonal to unit tests.


test/test_srtp.c
```diff
@@ -0,0 +1,86 @@
...
+    sequence_num = 0x7FFFFFFF + 0x1;
+
+    // When
+    srtp_err_status_t status = srtp_calc_aead_iv_srtcp(
+        &session_keys, &init_vector, sequence_num, &header);
+
+    // Then
+    TEST_CHECK(status == srtp_err_status_bad_param);
+}
+*/
```
pabuhler: ops newline :)

test/test_srtp.c
```diff
@@ -0,0 +1,86 @@
...
+              srtp_calc_aead_iv_srtcp_zero_happy_case},
+             /* {"srtp_calc_aead_iv_srtcp_seq_num_over_0x7FFFFFFF_bad_param()",
+              srtp_calc_aead_iv_srtcp_seq_num_over_0x7FFFFFFF_bad_param}, */
+             {NULL} /* End of tests */};
+
+/*
+ * Implementation
+ */
+
+void srtp_calc_aead_iv_srtcp_zero_happy_case()
```
pabuhler: I had to look at the code to determine what this did, not sure I can come up with a better name though, maybe just say what it does ie,   srtcp_calc_aead_iv_srtcp_index_zero ....

pabuhler:
:+1:  this works for me. I would prefer the functions exposed through an internal header rather than including the c file.
Waiting with an approval until others have a time to comment :-)

thisisG:
Yeah I would also prefer to expose the now `static` functions contained in `srtp.c` in `srtp_interal.[c|h]` files, or just add the declarations in `srtp_priv.h` since it is already present. I think the latter makes sense since the file already contains the internal API and adding some more declarations there is fairly low effort. 

Including a `.c` file always feels icky, but should be all right if we do not want to refactor out the `static`s yet.

test/test_srtp.c
thisisG: Huh, weird that clang-format did not fix that one. Will fix!

test/test_srtp.c
```diff
@@ -0,0 +1,86 @@
...
+    memset((v128_t *)&zero_vector, 0, sizeof(v128_t));
+
+    // Given
+    memset(&session_keys, 0, sizeof(srtp_session_keys_t));
+    memset(&init_vector, 0, sizeof(v128_t));
+    memset(&header, 0, sizeof(srtcp_hdr_t));
+    sequence_num = 0x0;
+
+    // When
+    srtp_calc_aead_iv_srtcp(&session_keys, &init_vector, sequence_num, &header);
```
paulej: Would we want to check the return value here?  Should indicate OK.

paulej:
@thisisG I assume the reason any function is static is to prevent misuse/abuse by people doing something like including srtp_(internal|priv).h in their software.  What's the motivation for defining the internal functions in a private header file?  Just to avoid #including a .c file (aesthetics) or is there a technical reason?  (I don't care one way or the other, but I will note that most unit test code I have seen does strange things in order to get down to the function level.  So seeing that include only shocked me for about 1/2 second.)

test/test_srtp.c
thisisG: Yup! Will fix as soon as #262 lands to avoid Travis yelling at me.

thisisG:
Aesthetics in this case. :-)

We might end up with multiple definition errors depending on what other things we might include in the source file but I think it is totally OK to do this for unit tests.

```.c
#include "../srtp/srtp.c"
#include "srtp.h"
```
works fine for example.

As long as no one does

```.c
#include "../srtp/srtp.c"
#include "../srtp/srtp.c"
```
we're good!

PR #266, opened 2017-03-06
Test for AES-256 mimicking srtp_protect/unprotect functions

rchan1611:
When OpenSSL is used as AES provider, make SRTCP choose the correct nonce.
We copy the logic from srtp_protect() and srtp_unprotect() for SRTCP

(AES-192 still doesn't work for SRTP/SRTCP as we don't check for SRTP_AES_192_ICM)


pabuhler:
This should be covered by #253

aalba6675:
Yes covered by #253. Please close.

PR #267, opened 2017-03-06
crypto/kernel/err.c: Include datatypes.h

thisisG:
Required for octet_string_set_to_zero() to silence warnings.

### DIFF
--- a/crypto/kernel/err.c
+++ b/crypto/kernel/err.c
@@ -47,6 +47,7 @@
 #endif
 
 #include "err.h"
+#include "datatypes.h"
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


paulej:
Related to this, or more precisely the "cleanse" function.  I noted a few places in the code that clear keys using memset().  For example, the function srtp_kdf_clear() uses memset().  A quick search for memset() showed a few places that look like perhaps we ought to be scrubbing memory more carefully.  Your thoughts?

thisisG:
Totally agree. Will have a `git grep` session and make a PR for suspicious memset()s.

Merging and closing this one though!

PR #269, opened 2017-03-07
memset to octet_string_set_to_zero for clearing secrets

thisisG:
@paulej noticed at least one place where `memset()` was used to clear up secrets in `srtp_kdf_clear()` so I fixed that one. Also found some missed zeroing in `srtp_stream_init_keys()` for early returns.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -944,6 +944,8 @@ srtp_stream_init_keys(srtp_stream_ctx_t *srtp, srtp_master_key_t *master_key,
   stat = srtp_kdf_init(&kdf, (const uint8_t *)tmp_key, kdf_keylen);
 #endif
   if (stat) {
+    /* zeroize temp buffer */
+    octet_string_set_to_zero(tmp_key, MAX_SRTP_KEY_LEN);
     return srtp_err_status_init_fail;
   }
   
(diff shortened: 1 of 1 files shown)


srtp/srtp.c
```diff
@@ -676,8 +676,8 @@ static srtp_err_status_t srtp_kdf_generate(srtp_kdf_t *kdf, srtp_prf_label label
 }
 
 static srtp_err_status_t srtp_kdf_clear(srtp_kdf_t *kdf) {
-    memset(kdf->master_key, 0x0, MAX_SRTP_AESKEY_LEN);
-    memset(kdf->master_salt, 0x0, MAX_SRTP_SALT_LEN);
+    octet_string_set_to_zero(kdf->master_key, 0x0, MAX_SRTP_AESKEY_LEN);
```
paulej: This function does not need the 0x0 parameter in the middle. Function signature is:

~~~
octet_string_set_to_zero(void *s, size_t len);
~~~

thisisG: My bad. Thanks! Need to set up a better build-all script. Got OpenSSL but not OpenSSL KDF so that went undetected.

fancycode: It would probably be a good idea to also compile/test these code paths in Travis to detect this automatically in the future.

thisisG: Totally agree. Got stuck on getting KDF to work locally so have not even attempted on getting Travis up to speed yet. :-)

PR #272, opened 2017-03-13
Reduce literal constants

pabuhler:
Use defines for all usages of key length and salt length values.
This attempts to resolve issue #260.

### DIFF
--- a/crypto/cipher/aes_icm.c
+++ b/crypto/cipher/aes_icm.c
@@ -102,12 +102,12 @@ static srtp_err_status_t srtp_aes_icm_alloc (srtp_cipher_t **c, int key_len, int
                 "allocating cipher with key length %d", key_len);
 
     /*
-     * The check for key_len = 30/38/46 does not apply. Our usage
+     * The check for key_len = 30/46 does not apply. Our usage
      * of aes functions with key_len = values other than 30
      * has not broken anything. Don't know what would be the
      * effect of skipping this check for srtp in general.
      */
-    if (key_len != 30 && key_len != 46) {
+    if (key_len != SRTP_AES_ICM_128_KEY_LEN_WSALT && key_len != SRTP_AES_ICM_256_KEY_LEN_WSALT) {
         return srtp_err_status_bad_param;
     }
 
@@ -129,7 +129,7 @@ static srtp_err_status_t srtp_aes_icm_alloc (srtp_cipher_t **c, int key_len, int
     (*c)->state = icm;
 
     switch (key_len) {
-    case 46:
+    case SRTP_AES_ICM_256_KEY_LEN_WSALT:
         (*c)->algorithm = SRTP_AES_ICM_256;
         (*c)->type = &srtp_aes_icm_256;
         break;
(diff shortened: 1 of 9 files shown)


crypto/cipher/aes_icm.c
```diff
@@ -199,8 +199,8 @@ static srtp_err_status_t srtp_aes_icm_context_init (void *cv, const uint8_t *key
 
     copy_len = c->key_size - base_key_len;
     /* force last two octets of the offset to be left zero (for srtp compatibility) */
-    if (copy_len > 14) {
-        copy_len = 14;
+    if (copy_len > SRTP_SALT_LEN) {
```
paulej: Curious, what if the salt is < 14? That's not valid for ICM.

srtp/srtp.c
```diff
@@ -769,13 +769,13 @@ static inline int base_key_length(const srtp_cipher_type_t *cipher, int key_leng
   case SRTP_AES_ICM_256:
     /* The legacy modes are derived from
      * the configured key length on the policy */
-    return key_length - 14;
+    return key_length - SRTP_SALT_LEN;
```
paulej: Several statements that follow return a computed value, but they should be constants.  Why do a computation here?  Is this to help spot errors? Anything checking for errors?

crypto/cipher/aes_icm.c
pabuhler: This should already have been verified I guess so this check should be redundant, it was not my intention to change the meaning of the code only the use of a define. If you find that the code was wrong then that should be changed in a separate PR.

srtp/srtp.c
pabuhler: A case of trying to be consistent but maybe failing.
AES_ICM was a fall through case statement with computed value where as the AES_GCM where individual cases with constant value.  You had previously expressed a dislike for fall though cases so I was reluctant to change it here.
Could change the AES_GCM to also be fall through and used computed value or else change the AES_ICM to be separate cases and used constant values. Do you have a preference?

This whole function is a bit strange, and really should look at where it is called from, why is key_length even needed if we have constant defines for all cipher types. But once again this change was not meant to refactor code just use defines instead of literals.

crypto/cipher/aes_icm.c
paulej: Can't say the check for >14 is wrong, since I assume there might be reasons.  When using GCM, for example, the salt is 12 octets, but when performing RTP header encryption, ICM is used.  So, the user will provide a 128-bit key and a 14-octet salt.  The first 12 octets of the salt are used in GCM and the other two are used in header encryption.  At least I think that's how it was implemented.

But, I'm just unsure of what examples exist where the salt is < 14.  Maybe it's valid, but not sure.  Certainly, I do know ICM needs a 14 octet salt.

srtp/srtp.c
paulej: The reason it exists is that the key and salt are assumed to be concatenated and there are a few places where the two need to be split.  Had the key and salt been separate fields, this function would not be needed.  And the function is useful to avoid a number switch statements in other places.

I don't have a really strong preference, though I do have a slight one toward reducing the number of computations.  That said, this is only invoked when initializing keys and that doesn't happen frequently, so it's not a big deal.

PR #273, opened 2017-03-13
Srtp aead srtcp iv regression tests

thisisG:
Last in the batch of srtcp AES GCM IV regression tests.

Let me know if this is a sane enough way to check that we are generating IVs in an expected manner.
If this works I will start making similar tests for other IV generation methods (and when those makes sense start making tests for the protect/unprotect functions.). 

Do yell if this is all bonkers or requires further clarification! :-) 

Also did a drive-by fixup for a couple of references to IETF drafts (now RFCs).


paulej:
Looks good.

PR #275, opened 2017-03-20
More updates for docs and references

thisisG:
Trying to keep the edits as atomic as possible to ease the review process.

These commits deal a bit with README.md and a bit with inline comments that was previous referencing IETF drafts which have made it in to RFCs.

I had to merge in the current state of the master branch. If there are less messy ways of doing this please let me know. Just disregard the commits that appear to touch the unit tests.


thisisG:
Closing this so I can avoid the in-the-middle-of-everything-merge.

PR #276, opened 2017-03-21
Reference and docs updates

thisisG:
Trying to keep the edits as atomic as possible to ease the review process.

These commits deal a bit with README.md and a bit with inline comments that was previous referencing IETF drafts which have made it in to RFCs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,8 @@ terms outlined in the [License and Disclaimer](#license-and-disclaimer) section.
     less than 32,768; this trick is no longer required as the
     `rdbx_estimate_index(...)` function has been made smarter.
 
+  * The replay window for (S)RTCP is hardcoded to 128 bits in length.
+
 --------------------------------------------------------------------------------
 
 <a name="installing-and-building-libsrtp"></a>
--- a/README.md
+++ b/README.md
@@ -273,45 +273,15 @@ make
 
 The configure script accepts the following options:
 
-Option                                                     | Description
-----------------------------------------------------------|--------------------
-\-\-bindir                                                | user executables (EPREFIX/bin)
-\-\-build                                                 | configure for building on BUILD (guessed)
-\-\-cache-file                                            | cache test results in FILE (disabled)
-\-\-config-cache            \-C                           | alias for `--cache-file=config.cache'
-\-\-datadir                                               | read-only architecture-independent data (DATAROOTDIR)
-\-\-datarootdir                                           | read-only arch.-independent data root (PREFIX/share)
-\-\-disable-debug-logging   \-\-enable-debug-logging      | Enable/disable debug logging in all modules
-\-\-disable-log-stdout      \-\-enable-log-stdout         | Enable/disable logging to stdout
-\-\-disable-openssl         \-\-enable-openssl            | Enable/disable OpenSSL crypto engine
-\-\-disable-openssl-kdf     \-\-enable-openssl-kdf        | Enable/disable OpenSSL KDF algorithm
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -278,16 +273,45 @@ make
...
+\-\-prefix                                                | install architecture-independent files in PREFIX
+\-\-psdir                                                 | ps documentation (DOCDIR)
+\-\-sbindir                                               | system admin executables (EPREFIX/sbin)
+\-\-sharedstatedir                                        | modifiable architecture-independent data (PREFIX/com)
+\-\-silent                  \-\-quiet                 \-q | do not print `checking ...' messages
+\-\-srcdir                                                | find the sources in DIR (configure dir or `..')
+\-\-sysconfdir                                            | read-only single-machine data (PREFIX/etc)
+\-\-version                 \-V                           | display version information and exit
+\-\-without-log-file        \-\-with-log-file             | Use file for logging
+\-\-without-openssl-dir     \-\-with-openssl-dir          | Location of OpenSSL installation
```
pabuhler: Are all these options actually usable? I kind of liked that only libsrtp specific options where documented here. Standard configure options are that interesting if you ask me.

srtp/srtp.c
```diff
@@ -3059,7 +3060,7 @@ void
 srtp_crypto_policy_set_aes_cm_256_hmac_sha1_80(srtp_crypto_policy_t *p) {
 
   /*
-   * corresponds to draft-ietf-avt-big-aes-03.txt
+   * corresponds to RFC 6188 (The Use of AES-192 and AES-256 in Secure RTP)
```
pabuhler: Why mention 192 here? this is 256.

README.md
```diff
@@ -250,9 +250,6 @@ terms outlined in the [License and Disclaimer](#license-and-disclaimer) section.
     less than 32,768; this trick is no longer required as the
     `rdbx_estimate_index(...)` function has been made smarter.
 
-  * The replay window is 128 bits in length, and is hard-coded to this
-    value for now.
-
 --------------------------------------------------------------------------------
```
pabuhler: I think it is still hard coded to 128 for RTCP, not sure if that is worth mentioning.

srtp/srtp.c
thisisG: Will remove it! Agree that it could be confusing.

README.md
thisisG: OK! Will dig a bit in the code. Worth mentioning IMO!

README.md
thisisG: Will have a look and remove the ones that are generic/not related to libSRTP!

README.md
```diff
@@ -459,8 +483,10 @@ September, 2005
 Secure RTP is defined in [RFC 3711](https://www.ietf.org/rfc/rfc3711.txt).
 The counter mode definition is in Section 4.1.1.
 
-SHA-1 is defined in FIPS-180-1, available online at the NIST website.
+SHA-1 is defined in [RFC 3174](https://www.ietf.org/rfc/rfc3174.txt).
```
paulej: We could reference that document, but it's really a copy of FIPS 180-1 (and it admits to such).  The current SHA-1 spec is in FIPS 180-4 (http://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.180-4.pdf).  That would be the most proper reference, I think, but it will be FIPS 180-5 the next time it is revised.

thisisG: Excellent. Thanks! Will update.

PR #278, opened 2017-03-27
Removed crypto/test/auth_driver.c and test/lfsr.c

thisisG:
This is dead code as noticed in PR #276

The TMMH auth was removed in 2013 around 0d3a268


paulej:
Is it code that should be moved to the test directory?

thisisG:
As far as I can tell this code is not used for anything.

`tmmhv2_add_big_test()` for example -> This function does not even exist in git log... The test itself seems to have been accidentally included in the initial commit 235513aa13d973d931202dfcca8b47a4a750a5e4 (after some archaeology).

paulej:
Ok. Yeah, looks a bit useless.

PR #284, opened 2017-03-29
Add strict-prototypes warning if supported.

pabuhler:
Ensure that all functions are correctly declared for c.
There was one function in libsrtp that was not declared correctly.

This flags needed a custom program to check in configure, the built
in function was invalid when using this flag.

### DIFF
--- a/configure
+++ b/configure
@@ -4812,8 +4812,8 @@ fi
 WERROR=""
 for w in -Werror -errwarn; do
   if test "x$WERROR" = "x"; then
-    { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether the compiler supports $w" >&5
-$as_echo_n "checking whether the compiler supports $w... " >&6; }
+    { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether ${CC-c} accepts $w" >&5
+$as_echo_n "checking whether ${CC-c} accepts $w... " >&6; }
     save_cflags="$CFLAGS"
     if test "x$CFLAGS" = "x"; then :
   CFLAGS="$w"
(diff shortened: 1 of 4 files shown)


paulej:
I'll have to take your word that this works as intended.

thisisG:
The change looks more of a headache than it is. The irony here is that the program used to check if `-Wstrict-prototypes` is allowed was incorrectly formed - disabling `-Wstrict-prototypes`. :-)

~~~.bash
└> gcc print.c -Wstrict-prototypes 
print.c:20:5: warning: function declaration isn’t a prototype [-Wstrict-prototypes]
 int main()
     ^~~~
~~~

The only thing I cannot really test (only review by reading) is the windows specific things.

paulej:
Are people using ./configure on Windows?  I just assumed they would use either the VS files or craft their own Makefile for VS.

mjerris:
In FreeSWITCH we currently embed a version of the lib in our tree (we have some specific version requirements).  We have project files built for libsrtp (vs2015) that we use to build.  If thats wanted to be in libsrtp tree, we can clean those up and get them contributed.  Just let me know

paulej:
@mjerris We actually have some VS2015 project files in the repo already.  I've not used them personally, but I assume they work since @JonathanLennox provided them.  In one of my projects, I used Linux to generate a config.h file, then modified it appropriately for Windows and manually included that in the project using VS2013.

Anyway, we should have a clean build for Windows.  I don't care how it gets done.

mjerris:
Thats exactly how we built what we have.  2013 should upgrade cleanly to 2015 no problem.  Haven't messed with 2017.  2013 is probably the right target at this point in time, and having anything else probably doesn't add a lot of value.

pabuhler:
Hi @mjerris , have you tired the project file that is currently checked in https://github.com/cisco/libsrtp/blob/master/srtp2.vcxproj If this differs to your own solution then can you create a new issues/pr with the details as it is off topic for this pr. We should definitely have an up to date windows build/project

Issue #229, opened 2017-01-18
regression 1..5.3 → 2.0.0: Fails testsuite on several archs


jonassmedegaard:
Hi,

In packaging libsrtp 2.0.0 for Debian, the testsuite fails for several architectures: https://buildd.debian.org/status/logs.php?pkg=libsrtp2

(switching to HEAD solved other issues but seemingly didn't change this).

As a comparison, libsrtp 1.5.3 suceeds on most archs covered by Debian: https://buildd.debian.org/status/package.php?p=srtp&suite=experimental

pabuhler:
I had a quit look at one of the failures and it was a test failure and not compile failure. Could this be an endlessness issue in some of the tests? 
I don't think I have access to any of those failing architectures in order to test my self, so can try to review the code based on the build logs you pointed to but would be much better if there was access to a machine to test on.

thisisG:
After discussing this we will not be able to prioritise this for the 2.1 release without further information/access to build machines.

jonassmedegaard:
Quoting Geir Istad (2017-03-22 15:33:27)
> After discussing this we will not be able to prioritise this for the 
> 2.1 release without further information/access to build machines.

Thanks for the status update - and sorry for my late response:

We have in Debian a procedure for granting temporary guest access to 
machines of each architecture: https://dsa.debian.org/doc/guest-account/ 
- in particular the section labelled "non-DMs".

I would be happy to sponsor requests from some of you guys!

You can reach me directly at email dr@jones.dk

 - Jonas

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

thisisG:
Excellent! Will do. Might not make it for the 2.1 release but will have to see how much work there is to be done.

thisisG:
Small update on this issue: 

Some failures are due to configuration not setting the correct shared library suffix.

Some failures (At least the mips test failure) seems to be inconsistent operations on bitfields depending on the context on (some) big endian machine. Need to find out if we're doing something wrong logically or hitting undefined/unspecified behaviour (which we could argue is also us doing something wrong :) ).

pabuhler:
@jonassmedegaard  Hi, is it possible to to get a new build against current master? We believe that most of the issues should be fixed now and it would be great to resolve this issue.

jonassmedegaard:
[...]

PR #285, opened 2017-03-29
test_srtp.c: Use #define instead of static const

thisisG:
Prevents generating warnings with clang (warning: variable length array
folded to constant array as an extension [-Wgnu-folding-constant])

This is also a better alternative from a portability point of view.


thisisG:
Did the #undef just for defensive programming as the value only makes sense in the context of that function. :-)

Did not know before now that enum was C89/C90, thought it was C99! If I revisit this code in the future I might change it. Will see what happens when we get access to more esoteric architectures (Re #229) so we can improve the portability of the code.

PR #287, opened 2017-03-30
Fix OOB read in key generation for encrypted headers with GCM ciphers.

### LINKED ISSUE #286: OOB read in key generation for encrypted headers with GCM
According to [RFC 7714 section 8.3](https://tools.ietf.org/html/rfc7714#section-8.3) GCM ciphers must use their corresponding Counter Mode ciphers for header encryption (i.e. AES-256-GCM uses AES-256-CM).

This is a problem in the current implementation because the salt of the CM cipher is 16 bits larger than the salt of the GCM cipher (e.g. AES-256-GCM uses a 256-bit key with a 96-bit salt, AES-256-CM uses a 256-bit key with a 112-bit salt).

The OOB happens where the CM key+salt are copied from the GCM key+salt:
https://github.com/cisco/libsrtp/blob/5f1642c36731b7199ba34b79676b4c02cd4142c9/srtp/srtp.c#L1010
(`rtp_xtn_hdr_base_key_len` is the size of the GCM/CM key, `rtp_xtn_hdr_salt_len` is the size of the CM salt, but `key` only contains the key+salt of the GCM cipher).
To verify, access the last byte read from `key` at that position and run in some memory checker like https://clang.llvm.org/docs/MemorySanitizer.html).

I would like to submit a PR to fix this, but I'm wondering what the correct solution is. I'm assuming it's the same as described in the errata by @paulej at https://www.rfc-editor.org/errata_search.php?rfc=7714 and the salt should be multiplied by 2^16.

@JonathanLennox you provided feedback in that area on my initial implementation in #128, do you know what the right behaviour should be?

fancycode:
The salt of the GCM cipher is shorter than the salt required for the ICM cipher which is used for the encrypted headers. This should fix #286.

There is no RFC refence data for GCM with header encryption, so that was not detected. The testcase `srtp_validate_encrypted_extensions_headers_gcm` is checking a manually constructed packet which relied on the incorrect behaviour. This testcase has also been adjusted for the fixed behaviour.


srtp/srtp.c
```diff
@@ -1006,6 +1006,10 @@ srtp_stream_init_keys(srtp_stream_ctx_t *srtp, srtp_master_key_t *master_key,
       rtp_xtn_hdr_base_key_len = base_key_length(session_keys->rtp_xtn_hdr_cipher->type,
                                                  rtp_xtn_hdr_keylen);
       rtp_xtn_hdr_salt_len = rtp_xtn_hdr_keylen - rtp_xtn_hdr_base_key_len;
+      if (rtp_xtn_hdr_salt_len > rtp_salt_len) {
```
paulej: This would be correct if session_keys->rtp_cipher->type == SRTP_AES_GCM_128 or SRTP_AES_GCM_256, but might not be correct for another cipher.  Presently, we do not have this issue with other ciphers, but I still think it might be better to have an explicit check here so people in the future will understand why this comparison is done.

fancycode: Updated, is that what you had in mind?

srtp/srtp.c
```diff
@@ -1006,6 +1006,17 @@ srtp_stream_init_keys(srtp_stream_ctx_t *srtp, srtp_master_key_t *master_key,
...
+        switch (session_keys->rtp_cipher->type->id) {
+        case SRTP_AES_GCM_128:
+        case SRTP_AES_GCM_256:
+          /* The shorter GCM salt is padded to the required ICM salt length. */
+          rtp_xtn_hdr_salt_len = rtp_salt_len;
+          break;
+        default:
+          return srtp_err_status_bad_param;
```
thisisG: (comment aside from the issue @paulej raised)
Do remember to clear the tmp key here before returning! :-)

~~~.c
default:
    /* zeroize temp buffer */
    octet_string_set_to_zero(tmp_key, MAX_SRTP_KEY_LEN);
~~~

fancycode: Thanks for the heads up, changed.

srtp/srtp.c
paulej: That's even better than what I had in mind.  I was going to take the lazy route and just make a more complex if statement, but I appreciate the switch statement that then results in an explicit error when we have a cipher mismatch and it's not due to GCM.

PR #289, opened 2017-04-03
Added support for set and get the roll-over-counter

ulfolsson:
First draft

### DIFF
--- a/crypto/include/rdbx.h
+++ b/crypto/include/rdbx.h
@@ -85,7 +85,6 @@ typedef uint64_t srtp_xtd_seq_num_t;
 typedef struct {
     srtp_xtd_seq_num_t index;
     bitvector_t bitmask;
-    uint32_t pending_roc;
 } srtp_rdbx_t;
 
 
--- a/crypto/include/rdbx.h
+++ b/crypto/include/rdbx.h
@@ -85,6 +85,7 @@ typedef uint64_t srtp_xtd_seq_num_t;
 typedef struct {
     srtp_xtd_seq_num_t index;
     bitvector_t bitmask;
+    int is_roc_set;
 } srtp_rdbx_t;
 
 
--- a/crypto/include/rdbx.h
+++ b/crypto/include/rdbx.h
@@ -85,7 +85,7 @@ typedef uint64_t srtp_xtd_seq_num_t;
 typedef struct {
     srtp_xtd_seq_num_t index;
     bitvector_t bitmask;
-    int is_roc_set;
+    uint32_t pending_roc;
 } srtp_rdbx_t;
 
 
@@ -148,7 +148,7 @@ srtp_err_status_t srtp_rdbx_add_index(srtp_rdbx_t *rdbx, int delta);
 srtp_err_status_t srtp_rdbx_set_roc(srtp_rdbx_t *rdbx, uint32_t roc);
 
 /*
- * srtp_rdbx_get_roc(rdbx) returns the value of the rollover counter for
+ * srtp_rdbx_get_packet_index(rdbx) returns the value of the rollover counter for
  * the srtp_rdbx_t pointed to by rdbx
  *
  */
@@ -186,6 +186,24 @@ void srtp_index_advance(srtp_xtd_seq_num_t *pi, srtp_sequence_number_t s);
  */
 int32_t srtp_index_guess(const srtp_xtd_seq_num_t *local, srtp_xtd_seq_num_t *guess, srtp_sequence_number_t s);
 
+/*
+ * srtp_rdbx_get_roc(rdbx)
+ *
+ * Get the current rollover counter
+ *
+ */
+uint32_t srtp_rdbx_get_roc(const srtp_rdbx_t *rdbx);
+
+/*
+ * srtp_rdbx_set_roc_seq(rdbx, roc, seq) initalizes the srtp_rdbx_t at the
+ * location rdbx to have the rollover counter value roc and packet sequence
+ * number seq.  If the new rollover counter value is less than the current
+ * rollover counter value, then the function returns
+ * srtp_err_status_replay_old, otherwise, srtp_err_status_ok is returned.
+ */
+srtp_err_status_t srtp_rdbx_set_roc_seq (srtp_rdbx_t *rdbx,
+                                         uint32_t roc,
+                                         uint16_t seq);
 
 #ifdef __cplusplus
 }
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -292,7 +292,9 @@ typedef enum {
   srtp_err_status_encode_err   = 22, /**< error encoding data                     */
   srtp_err_status_semaphore_err = 23,/**< error while using semaphores            */
   srtp_err_status_pfkey_err    = 24, /**< error while using pfkey                 */
-  srtp_err_status_bad_mki      = 25  /**< error MKI present in packet is invalid */
+  srtp_err_status_bad_mki      = 25, /**< error MKI present in packet is invalid  */
+  srtp_err_status_pkt_idx_old  = 26, /**< packet index is too old to consider     */
+  srtp_err_status_pkt_idx_adv  = 27 /**< packet index advanced, reset needed      */
 } srtp_err_status_t;
 
 
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -1855,6 +1855,29 @@ srtp_err_status_t srtp_get_protect_rtcp_trailer_length(srtp_t session, uint32_t 
 
 
 /**
+ * @brief srtp_set_stream_roc(session, ssrc, roc)
+ *
+ * Set the roll-over-counter on a session for a given SSRC
+ *
+ * returns err_status_ok on success, srtp_err_status_bad_param if there is no
+ * stream found
+ *
+ */
+srtp_err_status_t srtp_set_stream_roc(srtp_t session, uint32_t ssrc, uint32_t roc);
+
+/**
+ * @brief srtp_get_stream_roc(session, ssrc, roc)
+ *
+ * Get the roll-over-counter on a session for a given SSRC
+ *
+ * returns err_status_ok on success, srtp_err_status_bad_param if there is no
+ * stream found
+ *
+ */
+srtp_err_status_t srtp_get_stream_roc(srtp_t session, uint32_t ssrc, uint32_t *roc);
+
+
+/**
  * @}
  */
 /* in host order, so outside the #if */
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -2251,8 +2251,14 @@ srtp_unprotect_mki(srtp_ctx_t *ctx, void *srtp_hdr, int *pkt_octet_len,
     }
   } else {
   
-    /* estimate packet index from seq. num. in header */
-    delta = srtp_rdbx_estimate_index(&stream->rtp_rdbx, &est, ntohs(hdr->seq));
+    if (stream->rtp_rdbx.is_roc_set) {
+        est = (srtp_xtd_seq_num_t) ntohs(hdr->seq) + stream->rtp_rdbx.index;
+        delta = (int)est;
+        stream->rtp_rdbx.is_roc_set = 0;
+    } else {
+        /* estimate packet index from seq. num. in header */
+        delta = srtp_rdbx_estimate_index(&stream->rtp_rdbx, &est, ntohs(hdr->seq));
+    }
     
     /* check replay database */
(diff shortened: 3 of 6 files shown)


include/srtp.h
```diff
@@ -1855,6 +1855,29 @@ srtp_err_status_t srtp_get_protect_rtcp_trailer_length(srtp_t session, uint32_t
 
 
 /**
+ * @brief srtp_set_stream_roc(session, ssrc, roc)
```
paulej: Please document the parameters.  I'd personally prefer the ssrc to be provided in host byte order.  (We have one function that assumes network order and I think that's really unfortunate.)

pabuhler:
@paulej   srtp_get_est_pkt_index & srtp_estimate_index from the EKT branch seam a little complicated, at least for this use case??

I would suggest that the all this functionality was encapsulated inside of rdbx to make testing easy and the code in srtp.c cleaner. Ie is_roc_set should not be accessed outside of the rdbx.c . The rtp_rdbx_estimate_index function should take into account the value of is_roc_set and only once the packet has been successfully authenticated should is_roc_set be cleared by the srtp_rdbx_add_index function. How does that sound?

srtp/srtp.c
```diff
@@ -4461,3 +4467,27 @@ srtp_err_status_t srtp_install_log_handler(srtp_log_handler_func_t func, void *
...
+srtp_err_status_t
+srtp_get_stream_roc(srtp_t session, uint32_t ssrc, uint32_t *roc) {
+    srtp_stream_t stream;
+
+    stream = srtp_get_stream(session, htonl(ssrc));
+    if (stream == NULL)
+        return srtp_err_status_bad_param;
+
+    *roc = stream->rtp_rdbx.index >> 16;
+
```
pabuhler: Could either use srtp_rdbx_get_packet_index or even better add a srtp_rdbx_get_roc that does the shifting.

ulfolsson: I fixed it with a new function, srtp_rdbx_get_roc. Let us wait for Paul's comments on how to proceed

paulej:
@pabuhler I didn't mean to suggest borrowing all that logic, since part of it was just to allow for testing packet authentication before setting the ROC.  But, where I was going with that is that (if memory serves) once the packet authenticates, the ROC is assigned and the replay database is reset (if memory serves).

I'm a little confused as to why we need to have the "is_roc_set" at all.  If we call a function to set the ROC, then set it and reset the replay database.

ulfolsson:
OK, a bit confused myself now. I changed and tested the code from the EKT branch and that works as well

pabuhler:
@ulfolsson, what are you using to test? Would be great to have some tests covering this behavior.
Have you tested setting the ROC to 1 and then the first packet arriving has a seq of 60000?

ulfolsson:
@pabuhler, You mean that 60000 packets are sent and then receiver incorrectly sets the ROC to 1?

pabuhler:
@ulfolsson,  I mean that you join a conference where the seq num has already wrapped once so ROC is 1 and current seq num is 60000. So from libsrtp's prospective you first call set_roc(1) and then the first packet that is processed has seq num of 60000.

ulfolsson:
@pabuhler, I did the following test. Let the sender create X packets. Then set the ROC on the receiver and unprotect the last sent packet i.e

packets_to_send = 65535;
test_func(packets_to_send-1, roc=0); ==>  first receiver seq 65534
test_func(packets_to_send, roc=0);   ==>  first receiver packet seq 65535
test_func(packets_to_send+1, roc=1); ==>  first receiver packet seq 0
test_func(packets_to_send+60001, roc=1); ==>  first receiver packet seq 60000

pabuhler:
@ulfolsson , OK, if these test pass then I guess it covers basic usage. Will you update the PR, including tests?

ulfolsson:
@pabuhler, One last question. Should I put the tests in a new driver or?

pabuhler:
@ulfolsson  that would depend on the test, without seeing them it is hard to tell. If you are testing through the srtp api then you have the options of test/srtp_driver.c or test/test_srtp.c . You will have to find the best fit, or else create something new.

ulfolsson:
It has been a couple of days since I pushed my latest changes so I just wanted to check if you are missing something

paulej:
Only my time to review the code.  For whatever reason, I didn't get notified of the latest push. @pabuhler asked me to review it, but I didn't have time today. Hopefully this week.

ulfolsson:
@paulej, OK thanks.  Just wanted to check that I had not missed anything. Easter is here now in Europe so no rush :)

srtp/srtp.c
```diff
@@ -1532,6 +1532,110 @@ srtp_get_session_keys(srtp_stream_ctx_t *stream, uint8_t* hdr,
...
+#ifdef NO_64BIT_MATH
+    internal_roc = (uint32_t)(rdbx->index >> 16);
+    roc_difference = roc - internal_roc;
+    if (roc_difference > 1) {
+      *delta = 0;
+      return srtp_err_status_pkt_idx_adv;
+    }
+
+    internal_pkt_idx_reduced = (uint32_t)(rdbx->index & 0xFFFF);
+    external_pkt_idx_reduced = (uint32_t)((roc_difference << 16) | seq);
```
paulej: Why is these "_reduced" variables not 64-bit values?  What if I set the ROC to be some much larger number, like 0xFFFFFF00, when the original ROC was 0?

ulfolsson: @paulej,

<Perhaps I overlooked it, but what happens if you attempt to set the ROC value 
<prior to encrypting a packet? Did I miss the logic that handles that flow?

I did not add it since I had no use case for it my self. 
But it is added now and I will push it tomorrow plus some tests for it

<Why are these "_reduced" variables not 64-bit values? What if I set the ROC to 
<be some much larger number, like 0xFFFFFF00, when the original ROC was 0?

I have to admit that I just took the code from the EKT branch without studying it too much

srtp/srtp.c
```diff
@@ -4461,3 +4598,29 @@ srtp_err_status_t srtp_install_log_handler(srtp_log_handler_func_t func, void *
...
 }
+
+srtp_err_status_t
+srtp_set_stream_roc(srtp_t session, uint32_t ssrc, uint32_t roc) {
+    srtp_stream_t stream;
+
+    stream = srtp_get_stream(session, htonl(ssrc));
+    if (stream == NULL)
+        return srtp_err_status_bad_param;
+
```
pabuhler: Could potentially check if the current roc is the same as the new roc and so do nothing.... could argue that the application could do the same check.

srtp/srtp.c
```diff
@@ -2037,18 +2142,32 @@ srtp_protect_mki(srtp_ctx_t *ctx, void *rtp_hdr, int *pkt_octet_len,
...
-     if (status != srtp_err_status_replay_fail || !stream->allow_repeat_tx)
-       return status;  /* we've been asked to reuse an index */
-   }
-   else
-     srtp_rdbx_add_index(&stream->rtp_rdbx, delta);
+  /*
+   * estimate the packet index using the start of the replay window
+   * and the sequence number from the header
+   */
+  status = srtp_get_est_pkt_index(hdr,
```
pabuhler: Is all this really needed in the send direction? I would have thought when sending the new ROC should just be used as is.

srtp/srtp.c
```diff
@@ -2037,18 +2142,32 @@ srtp_protect_mki(srtp_ctx_t *ctx, void *rtp_hdr, int *pkt_octet_len,
...
+
+  if (advance_packet_index) {
+    srtp_rdbx_set_roc_seq(&stream->rtp_rdbx,
+                          (uint32_t)(est >> 16),
+                          (uint16_t)(est & 0xFFFF));
+    srtp_rdbx_add_index(&stream->rtp_rdbx, 0);
+  } else {
+    status = srtp_rdbx_check(&stream->rtp_rdbx, delta);
+    if (status)
+      return status;
```
pabuhler: The old code had a check for `status != srtp_err_status_replay_fail || !stream->allow_repeat_tx` this is missing here, any reason?

srtp/srtp.c
```diff
@@ -1532,6 +1532,110 @@ srtp_get_session_keys(srtp_stream_ctx_t *stream, uint8_t* hdr,
   return NULL;
 }
 
+static srtp_err_status_t
+srtp_estimate_index(srtp_rdbx_t *rdbx,
```
pabuhler: Took me a while to understand this function, some comments would have been nice. The NO_64BIT_MATH does not make it easier. In the end I think this function is correct at least for the case when there is 64bit maths.

srtp/srtp.c
pabuhler: @paulej  doesn't the roc_difference check avoid that problem?

paulej: It seems to me that if I set the ROC equal to 0xFFFFFFFF and the original value was 0x00000000, we're going to overflow those uint32 values.

pabuhler: @paulej I still think it is protected by this code just a couple of lines further up?
0xFFFFFFFF - 0x00000000 is greater than 1 right?
```
internal_roc = (uint32_t)(rdbx->index >> 16);
roc_difference = roc - internal_roc;
if (roc_difference > 1) {
    *delta = 0
    return srtp_err_status_pkt_idx_adv;
}
```

paulej: You're right. I overlooked that. But do we not want to allow ROC to be set to something much higher?  In a video conference call, for example,  a late-joining participant might receive the ROC in an EKT field and that ROC might be any value greater than zero.

pabuhler: If the difference is greater than 1 it is accepted, returns srtp_err_status_pkt_idx_adv, only if the difference is less than or equal to 1  does it try to determine if the current seq num should use new roc or old roc. ... If you ignore (remove) all of the #ifdef NO_64BIT_MATH stuff then the intention of the function becomes clearer. Then you can just verify that the stuff in side of the #ifdef NO_64BIT_MATH check does the same.  This function should have had some comments and the #ifdef NO_64BIT_MATH stuff should be abstracted out of srtp in general so it makes code cleaner to read.

ulfolsson: Maybe the code would be easier to read we split it in

#ifdef NO_64BIT_MATH
srtp_estimate_index_no_64bit_math()
#else
srtp_estimate_index()
#endif

crypto/include/rdbx.h
```diff
@@ -85,6 +85,7 @@ typedef uint64_t srtp_xtd_seq_num_t;
 typedef struct {
     srtp_xtd_seq_num_t index;
     bitvector_t bitmask;
+    uint32_t pending_roc;
```
ulfolsson: OK, I will try to make one push for each comment

srtp/srtp.c
ulfolsson: My mistake, copy and paste error

srtp/srtp.c
ulfolsson: Maybe there is a simpler way but  srtp_get_est_pkt_index helps you to get the estimated index and delta correct. For example if you start with roc=1 and seqnr=0xa8ca you want
estimated packet index: 000000000001a8ca delta=0
and not
estimated packet index: 000000000000a8ca delta=0xa8ca

srtp/srtp.c
pabuhler: will you push a change for this?

srtp/srtp.c
```diff
@@ -2037,18 +2148,36 @@ srtp_protect_mki(srtp_ctx_t *ctx, void *rtp_hdr, int *pkt_octet_len,
...
+                          (uint32_t)(est >> 16),
+                          (uint16_t)(est & 0xFFFF));
+    stream->pending_roc = 0;
+    srtp_rdbx_add_index(&stream->rtp_rdbx, 0);
+  } else {
+    status = srtp_rdbx_check(&stream->rtp_rdbx, delta);
+    if (status) {
+      return status;
+      if (status != srtp_err_status_replay_fail || !stream->allow_repeat_tx)
+        return status; /* we've been asked to reuse an index */
```
pabuhler: um @ulfolsson , I don't think this is what you really mean or?
One to many return statements.

ulfolsson: No, of course not :) New patch soon

paulej:
OK, I guess I was just misreading the code related to advancing the ROC.  I took another quick look and it would appear that srtp_rdbx_set_roc_seq() is going to prevent setting the ROC to a lower value.  I have no business case for doing that.  I assume others are also OK with that limitation?

Otherwise, I think it looks OK ... so long as all existing tests do pass, as @pabuhler mentioned.

PR #291, opened 2017-04-12
Use const char * for srtp_set_debug_module()

pabuhler:
Not sure why it was not const, but it makes calling it a pain.


pabuhler:
@paulej is LGTM the same as approved...
before I could not call 
`srtp_set_debug_module("srtp", 1);` 
instead had to call something like
`char srtpMod[] = "srtp";`
`srtp_set_debug_module(srtpMod, 1);`

paulej:
Yeah. Merge it.

Issue #299, opened 2017-05-02
libsrtp-2.0.0 # make runtest  ./rtpw: couldn't open file /usr/share/dict/words make: *** [runtest] Error 255


mariarti:
```
# make runtest
Build done. Please run 'make runtest' to run self tests.
running libsrtp2 test applications...
crypto/test/cipher_driver -v >/dev/null
crypto/test/kernel_driver -v >/dev/null
test/rdbx_driver -v >/dev/null
test/srtp_driver -v >/dev/null
test/roc_driver -v >/dev/null
test/replay_driver -v >/dev/null
test/dtls_srtp_driver >/dev/null
cd test; /root/asterisk/libsrtp-2.0.0/test/rtpw_test.sh >/dev/null	
./rtpw: couldn't open file /usr/share/dict/words
make: *** [runtest] Error 255
```
centos 6.9

paulej:
I think that last error is just due to the "spell" or "words" package missing from the machine.

pabuhler:
This should be fixed now by #300

PR #300, opened 2017-05-03
Standalone tests

itollefsen:
This is something I've been sitting on for a while that re-surfaced with issue #299.

I'd like the tests to be standalone, in the sense that they should not depend on files that may not be on the system. In this case, the dictionary file. On Windows, there was references to a non-existing "words.txt" file.

So this takes the dictionary file from a Linux system that had it and adds it to the repository as "words.txt". Then it changes tests that used them to always use the included "words.txt" file instead of going off and looking for a dictionary file.

The reason I've been sitting on this is because the dictionary file, to me, seems too big. However, without intimate knowledge of the tests and what they're supposed to test, I can't really say what sort of data set it requires.

Therefore, it would be great if someone could comment on that part and perhaps come up with a more suitable data set, one that isn't this big and could more easily be added to the repository.

The second part of this is some test script changes to make the tests work as intended on Cygwin/Windows as well.


paulej:
That's worrisome from a copyright perspective.

itollefsen:
> That's worrisome from a copyright perspective.

Probably that too. Yet another reason why it would be better to define a smaller, custom, data set for the tests.

I'm slightly curious about the original reference to "words.txt" which was used on Windows. That seems to indicate that someone, at some point, had such a data set defined and used it on Windows.

paulej:
My guess is that somebody just put something in place when testing the config. Probably not much mystery to it.

itollefsen:
Updated to use a list of words generated from a random word generator.

The list now contains 250 words. Far less than the full dictionary, but much more manageable with regards to having it as part of the repository.

Issue #257, opened 2017-02-22
Warnings compiling on FreeBSD


paulej:
The -pedantic compiler switch results in a number of compilation warnings.  We should investigate why and address the issues.

pabuhler:
I have had some second thoughts about -pedantic, need to either be "pedantic" in our coding including comments or else start adding some exceptions for different compilers.

paulej:
I did not explore all of the issues I saw, but some could be resolved by changing the code to conform to a specific C version.  Others I wasn't sure why the compiler was complaining, so I can't say what the issue was.

We might want to pass to the compiler the --std parameter.  I'm not sure what the compiler assumes by default.

thisisG:
With gcc 6.3.1 it always fails compilation if any -std=cNN is specified. 

I thought it was left undefined in order to be "as portable as possible", especially if looking at how many different systems we try to accommodate for in `configure` as well as optional includes in for example `integers.h` to try to get c99(ish) types.

pabuhler:
@paulej ,are there still a lot of warnings on FreeBSD ?

paulej:
Not with gcc 7, at least.

PR #303, opened 2017-05-11
Makefile.in: Add gnu as match for shared lib suffix

thisisG:
This will fix issues as reported in #229 where we fail to complete builds
because the libary suffix variable is not set.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -74,7 +74,7 @@ pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = libsrtp2.pc
 
 SHAREDLIBVERSION = 1
-ifeq (linux,$(findstring linux,@host@))
+ifneq (,$(or $(findstring linux,@host@), $(findstring gnu,@host@)))
 SHAREDLIB_DIR = $(libdir)
 SHAREDLIB_LDFLAGS = -shared -Wl,-soname,$@
 SHAREDLIBSUFFIXNOVER = so


pabuhler:
@thisisG  I take it you actually tested this and it worked :-)

thisisG:
@pabuhler - yes.
Failures of this type [1] seems to be fixed with this extra check\* :

```
gcc -shared -o libsrtp2.  \
			srtp/srtp.o srtp/ekt.o crypto/cipher/cipher.o crypto/cipher/null_cipher.o crypto/cipher/aes_icm.o crypto/cipher/aes.o crypto/hash/null_auth.o crypto/hash/auth.o crypto/hash/hmac.o crypto/hash/sha1.o crypto/math/datatypes.o crypto/math/stat.o crypto/kernel/crypto_kernel.o crypto/kernel/alloc.o crypto/kernel/key.o crypto/kernel/err.o crypto/replay/rdb.o crypto/replay/rdbx.o crypto/replay/ut_sim.o -L. -specs=/usr/share/dpkg/pie-link.specs -Wl,-z,relro -fPIC -lpcap 
if [ -n "1" ]; then \
	ln -sfn libsrtp2. libsrtp2.; \
fi
ln: 'libsrtp2.' and 'libsrtp2.' are the same file
Makefile:132: recipe for target 'libsrtp2.' failed
make[1]: *** [libsrtp2.] Error 1
make[1]: Leaving directory '/«BUILDDIR»/libsrtp2-2.0.0+20170223'
/usr/share/cdbs/1/class/makefile.mk:77: recipe for target 'debian/stamp-makefile-build' failed
make: *** [debian/stamp-makefile-build] Error 2
dpkg-buildpackage: error: debian/rules build-arch gave error exit status 2
```

[1] https://buildd.debian.org/status/logs.php?pkg=libsrtp2&arch=kfreebsd-amd64&suite=experimental
\* Could argue that just checking for gnu would work for linux as well - but I preferred keeping the check there

itollefsen:
I've never worked with kfreebsd. As I understand it, it's a BSD kernel with Debian GNU user land and packaging system.

While this patch works because kfreebsd identifies itself as `x86_64-pc-kfreebsd-gnu`, it's not actually Linux - it's BSD. It only works because the two use the same suffix. It would not work if the two needed different linker flags for example. Also, without having checked, I believe this change will match on GNU/Hurd systems as well, which I know nothing about - let alone their shared library suffix (although that may not be relevant).

As far as I can tell, there's no explicit support for BSD in that `Makefile.in` at all? But it does actually compile on BSD (see issue #257 for example), but that might just be the static version?

So I wonder if it would be better to add explicit support for BSD rather than tacking on the "gnu" match here.

In the longer run, I'd like to try full use of Automake and libtool to avoid this sort of issues. There seems to have been a concerted effort to avoid them (for more or less understandable reasons), but they would solve this and future similar cases.

pabuhler:
Ok lets use this for now, and if needed to add better support for BSD or other platforms we can in the future.

PR #304, opened 2017-05-13
Fix (S)RTP for big endian machines

thisisG:
srtp.h was included in srtp.c before config.h was included. As a result of this the WORDS_BIGENDIAN define was not set, and srtp_hdr_t would be defined as per little endian implementation internally in srtp.c.
The srtp_driver included srtp_priv directly and had the right structure for big endian set. The tests failed because one part of the system had a different bitfield structure compared to the other.
  
This is a fix for issue #229 for big endian machines.

In addition there are several commits for removing and cleaning up files that are not required in order to use the public API as well as an update to the docs to reflect the change in removed files.


include/srtp.h
```diff
@@ -122,52 +122,6 @@ extern "C" {
...
-/*
- * an srtp_hdr_t represents the srtp header
- *
- * in this implementation, an srtp_hdr_t is assumed to be 32-bit aligned
- * 
- * (note that this definition follows that of RFC 1889 Appendix A, but
- * is not identical)
- */
- 
-#ifndef WORDS_BIGENDIAN
```
ukreator: there are also SRTCP structs in this file branched with the same WORDS_BIGENDIAN

thisisG: WillFix! Thanks!

PR #308, opened 2017-05-29
srtp.c: Save the ROC and sequence number before usage

ulfolsson:
The ROC and the sequence number to set must be calculated
before the estimated index is used otherwise the packets
following the first one can't be decoded/authenticated

Change-Id: Ib2950b37771d39607fdead33d32245fa08fb0ab1


pabuhler:
I guess this was due to est being converted to network byte order, it may have been better to not convert est but  instead make a copy. Also a shame that the exsisitng code uses a function called be64_to_cpu to convert to network byte order!

Issue #254, opened 2017-02-21
Standardize indentation style in libsrtp


JonathanLennox:
The code indentation style in libsrtp has ended up extremely inconsistent -- two-space, four-space, three-space, leading space on the left before function declarations, etc.  This makes reading it very confusing. I don't much care what form it uses, but it should be consistent.

pabuhler:
Perfect, we have discussed this and decided to wait for after 2.1 release, but it is great that it is now an official issues :)

pabuhler:
Reopen this issues as it was not as easily fixed as I thought. After trying to perform all formatting in on PR it was obvious the reviewing it all was difficult. We will now try to do it in a series of small PR's to make the review easier.

PR #311, opened 2017-06-12
Add support for the double GCM transform

bifurcation:
The IETF [PERC WG](https://tools.ietf.org/wg/perc) is working on defining a "double GCM" SRTP transform.  The transform uses two invocations of the AES-GCM transform to provide "end to end" and "hop by hop" protections, with the idea that an intermediary with only the "hop by hop" key can perform some limited modifications to protected packets, but not read the payload.  This PR implements the current draft of that transform in `libsrtp`.

**Note:** The specification for the transform is still evolving, so it is not appropriate to merge this PR quite yet.


fluffy:
I think all the white space changes should get moved out to a separate PR as it makes this hard to review. If someone has a good clang-format settings we could just run that over all the code to clean up the whitespace.

thisisG:
@fluffy - yeah that is planned to do soon after the 2.1 release (tracked in issue #254 ) :-)

pabuhler:
@bifurcation can you try to rebase this on master now. Master has had code formatting applied so there should be no need for any of the white space changes.

Issue #313, opened 2017-06-13
wrong packet length in call to srtp_protect() in srtp_driver.c


carlh54:
This problem appears in the functions srtp_test() and srtcp_test(). I will describe the problem in srtp_test() as the exact same problem exists in srtcp_test().

In the function srtp_test() on line 857, the variable len is loaded from the variable 'msg_len_octet' which is set to 28 which is the payload length and doesn't take into account the RTP header size of 12. The variable 'len' is then incorrectly used in the call to srtp_protect().

![capture](https://user-images.githubusercontent.com/25154348/27102729-50da91b0-5054-11e7-985e-adbd2af291e8.JPG)

Also at line 952 the entire un-protected packet is not being verified as it is using 'msg_len_octet' which is only 28 and doesn't take into account the additional RTP header size. 

![capture1](https://user-images.githubusercontent.com/25154348/27103083-770f43c0-5055-11e7-802d-a34697d5c2bd.JPG)

pabuhler:
Hi and thanks. You are correct these tests are not really doing what they should. I tried to address this in PR #314 but that exposed further weakness in the test, and the fact that they where incorrectly passing on the travis build machine.  Since this will require a change to our travis script and so far it is only bugs in the test and not the actual code I think I will close #314 and re submit against master.

carlh54:
You're welcome.

PR #314, opened 2017-06-14
Return pkt_len from srtp_create_test_packet_xxx

pabuhler:
Previously the pkt_len was calculated outside the function,
this was incorrect or missing in some cases.

Also mki was not always taken into account when calculating tag
length.

This should address PR #313


pabuhler:
Closing this, will reapply against master.

PR #318, opened 2017-07-03
Check explicitly for openssl version >= 1.0.1d

pabuhler:
Prior to 1.0.1d EVP_aes_128_gcm was broken.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 sudo: false
 language: c
 cache: ccache
--- a/configure
+++ b/configure
@@ -5544,6 +5544,43 @@ See \`config.log' for more details" "$LINENO" 5; }
 fi
 
 
+   { $as_echo "$as_me:${as_lineno-$LINENO}: checking whether EVP_aes_128_gcm() is broken" >&5
+$as_echo_n "checking whether EVP_aes_128_gcm() is broken... " >&6; }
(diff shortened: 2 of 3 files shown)


mjerris:
you can never depend on the version for things like this... distros will routinely patch things without changing version.  Detecting the actual working feature in this case would be much better.  This probably also breaks working with libressl

itollefsen:
The issue with detecting a actual working feature is that it requires the use of AC_TRY_RUN. And that doesn't fly when cross compiling.

I'm not overly worried about distros patching things without updating their version number. For that to be a problem, you'd have to start with a version prior to 1.0.1d and patch this specific issue. I don't have numbers to back it up, but a hunch that you won't freqeunlty encounter that particular combination.

I'm far more concerned by the LibreSSL issue. If it should work with LibreSSL, then this needs some tweaking.

If AC_TRY_RUN is to be used, then I'd still consider keeping what this patch does as a fallback when cross compiling.

And/or add a override mechanism that would disable this particular check.

pabuhler:
I different options is to move it from a compile/configure time check to a run time check that is caught by `make runtest` or by the self test of the gsm implementation when the library is loaded.
PR #320  just modified the tests and now exposes this issue, hence it currently failing on Travis. If the self test was also modified to trigger the issue then srtp_init() would fail.

pabuhler:
make runtest will now catch this so that should be good enough. Had tried to add it to the self test of the cipher but that proved difficult to reproduce.

PR #319, opened 2017-07-04
Clang format suggestion

### LINKED ISSUE #254: Standardize indentation style in libsrtp

thisisG:
Here is a suggestion of a clang-format file (and the results of applying this formatting to two functions) in order to resolve #254 .

If someone prefers wider column limits or different style of braces let me know and I will attempt to tune the format file! I am happy as long as we standardize on something. :-)

If this looks OK I will apply it to the project files and go through and manually fix up for example comments that have been split across multiple lines and other nitpicks and make a new PR.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,77 @@
+AccessModifierOffset: -4
+
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: false
+AlignOperands:   true
+AlignTrailingComments: true
+
+AllowAllParametersOfDeclarationOnNextLine: false
+
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: false
+
+BinPackArguments: true
+BinPackParameters: false
+BraceWrapping:
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: true
+
+ColumnLimit:     80
+CommentPragmas:  '^ IWYU pragma:'
+ContinuationIndentWidth: 4
+DerivePointerAlignment: false
+DisableFormat:   false
+ExperimentalAutoDetectBinPacking: false
+
+IndentCaseLabels: false
+IndentWidth:     4
+IndentWrappedFunctionNames: false
+
+KeepEmptyLinesAtTheStartOfBlocks: false
+
+MaxEmptyLinesToKeep: 1
+
+PenaltyBreakBeforeFirstCallParameter: 16
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000
+PenaltyReturnTypeOnItsOwnLine: 60
+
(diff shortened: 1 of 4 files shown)


pabuhler:
Good start, I am not to concerned with the style chosen as long as it is consistent. If there are any major objections they should be raised soon!

.clang-format
```diff
@@ -0,0 +1,76 @@
...
+ObjCSpaceBeforeProtocolList: true
+
+PenaltyBreakBeforeFirstCallParameter: 16
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000
+PenaltyReturnTypeOnItsOwnLine: 60
+
+PointerAlignment: Middle
```
paulej: Would prefer the value "Right".

.clang-format
```diff
@@ -0,0 +1,76 @@
...
+BraceWrapping:
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterStruct:     false
+  AfterUnion:      false
+  BeforeCatch:     false
+  BeforeElse:      false
+  IndentBraces:    false
+BreakBeforeBinaryOperators: NonAssignment
```
paulej: Would prefer the value "None"

.clang-format
```diff
@@ -0,0 +1,76 @@
...
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 1000
+PenaltyReturnTypeOnItsOwnLine: 60
+
+PointerAlignment: Middle
+
+ReflowComments:  true
+
```
paulej: Need to add:
```
SortIncludes: false
```

to prevent headers from being included in the wrong dependency order.  Or, we need to fix the dependency order.

.clang-format
```diff
@@ -0,0 +1,76 @@
...
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+TabWidth:        4
+UseTab:          Never
```
paulej: Missing newline in the file, says github.

.clang-format
thisisG: Done!

.clang-format
thisisG: Fixed!

.clang-format
thisisG: Much prefer to not having to fix dependencies just to be in alphabetical order. :-) Will add!

thisisG:
Made the changes to the format file and applied it to the whole srtp.c file.
Currently going through and doing manual adjustments to comments - will push the whole file here when done for a second round of getting-looked-at before merging it in!

paulej:
If we have to make manual adjustments anytime we run the formatter, that seems a bit unfortunate.

thisisG:
The manual adjustments are to comments that have been clamped to 80 chars - some look a bit odd so just doing some touchups.
For a regular commit no manual adjustments should be required!

paulej:
Visually, the code changes seem to look pretty good.  I didn't check to see if anything was broken in the process.

You had asked before whether we should increase the line length from 80 to something else (132?).  I think 80 works pretty well here, but I did see a few lines that seemed wrapped were perhaps they would not with 132.  I have historically tried sticking with an 80-column format, but Python is teaching me that keeping it's formatting preferences and at 80 columns and while clear variable names is a bit challenging.  So, I've gotten used to longer lines these days.

What do you think?  Would be interesting to see if folks are sticking to the legacy terminal window size or accepting the fact that the GUI world has made that obsolete. :)

thisisG:
Personally I stick to 80. I find it helps me write clearer and more concise code as well as avoiding deep nesting. For Python I stick with PEP8 which is also 80 column width. :-)

The formatting suggestion was met by a surprising lack of opposition, so if it is OK I'd like to just hit merge now and I'll pick up on the rest of the code base after my holidays!

pabuhler:
Great start, please create a new PR with the rest of the project update as soon as possible!

PR #324, opened 2017-08-01
Format changes for rest of files in repo

thisisG:
This applies clang-format as defined in the .clang-format file to the rest of the files in the repo.


crypto/cipher/null_cipher.c
```diff
@@ -94,63 +94,57 @@ static srtp_err_status_t srtp_null_cipher_dealloc (srtp_cipher_t *c)
...
-    NULL             /* pointer to next testcase */
+    0,                  /* octets in key            */
+    NULL,               /* key                      */
+    0,                  /* packet index             */
+    0,                  /* octets in plaintext      */
+    NULL,               /* plaintext                */
+    0,                  /* octets in plaintext      */
+    NULL,               /* ciphertext               */
+    0,    NULL, 0, NULL /* pointer to next testcase */
 };
```
pabuhler: this is not a change I would want, it was structured for a reason, and now  that is lost ...

crypto/cipher/null_cipher.c
```diff
@@ -94,63 +94,57 @@ static srtp_err_status_t srtp_null_cipher_dealloc (srtp_cipher_t *c)
...
+const srtp_cipher_type_t srtp_null_cipher = {srtp_null_cipher_alloc,
+                                             srtp_null_cipher_dealloc,
+                                             srtp_null_cipher_init,
+                                             0, /* set_aad */
+                                             srtp_null_cipher_encrypt,
+                                             srtp_null_cipher_encrypt,
+                                             srtp_null_cipher_set_iv,
+                                             0, /* get_tag */
+                                             srtp_null_cipher_description,
+                                             &srtp_null_cipher_test_0,
```
pabuhler: this also makes no sense ...

crypto/hash/auth.c
```diff
@@ -138,10 +136,11 @@ srtp_auth_type_test (const srtp_auth_type_t *at, const srtp_auth_test_case_t *te
...
         debug_print(srtp_mod_auth, "data: %s",
-                    srtp_octet_string_hex_string(test_case->data,
-                                                 test_case->data_length_octets));
-        debug_print(srtp_mod_auth, "tag computed: %s",
-                    srtp_octet_string_hex_string(tag, test_case->tag_length_octets));
+                    srtp_octet_string_hex_string(
+                        test_case->data, test_case->data_length_octets));
+        debug_print(
+            srtp_mod_auth, "tag computed: %s",
+            srtp_octet_string_hex_string(tag, test_case->tag_length_octets));
```
pabuhler: This will be a strange, the 3 debug statements will end up formatted differently ..

crypto/hash/hmac.c
```diff
@@ -246,21 +256,14 @@ static const srtp_auth_test_case_t srtp_hmac_test_case_0 = {
...
-    srtp_hmac_start,
-    srtp_hmac_description,
-    &srtp_hmac_test_case_0,
-    SRTP_HMAC_SHA1
-};
-
+const srtp_auth_type_t srtp_hmac = {
+    srtp_hmac_alloc,       srtp_hmac_dealloc,      srtp_hmac_init,
+    srtp_hmac_compute,     srtp_hmac_update,       srtp_hmac_start,
+    srtp_hmac_description, &srtp_hmac_test_case_0, SRTP_HMAC_SHA1};
```
pabuhler: don't agree but hey a rule is a rule

crypto/hash/hmac_ossl.c
```diff
@@ -225,34 +233,28 @@ static const uint8_t srtp_hmac_test_case_0_tag[SHA1_DIGEST_SIZE] = {
...
-    srtp_hmac_start,
-    srtp_hmac_description,
-    &srtp_hmac_test_case_0,
-    SRTP_HMAC_SHA1
-};
-
+const srtp_auth_type_t srtp_hmac = {
+    srtp_hmac_alloc,       srtp_hmac_dealloc,      srtp_hmac_init,
+    srtp_hmac_compute,     srtp_hmac_update,       srtp_hmac_start,
+    srtp_hmac_description, &srtp_hmac_test_case_0, SRTP_HMAC_SHA1};
```
pabuhler: why does it need to change from the old initialization structure, no rules where broken and I think it made more sense.

crypto/hash/null_auth.c
```diff
@@ -133,28 +142,25 @@ static srtp_err_status_t srtp_null_auth_start (void *statev)
...
-
+const srtp_auth_type_t srtp_null_auth = {srtp_null_auth_alloc,
+                                         srtp_null_auth_dealloc,
+                                         srtp_null_auth_init,
+                                         srtp_null_auth_compute,
+                                         srtp_null_auth_update,
+                                         srtp_null_auth_start,
+                                         srtp_null_auth_description,
+                                         &srtp_null_auth_test_case_0,
+                                         SRTP_NULL_AUTH};
```
pabuhler: ??

crypto/cipher/null_cipher.c
ukreator: typical autoformatter-friendly way to comment fields is ```{/*comment1*/val1, /*comment2*/val2}```. Looks like all the places should be reviewed and fixed manually

thisisG: I did actually go through and fix _a lot_ of these manually since the comments were pretty much all over the place. Did miss out on this one though because it was just a tiny bit less horrible as compared to others.

Any place in the code where there are trailing aligned comments that look OK -> was fixed by hand. :-)

crypto/hash/auth.c
thisisG: That similarish things end up looking different - no way around that really.
I might have an idea of making them slightly more uniform - breaking the line for each and every argument to a function - but need to read up on my clang-format foo in order to do that.

crypto/hash/hmac_ossl.c
thisisG: It does not need to change. I will try to disable formatting for initialisation since that seems to be the largest issue.

crypto/hash/null_auth.c
thisisG: ¯\_(ツ)_/¯ heuristics keeps being heuristics. But yeah. Will try to disable formatting for initialisation.

thisisG:
Will open smaller PRs for separate files/collection of files.

PR #327, opened 2017-08-14
Add a section on config-time dependencies

### LINKED ISSUE #326: `configure != autoconf(configure.in)` on macOS
While preparing to attempt to make some changes to the build system for libsrtp, I tried to replicate the current autoconfig process before making any changes.  I immediately ran into a few problems:

1. The current `configure.in` file cannot be properly processed by autoconf
2. If you patch `configure.in` so that it works, you don't get the same `configure` file
3. The `configure` file you do get doesn't work

This is with autoconf 2.69, the same version that the current `configure` file claims to have been built with.

```
$ autoconf --version
autoconf (GNU Autoconf) 2.69
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+/Autoconf: GNU GPL version 3 or later
<http://gnu.org/licenses/gpl.html>, <http://gnu.org/licenses/exceptions.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by David J. MacKenzie and Akim Demaille.
```

When that version is run with the current `configure.in` file, it produces the following error:

```
$ autoconf
configure.in:17: error: possibly undefined macro: AM_PROG_AR
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
```

If you make the following change in `configure.in`...

```
-AM_PROG_AR
+m4_ifdef([AM_PROG_AR], [AM_PROG_AR])
```

... then autoconf runs without reporting any errors, but then `./configure` fails.

```
[[ trimmed ]]
checking for usleep... yes
checking for sigaction... yes
checking whether to enable debug logging in all modules... no
./configure: line 4806: PKG_PROG_PKG_CONFIG: command not found
checking whether to leverage OpenSSL crypto... no
./configure: line 4860: syntax error near unexpected token `crypto,'
./configure: line 4860: `     PKG_CHECK_MODULES(crypto, libcrypto >= 1.0.1,'
```

If the process for generating `configure` is something more advanced than running `autoconf`, then we should document that process.

bifurcation:
Fixes #326

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ because it does its work behind the scenes.
   - [Supported Features](#supported-features)
   - [Implementation Notes](#implementation-notes)
 - [Installing and Building libSRTP](#installing-and-building-libsrtp)
+  - [Changing Build Configuration](#changing-build-configuration)
 - [Applications](#applications)
   - [Example Code](#example-code)
 - [Credits](#credits)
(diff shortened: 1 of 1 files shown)


suhasHere:
Thanks for adding this clarification

PR #329, opened 2017-09-08
clang-format null_cipher.c

pabuhler:
My first attempt, would be good to get quick review to move on to next file

part of issue #254


.clang-format
```diff
@@ -21,6 +21,7 @@ AlwaysBreakTemplateDeclarations: false
 
 BinPackArguments: true
 BinPackParameters: false
+
```
thisisG: Added some whitespace?

PR #330, opened 2017-09-08
clang-format aes_gcm_ossl.c

pabuhler:
part of issue #254

### DIFF
--- a/crypto/cipher/aes_gcm_ossl.c
+++ b/crypto/cipher/aes_gcm_ossl.c
@@ -45,20 +45,19 @@
  */
 
 #ifdef HAVE_CONFIG_H
-    #include <config.h>
+#include <config.h>
 #endif
 
 #include <openssl/evp.h>
 #include "aes_icm_ossl.h"
 #include "aes_gcm_ossl.h"
 #include "alloc.h"
-#include "err.h"                /* for srtp_debug */
+#include "err.h" /* for srtp_debug */
 #include "crypto_types.h"
 
-
 srtp_debug_module_t srtp_mod_aes_gcm = {
-    0,               /* debugging is off by default */
-    "aes gcm"        /* printable module name       */
+    0,        /* debugging is off by default */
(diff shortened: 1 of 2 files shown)


crypto/cipher/aes_gcm_ossl.c
```diff
@@ -207,35 +208,40 @@ static srtp_err_status_t srtp_aes_gcm_openssl_context_init (void* cv, const uint
...
-
 /*
  * aes_gcm_openssl_set_iv(c, iv) sets the counter value to the exor of iv with
  * the offset
  */
-static srtp_err_status_t srtp_aes_gcm_openssl_set_iv (void *cv, uint8_t *iv, srtp_cipher_direction_t direction)
+static srtp_err_status_t srtp_aes_gcm_openssl_set_iv(
+    void *cv,
+    uint8_t *iv,
+    srtp_cipher_direction_t direction)
```
pabuhler: When parameters are to long to align with end of function they will be moved to start of line. Talked to @thisisG  about this and we agree it is an OK exception for now.

crypto/cipher/aes_gcm_ossl.c
```diff
@@ -439,7 +460,7 @@ static const srtp_cipher_test_case_t srtp_aes_gcm_test_case_0a = {
     srtp_aes_gcm_test_case_0_ciphertext, /* ciphertext  + tag        */
     20,                                  /* octets in AAD            */
     srtp_aes_gcm_test_case_0_aad,        /* AAD                      */
-    GCM_AUTH_TAG_LEN_8,
+    GCM_AUTH_TAG_LEN_8,                  /* */
```
paulej: I saw a few empty comment lines.  Did clang-format add those or did you?

pabuhler: I added them in order that clang format does not put multiple assignments on the same line.
Seamed like the lesser evil in order to keep some of the existing formatting.

PR #331, opened 2017-09-11
clang-format aes_icm_xxx.c

pabuhler:
part of issue #254

### DIFF
--- a/crypto/cipher/aes_icm.c
+++ b/crypto/cipher/aes_icm.c
@@ -44,7 +44,7 @@
  */
 
 #ifdef HAVE_CONFIG_H
-    #include <config.h>
+#include <config.h>
 #endif
 
 #define ALIGN_32 0
@@ -52,10 +52,9 @@
 #include "aes_icm.h"
 #include "alloc.h"
 
-
 srtp_debug_module_t srtp_mod_aes_icm = {
-    0,               /* debugging is off by default */
-    "aes icm"        /* printable module name       */
+    0,        /* debugging is off by default */
+    "aes icm" /* printable module name       */
 };
 extern const srtp_cipher_type_t srtp_aes_icm_128;
 extern const srtp_cipher_type_t srtp_aes_icm_256;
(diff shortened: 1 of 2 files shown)


crypto/cipher/aes_icm.c
```diff
@@ -198,21 +200,22 @@ static srtp_err_status_t srtp_aes_icm_context_init (void *cv, const uint8_t *key
...
-    debug_print(srtp_mod_aes_icm,
-                "offset: %s", v128_hex_string(&c->offset));
+    debug_print(srtp_mod_aes_icm, "key:  %s",
+                srtp_octet_string_hex_string(key, base_key_len));
+    debug_print(srtp_mod_aes_icm, "offset: %s", v128_hex_string(&c->offset));
 
     /* expand key */
-    status = srtp_aes_expand_encryption_key(key, base_key_len, &c->expanded_key);
+    status =
+        srtp_aes_expand_encryption_key(key, base_key_len, &c->expanded_key);
```
pabuhler: Not sure about this one, would have thought the assignment should stay on the same line but the function arguments break. @thisisG  what do you think?

thisisG: We can either play around with

PenaltyBreakAssignment (unsigned)
and
PenaltyBreakBeforeFirstCallParameter (unsigned)

or we can accept this as a quirk.

I am fine with either, let me know if we should have a stab at fixing it!

paulej: Personally, I like the way it turned out.

crypto/cipher/aes_icm.c
```diff
@@ -291,10 +293,11 @@ static void srtp_aes_icm_advance (srtp_aes_icm_ctx_t *c)
  *  - fill buffer then add in remaining (< 16) bytes of keystream
  */
 
-static srtp_err_status_t srtp_aes_icm_encrypt (void *cv,
-                                               unsigned char *buf, unsigned int *enc_len)
+static srtp_err_status_t srtp_aes_icm_encrypt(void *cv,
```
paulej: Notices an "e" out of place a few lines above (line 283 in the new code).  That's not from these changes, but ... wanna fix it?

pabuhler: done

crypto/cipher/aes_icm.c
pabuhler: It is not my personal preference but if it is not to common then it is ok. Will keep it like this for now and we can see what happens with the rest.

PR #334, opened 2017-09-18
clang-format crypto/replay/*

pabuhler:


crypto/replay/rdbx.c
```diff
@@ -259,16 +253,17 @@ unsigned long srtp_rdbx_get_window_size (const srtp_rdbx_t *rdbx)
...
         return srtp_err_status_ok;
     } else if ((int)(bitvector_get_length(&rdbx->bitmask) - 1) + delta < 0) {
         /* if delta is lower than the bitmask, it's bad */
         return srtp_err_status_replay_old;
-    } else if (bitvector_get_bit(&rdbx->bitmask,
-                                 (int)(bitvector_get_length(&rdbx->bitmask) - 1) + delta) == 1) {
+    } else if (bitvector_get_bit(
+                   &rdbx->bitmask,
+                   (int)(bitvector_get_length(&rdbx->bitmask) - 1) + delta) ==
+               1) {
```
paulej: We might be able to remove this lonely one here.  I think that function call returns either a 0 or 1, and I think the aim is to just check for a non-zero response.

pabuhler: Not sure if there is a better way of doing it, maybe if it was changed to if (1 == bitvector...... 
Would think it was possible to do it on 2 lines instead of 4 but since you approved I am going to leave it as is. Feel free to play with clang-format to get this better!

PR #335, opened 2017-09-20
clang-format crypto/hash/*

pabuhler:


pabuhler:
This did change quite a lot of preformatted code but I think the changes are ok.

crypto/hash/auth.c
```diff
@@ -177,14 +176,12 @@ srtp_auth_type_test (const srtp_auth_type_t *at, const srtp_auth_test_case_t *te
     return srtp_err_status_ok;
 }
 
-
 /*
  * auth_type_self_test(at) performs srtp_auth_type_test on at's internal
```
paulej: Historical, but since you're here: in the comment, we should have srtp_ prefixing the name of the function.

crypto/hash/null_auth.c
```diff
@@ -81,44 +86,51 @@ static srtp_err_status_t srtp_null_auth_alloc (srtp_auth_t **a, int key_len, int
...
 
-static srtp_err_status_t srtp_null_auth_dealloc (srtp_auth_t *a)
+static srtp_err_status_t srtp_null_auth_dealloc(srtp_auth_t *a)
 {
     extern const srtp_auth_type_t srtp_null_auth;
 
     /* zeroize entire state*/
-    octet_string_set_to_zero(a, sizeof(srtp_null_auth_ctx_t) + sizeof(srtp_auth_t));
+    octet_string_set_to_zero(a, sizeof(srtp_null_auth_ctx_t) +
+                                    sizeof(srtp_auth_t));
```
paulej: This line looks like unusual indentation.

pabuhler: I think it sees the two sizeof call as one operation, ie no "," and so when breaking an operation on to two lines it will indent.

PR #345, opened 2017-09-26
test/rtpw.c: Apply clang-format

thisisG:


test/rtpw.c
```diff
@@ -139,565 +133,568 @@ volatile int interrupted = 0;
...
+            exit(1);
+        }
+        if ((int)strlen(input_key) > policy.rtp.cipher_key_len * 2) {
+            fprintf(stderr, "error: too many digits in key/salt "
+                            "(should be %d hexadecimal digits, found %u)\n",
+                    policy.rtp.cipher_key_len * 2, (unsigned)strlen(input_key));
+            exit(1);
         }
+
+        printf("set master key/salt to %s/", octet_string_hex_string(key, 16));
```
paulej: I can see this is old code, but I wonder why there are two printf() statements to do the work one could have done?

pabuhler: Not sure there is a good answer to that. Feel free to clean up.

PR #352, opened 2017-09-27
clang-format crypto/test/*

pabuhler:


crypto/test/aes_calc.c
```diff
@@ -44,117 +44,113 @@
  */
 
 /*
-  
+
  Example usage (with first NIST FIPS 197 test case):
- 
-[sh]$ test/aes_calc 000102030405060708090a0b0c0d0e0f 00112233445566778899aabbccddeeff -v
+
+[sh]$ test/aes_calc 000102030405060708090a0b0c0d0e0f
```
paulej: Maybe insert a \\ character in the comments and indent the second line to make it more readable?

crypto/test/sha1_driver.c
```diff
@@ -56,505 +56,332 @@
...
+               "75c604ad7c0071c4fd78a7cb64786e1bece548fa4833c04065fe593f6fb10",
+          "f220a7457f4588d639dc21407c942e9843f8e26b" },
+        { 504, "836dfa2524d621cf07c3d2908835de859e549d35030433c796b81272fd8bc03"
+               "48e8ddbc7705a5ad1fdf2155b6bc48884ac0cd376925f069a37849c089c864"
+               "5",
+          "ddd2117b6e309c233ede85f962a0c2fc215e5c69" },
+        { 512, "7e3a4c325cb9c52b88387f93d01ae86d42098f5efa7f9457388b5e74b6d28b2"
+               "438d42d8b64703324d4aa25ab6aad153ae30cd2b2af4d5e5c00a8a2d0220c61"
+               "16",
+          "a3054427cdb13f164a610b348702724c808a0dcc" }
```
paulej: This struct is inconsistently formatted.  This line is perhaps the most inconsistent.  Not sure what we can do about it, but it makes for a great example of why the tool isn't always right.

pabuhler: Not sure if I agree in this case. When the tool breaks on "," then it aligns the break with the start of the assignment, in this case the last line is aligned with 512.  If it needs to break one expression then the break aligns with the start of the expression, in this case the long string is broken but stays aligned to indicate it is part of the same value.

At first I also though this strange but now I am beginning to accept it.  If you would like me to change this then please make a comment indicating what would be better. I will wait a day with merging :)

crypto/test/aes_calc.c
pabuhler: I agree, when the tool breaks comments it usually makes a mess.

crypto/test/sha1_driver.c
paulej: No need to fix it. It was only a remark on the tool that I understand is beyond our control.  I was just saddened to see different indentation styles in this block of code.

PR #353, opened 2017-09-27
clang-format crypto/math/*

pabuhler:


crypto/math/datatypes.c
```diff
@@ -53,48 +53,26 @@
...
-  2, 3, 3, 4, 3, 4, 4, 5,
-  3, 4, 4, 5, 4, 5, 5, 6,
-  3, 4, 4, 5, 4, 5, 5, 6,
-  4, 5, 5, 6, 5, 6, 6, 7,
-  3, 4, 4, 5, 4, 5, 5, 6,
-  4, 5, 5, 6, 5, 6, 6, 7,
-  4, 5, 5, 6, 5, 6, 6, 7,
-  5, 6, 6, 7, 6, 7, 7, 8
+int octet_weight[256] = {
+    0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4, 1, 2, 2, 3, 2, 3, 3, 4,
```
paulej: maybe we should turn off clang-format for this struct?

pabuhler: I did wounder out this but since I did not find any real reason for having 8 values in a row I let the tool do it's thing.
I looked more in to the code today and this is an interesting table containing the number bits set to "1" for each possible value of one byte, the Hamming weight.  I am not sure there is any real advantage in having 8 values per line, what do you think?   Interestingly it is not used in the library , only by tests.

paulej: The pattern seems to suggest a group of eight octets is significant.  I don't know, myself.

pabuhler: I don't see it, we can change it back at some point if needed.

PR #354, opened 2017-09-27
Ensure stream is freed if cloning fails

pabuhler:
srtp_stream_ctx_t is a complex struct, so once it has been
been allocated with srtp_crypto_alloc and at least partially
initialized it should always be freed with srtp_stream_free.
For this to work the struct needs to be initialized to 0.

Found in coverity.


pabuhler:
@paulej, @thisisG and I discussed this yesterday and came to the same conclusion but thought it should be a separate commit/issue.  Will look at creating a crypto_zalloc() and then start replace occurrences, look out for a PR coming soon.

PR #361, opened 2017-09-29
clang-format remain files

pabuhler:
This should be the last formatting change to existing code required to complete #254.
The provide script can be used to run clang-format on all source code in git repo, this can be used before creating a PR to check that new code is formatted correctly.
Will try to modify the travis build script to fail if running this script results in any changes to a PR.

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -5,7 +5,7 @@
 # run clang-format on each .c & .h file
 
 if [ -z "${CLANG_FORMAT}" ]; then
-    CLANG_FORMAT=clang-format-3.9
+    CLANG_FORMAT=clang-format
 fi
 
 a=`git ls-files | grep "\.h$\|\.c$"`
--- a/format.sh
+++ b/format.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+#
+# format.sh
+#
+# run clang-format on each .c & .h file
+
+if [ -z "${CLANG_FORMAT}" ]; then
+    CLANG_FORMAT=clang-format-3.9
+fi
+
+a=`git ls-files | grep "\.h$\|\.c$"`
+for x in $a; do
+     if [ $x != "config_in.h" ]; then
+         $CLANG_FORMAT -i -style=file $x
+     fi
+done
--- a/test/cutest.h
+++ b/test/cutest.h
@@ -10,10 +10,10 @@
  * the rights to use, copy, modify, merge, publish, distribute, sublicense,
  * and/or sell copies of the Software, and to permit persons to whom the
  * Software is furnished to do so, subject to the following conditions:
- * 
+ *
  * The above copyright notice and this permission notice shall be included in
  * all copies or substantial portions of the Software.
- * 
+ *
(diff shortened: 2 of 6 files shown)


format.sh
```diff
@@ -0,0 +1,16 @@
...
+#
+# format.sh
+#
+# run clang-format on each .c & .h file
+
+if [ -z "${CLANG_FORMAT}" ]; then
+    CLANG_FORMAT=clang-format-3.9
+fi
+
+a=`git ls-files | grep "\.h$\|\.c$"`
```
paulej: would `git ls-files '*.c' '*.h'` work just as well?

test/cutest.h
```diff
@@ -78,9 +76,9 @@
  *       TEST_CHECK(ptr->member2 > 200);
  *   }
  */
-#define TEST_CHECK_(cond,...)  test_check__((cond), __FILE__, __LINE__, __VA_ARGS__)
-#define TEST_CHECK(cond)       test_check__((cond), __FILE__, __LINE__, "%s", #cond)
-
+#define TEST_CHECK_(cond, ...)                                                 \
```
paulej: This line appears to have a ton of white space and then a \\ character now.

thisisG: This is expected.
```
#define TEST_CHECK_(cond,...)  test_check__((cond), __FILE__, __LINE__, __VA_ARGS__)
```
is 85 characters. clang-format wraps it since we have a column limit of 80. For macros the wrapping is done at column 80 for consistency across all macros (instead of some macros wrapping at column 25, 34, 42 etc. which would arguably look a bit messy).

format.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+#
+# format.sh
+#
+# run clang-format on each .c & .h file
+
+if [ -z "${CLANG_FORMAT}" ]; then
+    CLANG_FORMAT=clang-format-3.9
```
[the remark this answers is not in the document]
thisisG: IMO it'd be better to assume ```clang-format``` exists. Most systems I've seen does not specify the version in the executable.

test/cutest.h
paulej: Where it wraps makes sense.  It's all the space leading up to that \\ character that was a surprise.  The tool doesn't aim for consistency in other things, like enum assignments, etc.

format.sh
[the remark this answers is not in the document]
pabuhler: I was lazy and used the version I have ... but can be neutral.

test/cutest.h
pabuhler: the tool is not perfect

format.sh
pabuhler: no but `git ls-files *.c *.h` seams to will change it later. Plan to make the script report that code was changed so can be used as part of a travis script.

paulej: Maybe the single quote mark needs to be escaped.  My concern with having a raw *.c there is that the shell might expand that to be whatever .c file you have in the current directory.  (We don't have any in the root directory, but what if one is there?

PR #363, opened 2017-10-12
change srtp_crypto_alloc to initialize memory to zero

pabuhler:
Proactively prevent accessing uninitialized memory.
Majority of calls to srtp_crypto_alloc had a corresponding
call to memset.


paulej:
Looks good.

I noticed some existing code will immediately call free if the returned pointer is NULL.  Think we should change that?

pabuhler:
@paulej  yes, we should remove calls that do nothing and make no sense.

PR #365, opened 2017-10-18
Replace calls to free() with srtp_crypto_free()

paulej:
I noticed a couple of places where free() was called when it should be calls to srtp_crypto_free().

### DIFF
--- a/crypto/test/cipher_driver.c
+++ b/crypto/test/cipher_driver.c
@@ -456,8 +456,8 @@ srtp_err_status_t cipher_array_alloc_init(srtp_cipher_t ***ca,
     int klen_pad = ((klen + 15) >> 4) << 4;
 
     /* allocate array of pointers to ciphers */
-    cipher_array =
-        (srtp_cipher_t **)malloc(sizeof(srtp_cipher_t *) * num_ciphers);
+    cipher_array = (srtp_cipher_t **)srtp_crypto_alloc(sizeof(srtp_cipher_t *) *
+                                                       num_ciphers);
     if (cipher_array == NULL)
         return srtp_err_status_alloc_fail;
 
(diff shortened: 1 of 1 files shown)


crypto/test/cipher_driver.c
```diff
@@ -467,7 +467,7 @@ srtp_err_status_t cipher_array_alloc_init(srtp_cipher_t ***ca,
     /* allocate key */
     key = srtp_crypto_alloc(klen_pad);
     if (key == NULL) {
-        free(cipher_array);
+        srtp_crypto_free(cipher_array);
```
pabuhler: cipher_array is allocated with malloc not srtp_crypto_alloc, see 
` cipher_array =  (srtp_cipher_t **)malloc(sizeof(srtp_cipher_t *) * num_ciphers);`
so either that also needs to change or else go back to using free.

crypto/test/cipher_driver.c
```diff
@@ -509,7 +509,7 @@ srtp_err_status_t cipher_array_delete(srtp_cipher_t *cipher_array[],
         srtp_cipher_dealloc(cipher_array[i]);
     }
 
-    free(cipher_array);
+    srtp_crypto_free(cipher_array);
```
pabuhler: same as previous comment

crypto/test/cipher_driver.c
```diff
@@ -562,7 +562,7 @@ uint64_t cipher_array_bits_per_second(srtp_cipher_t *cipher_array[],
     }
     timer = clock() - timer;
 
-    free(enc_buf);
+    srtp_crypto_free(enc_buf);
```
pabuhler: this one looks correct

paulej:
I made changed and pushed a new set of changes.  There was one malloc() in the code and I changed it to crypto_alloc(), since there were a couple of places using that.  Given it is test code, it's not so important, but I do like consistency.

One thing I noted when I ran the format script is that it re-formatted a couple of other files in minor ways.  I did not commit those other files, though, as I did not want to expand the scope of this commit.  But, what this tells me is that if I were to change srtp.c, for example, I might run into trouble if Travis enforces proper formatting.  I have clang-format version 4.0.1 installed.

pabuhler:
@paulej  yes I saw there where differences and the default install on trasvis is 3.9 . Maybe can get it to upgrade to 4.0, I'll have a check, either way making the format scrip call an explicit version will help stop confusion and can then be considered the official version for this project.

Issue #328, opened 2017-08-29
libsrtp has automatic dependency on libcap


joakim-tjernlund:
One should be able to control linkage with libpcap, this patch is a start:
```
--- a/configure.in	2015-03-25 15:37:31.953386365 +0000
+++ b/configure.in	2015-03-25 15:39:49.192183689 +0000
@@ -182,6 +182,8 @@
 AC_SUBST(HMAC_OBJS)
 
 dnl Checking for PCAP
+AC_ARG_ENABLE([pcap], AS_HELP_STRING([--disable-pcap], [Build without `pcap' library (-lpcap)]))
+AS_IF([test "x$enable_pcap" != "xno"], [
 AC_CHECK_LIB([pcap],[pcap_create],
     [PCAP=1
      LIBS="-lpcap $LIBS"
@@ -189,6 +191,7 @@
      AC_DEFINE(HAVE_PCAP, 1, [Define to 1 if you have the `pcap' library (-lpcap)])
      AC_SUBST(HAVE_PCAP)
 ])
+])
 
 AC_MSG_CHECKING(whether to use syslog for error reporting)
 AC_ARG_ENABLE(syslog,

pabuhler:
Hi @joakim-tjernlund , I am not sure exactly what the issues is. I have made a small patch that will only add -lpcap to the one executable that requires it instead of to all executable, this will still be done automatically if libpcap is detected.  Does this help for you?
If not can you please explain why you need to explicitly not link against libpcap? I would prefer not adding a configure option for this.

joakim-tjernlund:
Hi @pabuhler 

Your patch is a step in the right direction, only link pcap with apps that need it. Now one
can see that pcap is only needed for test app(s).
But this will still build against pcap if it is installed, I should/need to be able to control explicitly
whether pcap should be used or not.
If --enable-pcap is active one should get a configure error that pcap is missing and
with --disable-pcap libsrtp should not build against pcap even if pcap is installed.

joakim-tjernlund:
If you really want to keep the current auto dependency on pcap you could have
enable, disable and auto(default)

Also, can build of test progs be configurable?

pabuhler:
I don't wish to be difficult but I do not see the benefit of adding more configure options, they will just require maintenance. If you have a real reason to add them then please explain and we can consider it. ( even better create a PR) .

The make target `make shared_library` will build just the shared library and no test apps. In general though I would recommend building the tests and running them.

[...]

PR #366, opened 2017-10-20
Only link against pcap when needed

pabuhler:
This is a potential fix for #328.

do not add -lpcap to LIB's, instead added
it to the line of onlt executable that requires
it.


pabuhler:
Closing this pr as it is now in conflict and does not offer anything new.

PR #368, opened 2017-10-30
Rename configure.in to configure.ac

floppym:
The name "configure.ac" has been preferred for many years.


pabuhler:
I don't think we will do this in the throttle branches, master only

PR #372, opened 2017-11-01
make rtp_decoder compile on MinGW

MoeRT09:
For several reasons the compilation of rtp_decoder fails when compiling on Windows using MinGW. This pull request fixes this.


pabuhler:
Hi @MoeRT09, thanks for the patch, configure.in was just renamed to configure.ac so unfortunately you will need to reapply.  There is also an out standing issue/pr relating to pcap support, #328 & #366, that you might need to consider, I would need to look a little closer at your patch to see how it would integrate there.

MoeRT09:
@pabuhler I tried to address your suggestions. The correct Lib(w)pcap now gets only linked to rtp_decoder.

pabuhler:
looks like clang-format did not like your commit
https://travis-ci.org/cisco/libsrtp/jobs/296342642

could you fix it up?

thanks

MoeRT09:
done 🙂

Issue #251, opened 2017-02-16
Register libsrtp with Coverity Scan


JonathanLennox:
Coverity Scan is free source code analysis for open-source tools https://scan.coverity.com/ -- it seems like it'd be a good idea to register LibSRTP with it.

pabuhler:
Seams like a good idea, although a bit of work to set up. Will try to follow up and get this run before the release but no promises.

thisisG:
After discussing this we're mocing this to 2.2 milestone instead of 2.1.

pabuhler:
This is now in and the results are at https://scan.coverity.com/projects/cisco-libsrtp

PR #376, opened 2017-11-11
Add coverity scan to tavis builds

pabuhler:
Current will run the scan on the master branch with
each build in master. If the frequency of commits
increases then can consider a coverity_scan branch.

Results can be found at:
https://scan.coverity.com/projects/cisco-libsrtp

This should resolve issue #251.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ dist: trusty
 sudo: false
 language: c
 
+env:
+  global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,6 +2,10 @@ dist: trusty
 sudo: false
 language: c
 
+env:
+  global:
+    - secure: "YJSSinoWNfMicNYCdxsnlgZc9AIi1yKkgJTmKk41zRI23r8c5Z339Zl1p/boxOxD+YLCQshBYR00ckJXT9R8wQy0WhpWdA3YJYAMWjnVkbBMBNf3VEQSIvHwV8kLDBMPBikhks7b4ACPG3UYLsQrxUJFfJz4Grn7z8tbUedv1Gk="
```
paulej: It's not secure now. Does this need to be there?

pabuhler: Should be secure, can you read it ?

paulej: Just being funny.

PR #378, opened 2017-11-12
make read only arrays const

nils-ohlmeier:


crypto/math/datatypes.c
```diff
@@ -53,7 +53,7 @@
 
 #include "datatypes.h"
 
-int octet_weight[256] = {
+static const int8_t octet_weight[256] = {
     0, 1, 1, 2, 1, 2, 2, 3, 1, 2, 2, 3, 2, 3, 3, 4, 1, 2, 2, 3, 2, 3, 3, 4,
```
pabuhler: static const  ok, but int8_t is that just to save space ?

nils-ohlmeier: The original bug report on the Firefox side is https://bugzilla.mozilla.org/show_bug.cgi?id=1255655

But yes since none of the data in the array is bigger then int8 this as is currently wasting memory is the justification for shrinking this.

pabuhler: ok, was just checking.

PR #379, opened 2017-11-12
fixed several compiler warnings from Firefox builds

nils-ohlmeier:


pabuhler:
I can see why once an int is changed ti unsigned int that the changes will need to propagate, but why was the original change needed? What compiler/flags is firefox using that cause this to be an issue?

nils-ohlmeier:
Here is the try run from Firefox CI system without any of the fixes applied: https://treeherder.mozilla.org/#/jobs?repo=try&revision=75cfcae77efe95781b61ecfd73ca5ef263a4341f

Note: in our current code most of the places got fixed with casting. 
But I'm not a big fan of casting as it only tells the compiler to not complain about the problem any more. Since I'm not expert on the code base I thought it's safer to not use casting if I can avoid it.

But I would be happy to change them all into casts if an expert on the code assures me they are all safe to cast.

nils-ohlmeier:
Here is an example of the compiler line used when compiling SRTP as part of my Firefox locally on my Mac:

/usr/bin/clang -std=gnu99 -o Unified_c_netwerk_srtp_src0.o -c -fvisibility=hidden -fvisibility-inlines-hidden -DDEBUG=1 -DOS_POSIX=1 -DOS_MACOSX=1 -DHAVE_STDLIB_H=1 -DHAVE_UINT8_T=1 -DHAVE_UINT16_T=1 -DHAVE_UINT32_T=1 -DHAVE_UINT64_T=1 -DCPU_CISC=1 -DHAVE_X86 -DSTATIC_EXPORTABLE_JS_API -DMOZ_HAS_MOZGLUE -DMOZILLA_INTERNAL_API -DIMPL_LIBXUL -I/Users/nohlmeier/src/mozilla-central/netwerk/srtp/src -I/Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/netwerk/srtp/src -I/Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/ipc/ipdl/_ipdlheaders -I/Users/nohlmeier/src/mozilla-central/ipc/chromium/src -I/Users/nohlmeier/src/mozilla-central/ipc/glue -I/Users/nohlmeier/src/mozilla-central/netwerk/srtp/src/crypto/include -I/Users/nohlmeier/src/mozilla-central/netwerk/srtp/src/include -I/Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/dist/include -I/Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/dist/include/nspr -I/Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/dist/include/nss -fPIC -include /Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/mozilla-config.h -DMOZILLA_CLIENT -Qunused-arguments -Qunused-arguments -Wall -Wempty-body -Wignored-qualifiers -Wpointer-arith -Wsign-compare -Wtype-limits -Wunreachable-code -Wunreachable-code-return -Wclass-varargs -Wloop-analysis -Wstring-conversion -Wno-error=deprecated-declarations -Wno-error=array-bounds -Wformat -Wno-gnu-zero-variadic-macro-arguments -Wformat-security -fno-strict-aliasing -fno-math-errno -pthread -pipe -g -fno-omit-frame-pointer -std=gnu99  -MD -MP -MF .deps/Unified_c_netwerk_srtp_src0.o.pp  -fcolor-diagnostics  /Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/netwerk/srtp/src/Unified_c_netwerk_srtp_src0.c

If you check out the results from the try run in my previous comment you will see that the different compilers across different OS's complain about different things. This patch addresses all of them at once.

crypto/replay/rdb.c
```diff
@@ -100,7 +100,7 @@ srtp_err_status_t srtp_rdb_check(const srtp_rdb_t *rdb, uint32_t p_index)
  */
 srtp_err_status_t srtp_rdb_add_index(srtp_rdb_t *rdb, uint32_t p_index)
 {
-    int delta;
+    unsigned int delta;
```
pabuhler: This worried me a bit but the nice comment on the line below  ` /* here we *assume* that p_index > rdb->window_start */`  made it all good again. Maybe we should add an actual check instead of just a comment.

crypto/include/cipher.h
```diff
@@ -132,16 +132,16 @@ typedef srtp_err_status_t (*srtp_cipher_get_tag_func_t)(void *state,
...
+    const uint8_t *key;                    /* key                      */
+    uint8_t *idx;                          /* packet index             */
+    unsigned int plaintext_length_octets;  /* octets in plaintext      */
+    const uint8_t *plaintext;              /* plaintext                */
+    unsigned int ciphertext_length_octets; /* octets in plaintext      */
+    const uint8_t *ciphertext;             /* ciphertext               */
+    int aad_length_octets;                 /* octets in AAD            */
+    const uint8_t *aad;                    /* AAD                      */
+    int tag_length_octets;                 /* Length of AEAD tag       */
     const struct srtp_cipher_test_case_t
```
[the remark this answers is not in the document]
nils-ohlmeier: Agreed. I was trying to keep the changes as small as possible. But I would be happy to make this consistent in a follow up PR.

crypto/replay/rdb.c
nils-ohlmeier: Good point. Happy to add that error check. Not sure about which error to return though. srtp_err_status_replay_fail?

srtp/srtp.c
```diff
@@ -4156,12 +4156,16 @@ srtp_err_status_t srtp_unprotect_rtcp_mki(srtp_t ctx,
 
     /* we assume the hdr is 32-bit aligned to start */
 
+    if (pkt_octet_len < 0)
```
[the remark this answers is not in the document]
nils-ohlmeier: Yes. Fixing it.

crypto/replay/rdb.c
pabuhler: Unfortunately the return value is not currently checked so srtp_err_status_replay_fail should be fine.

nils-ohlmeier:
@pabuhler do you think we can merge this? Or do you want someone else to review?

I'm asking because I would prefer to merge from upstream with our patches included, rather then continuing to maintain them separately.

pabuhler:
I am ok with merging, was wait to see if anyone else had an opinion, but they are far and few between on this project.

PR #382, opened 2017-11-12
patch from libjingle

nils-ohlmeier:
This patch has been used in the Firefox code base for a long time.
It got originally landed in as part of a bigger patch referencing upstream libjingle as the source here https://bugzilla.mozilla.org/page.cgi?id=splinter.html&bug=792325&attachment=662542

I can't really explain or justify why it's good, but I did not wanted to withhold this in case it fixes a potential issue.
In other words: a real expert should review this :-)


nils-ohlmeier:
Looks like this is actually breaking the ROC.

PR #386, opened 2017-11-17
fixed non-const redeclaraion error

nils-ohlmeier:
Without this fix libsrtp fails to compile as part of Firefox with errors like this:

 0:33.90 In file included from /Users/nohlmeier/src/mozilla-central/objdir-x86_64-apple/netwerk/srtp/src/Unified_c_netwerk_srtp_src0.c:83:
 0:33.90 /Users/nohlmeier/src/mozilla-central/netwerk/srtp/src/crypto/kernel/crypto_kernel.c:73:27: error: redeclaration of 'srtp_null_cipher' with a different type: 'srtp_cipher_type_t' (aka 'struct srtp_cipher_type_t') vs 'const srtp_cipher_type_t' (aka 'const struct srtp_cipher_type_t')
 0:33.90 extern srtp_cipher_type_t srtp_null_cipher;
 0:33.90                           ^
 0:33.90 /Users/nohlmeier/src/mozilla-central/netwerk/srtp/src/crypto/cipher/null_cipher.c:144:26: note: previous definition is here
 0:33.90 const srtp_cipher_type_t srtp_null_cipher = {
 0:33.90                          ^

### DIFF
--- a/crypto/kernel/crypto_kernel.c
+++ b/crypto/kernel/crypto_kernel.c
@@ -70,9 +70,9 @@ extern srtp_debug_module_t mod_alloc;
  * cipher types that can be included in the kernel
  */
 
-extern srtp_cipher_type_t srtp_null_cipher;
-extern srtp_cipher_type_t srtp_aes_icm_128;
-extern srtp_cipher_type_t srtp_aes_icm_256;
+extern const srtp_cipher_type_t srtp_null_cipher;
+extern const srtp_cipher_type_t srtp_aes_icm_128;
+extern const srtp_cipher_type_t srtp_aes_icm_256;
 #ifdef OPENSSL
 extern srtp_cipher_type_t srtp_aes_icm_192;
 extern srtp_cipher_type_t srtp_aes_gcm_128_openssl;
@@ -89,8 +89,8 @@ extern srtp_debug_module_t srtp_mod_aes_gcm;
  * auth func types that can be included in the kernel
  */
 
-extern srtp_auth_type_t srtp_null_auth;
-extern srtp_auth_type_t srtp_hmac;
+extern const srtp_auth_type_t srtp_null_auth;
+extern const srtp_auth_type_t srtp_hmac;
 
 /* debug modules for auth types */
 extern srtp_debug_module_t srtp_mod_hmac;


pabuhler:
I agree they should be consistent. Is there a specific compiler flag that is catching this? Would be good to included it.

What about srtp_aes_gcm_128_openssl, srtp_aes_gcm_256_openssl & srtp_aes_icm_192 ? Are you not building with openssl support?

There are a few other places in the code (tests) where there is the same problem, would be great to be consistent so have being able to generate the same errors on travis would be good.

JonathanLennox:
It looks like Mozilla is compiling all the libsrtp code as a single compilation unit (by including all the C files in "Unified_c_netwerk_srtp_src0.c"). This is how the compiler is catching the inconsistent declarations of srtp_null_cipher in crypto_kernel.c and null_cipher.c.

In the normal build these separate compilation units and so the compiler doesn't notice the inconsistency.

To catch these normally, we'd need to have a header file with the declarations, included both by the definition of the cipher and crypto_kernel.c, rather than crypto_kernel.c having its own inline declarations.  This is probably a good idea.

nils-ohlmeier:
> What about srtp_aes_gcm_128_openssl, srtp_aes_gcm_256_openssl & srtp_aes_icm_192 ? Are you not building with openssl support?

Firefox gets build without OpenSSL, so that's why we hadn't fixed these.

And @JonathanLennox is right, that the Firefox builds combines sources files where it can.

Should I start looking into moving all these declarations into headers to fix this?

pabuhler:
Moving them to the bottom of chiper.h looks like the best alternative without introducing a new header file ?

JonathanLennox:
cipher.h is public, whereas I don't think we want these structures to be exported.

pabuhler:
Right, I was just looking at the source tree and had forgotten the installed files.

nils-ohlmeier:
Is everything in crypto/include public? Or can I add a new header in there?

JonathanLennox:
In crypto/include, only cipher.h, auth.h, and crypto_types.h are public (see the install rule in Makefile.in.)  Adding a new private header there is fine.

If you add a new private header file, please also add it to the Visual Studio project files (srtp2.vcxproj and srtp2.vcxproj.filters) or ping me to send you a patch to do it.

nils-ohlmeier:
Did a first run in #388

pabuhler:
Closing this as  #388 should have addressed the issue. Please reopen if this is not the case.

PR #388, opened 2017-11-17
moved externs from crypto_kernel into its own header

nils-ohlmeier:


nils-ohlmeier:
Not sure if the `extern srtp_debug_module_t` in crypto_kernel.c should be moved out as well, or not.

pabuhler:
I guess extern is no longer needed?

Personally I think the loading of a cihper should have triggered the loading of the ciphers debug module  but that is a bit of a larger change.

Moving the cipher debug modules to be declared in the same place as the corresponding ciphers is probably a good idea. Less extern's the better.

JonathanLennox:
The Visual Studio project files list internal headers, so this new file should be listed there.

nils-ohlmeier:
@JonathanLennox thanks for catching that. I forgot about that.
I edited the files now in vi to add the new header. But as I don't have Visual Studio I can't verify if they work.

JonathanLennox:
It looks like vi added newlines to the end of the file (Visual Studio usually likes to leave its project files without a trailing newline), but that's harmless; looks good.

nils-ohlmeier:
Is this ready to be merged? This would close issue #386 and would allow me to finally update Firefox :-)

pabuhler:
Yip, once again giving anyone else a chance to have an opinion ..

PR #390, opened 2017-11-28
bitvector of length zero is not valid

pabuhler:
Fixes coverity scan issues 179781

If the length resulted in l == 0 then there was no
allocation of word, which could subsequently be
dereferenced. Treat length of zero as invalid to
avoid dealing with it.


pabuhler:
I agree that we should look into 0 length allocations.

PR #395, opened 2018-01-24
Memory access fixes

pabuhler:
Reported by @guidovranken


pabuhler:
Closing this PR until I figure why format check complains so much as this should not cause such a problem

PR #398, opened 2018-01-26
Memory access fixes

pabuhler:
This contains a group of fixes to various memory access violations detected using a fuzzer provided by @guidovranken.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -257,12 +257,6 @@ srtp_err_status_t srtp_stream_dealloc(srtp_stream_ctx_t *stream,
                 srtp_crypto_free(session_keys->limit);
             }
         }
-    }
-
-    if (stream_template &&
-        stream->session_keys == stream_template->session_keys) {
-        /* do nothing */
-    } else {
         srtp_crypto_free(stream->session_keys);
     }
 
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -4540,42 +4540,85 @@ unsigned int srtp_profile_get_master_salt_length(srtp_profile_t profile)
     }
 }
 
+srtp_err_status_t stream_get_protect_trailer_length(srtp_stream_ctx_t *stream,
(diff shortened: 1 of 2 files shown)


srtp/srtp.c
```diff
@@ -147,53 +147,141 @@ unsigned int srtp_get_version()
...
+        srtp_crypto_free(stream->session_keys);
     }
 
-    if (str->enc_xtn_hdr) {
-        srtp_crypto_free(str->enc_xtn_hdr);
+    status = srtp_rdbx_dealloc(&stream->rtp_rdbx);
+    if (status)
+        return status;
+
+    /* DAM - need to deallocate EKT here */
```
thisisG: DAM?

thisisG: Might as well implement the EKT deallocation now?

thisisG: Talked to @pabuhler about this. EKT is not really up to scratch so work should be done to either disable it (or if someone is up to it) fix it - but that is a different chain of changes.

srtp/srtp.c
```diff
@@ -147,51 +147,136 @@ unsigned int srtp_get_version()
...
+            session_keys = &stream->session_keys[i];
 
-        if (session_keys->rtp_xtn_hdr_cipher) {
-            srtp_cipher_dealloc(session_keys->rtp_xtn_hdr_cipher);
-        }
+            if (stream_template &&
+                stream->num_master_keys == stream_template->num_master_keys) {
+                template_session_keys = &stream_template->session_keys[i];
+            } else {
+                template_session_keys = NULL;
```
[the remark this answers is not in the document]
pabuhler: You are right it does not look like this is needed, either it will always be set to something or never set. But it will not change the behavior.

srtp/srtp.c
pabuhler: And to be clear this is not a new change, this code was mostly moved from further down the file.

PR #399, opened 2018-01-26
Fix memory leak (free RDBX)

guidovranken:


pabuhler:
Looking at the code I can see how this happens and it appears to be safe to call it, but I think the real problem is that srtp_stream_dealloc should have been called instead of srtp_stream_free.  It is a bit confusing when reading the code and the relationship of the function srtp_stream_alloc, srtp_stream_init, srtp_stream_dealloc and srtp_stream_free is confusing. I think renaming  srtp_stream_dealloc to srtp_stream_deinit would help. I will do some more follow up based on you mail.

pabuhler:
@guidovranken  I think commit https://github.com/cisco/libsrtp/pull/398/commits/8ec13a7eb1fa099daef61745533f4661277c1e16 which is part of #398 will make this commit obsolete as  srtp_stream_free is removed and srtp_stream_dealloc is used instead.

pabuhler:
I am closing this as after #398 I no longer can reproduce this issue. Please reopen if the issue persists for you.

Issue #371, opened 2017-11-01
srtp_init failed in FIPS mode with CiscoFOM 6.2


weilhuan:
Combine libSRTP and CiscoSSL6.2.83(CiscoFOM 6.2), when calling srtp_init() during startup procedure, it's always failed if FIPS mode is enabled.

pabuhler:
hi @weilhuan if you apply the following patch does it still fail?

```
diff --git a/crypto/cipher/aes_gcm_ossl.c b/crypto/cipher/aes_gcm_ossl.c
index 75519b7..b6aba21 100644
--- a/crypto/cipher/aes_gcm_ossl.c
+++ b/crypto/cipher/aes_gcm_ossl.c
@@ -241,7 +241,7 @@ static srtp_err_status_t srtp_aes_gcm_openssl_set_iv(
                              (void *)iv)) {
         return (srtp_err_status_init_fail);
     }
-    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_IV_GEN, 0, (void *)iv)) {
+    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_IV_GEN, 12, (void *)iv)) {
         return (srtp_err_status_init_fail);
     }
```

weilhuan:
Hi,

I cannot verify the change now since I’m not in office. Besides this change, according to the migration guide of FOM6.2, perhaps we need add *NON_FIPS_ALLOW flag for legacy use such as SHA-1, HMAC.

BTW, can I request fully test with FOM6.2 which has been released quite a while.

Thanks,
Weiliqi



On 1 Nov 2017, at 6:20 PM, Pascal Bühler <notifications@github.com<mailto:notifications@github.com>> wrote:


hi @weilhuan<https://github.com/weilhuan> if you apply the following patch does it still fail?

diff --git a/crypto/cipher/aes_gcm_ossl.c b/crypto/cipher/aes_gcm_ossl.c
index 75519b7..b6aba21 100644
--- a/crypto/cipher/aes_gcm_ossl.c
+++ b/crypto/cipher/aes_gcm_ossl.c
@@ -241,7 +241,7 @@ static srtp_err_status_t srtp_aes_gcm_openssl_set_iv(
                              (void *)iv)) {
         return (srtp_err_status_init_fail);
     }
[...]

PR #403, opened 2018-02-26
Set gcm IV directly with EVP_CipherInit_ex

pabuhler:
Setting the IV directly with EVP_CipherInit_ex will
save two calls to openssl and simplify the code.

There is a comment that the 3 calls are required but
I am not sure why, EVP_CTRL_GCM_SET_IV_FIXED will
just store the vector and EVP_CTRL_GCM_IV_GEN will use
it internally and then increment the last digits before
returning it in iv variable passed in.
EVP_CipherInit_ex will store the iv and use it internally.
Incrementing and retrieving the new IV is not required for
SRTP-GCM, a new IV is used for each packet.

Tested with openssl 1.0.1d & 1.0.2g

This will also address the issue with #371

### DIFF
--- a/crypto/cipher/aes_gcm_ossl.c
+++ b/crypto/cipher/aes_gcm_ossl.c
@@ -220,20 +220,12 @@ static srtp_err_status_t srtp_aes_gcm_openssl_set_iv(
     debug_print(srtp_mod_aes_gcm, "setting iv: %s",
                 v128_hex_string((v128_t *)iv));
 
-    if (!EVP_CipherInit_ex(c->ctx, NULL, NULL, NULL, NULL,
-                           (c->dir == srtp_direction_encrypt ? 1 : 0))) {
-        return (srtp_err_status_init_fail);
-    }
-
-    /* set IV len  and the IV value, the followiong 3 calls are required */
     if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_IVLEN, 12, 0)) {
         return (srtp_err_status_init_fail);
     }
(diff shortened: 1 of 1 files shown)


pabuhler:
@jfigus as original author are you able to comment on this?

jfigus:
I'm surprised this change works, as the IV length is no longer specified.  Unfortunately the OpenSSL API for AEAD ciphers is quite horrendous.  The safe play here is to not change the code.

pabuhler:
@jfigus thanks for the quick reply, EVP_CTRL_GCM_SET_IVLEN is still called before EVP_CipherInit_ex so the length is set. This follows the example code from https://wiki.openssl.org/index.php/EVP_Authenticated_Encryption_and_Decryption . 

The current code using EVP_CTRL_GCM_IV_GEN will overwrite the iv passed in to srtp_aes_gcm_openssl_set_iv, which seams strange to me.

As #371 indicates the current code no longer works with the cisco openssl fork, due to to requirements on the input values, and after looking in to that issue it seams that EVP_CTRL_GCM_IV_GEN is not required at all.

pabuhler:
@jfigus yes for SRTP the IV is set for every packet so the incrementing of the counter in EVP_CTRL_GCM_IV_GEN seams redundant. I will double check again but so far testing between code running with this change and the original code does not indicate any errors.
I understand it is hard for you to make a comment especially relating to Cisco's Openssl.
Mostly I was interested if you remember the reason for doing it the the way did  originally.

jfigus:
@pabuhler It's been quite a while, but I think this code was modeled after the OpenSSL utility in their apps/enc.c source.  Their API was finicky to get working with AEAD at the time, so I just followed what they were doing.  But maybe things have changed since then. If the new code is working for you now, you're probably good to go.

pabuhler:
@jfigus thanks, thats as much as I can hope for.

PR #406, opened 2018-03-07
Fix unprotect when pktlen < (2*mki_size + tag_len)

vopatek:
The condition mki_start_location >= *mki_size in
srtp_get_session_keys() should use base_mki_start_location.
Now the condition is false for packets < 2*mki_size + tag_len.
But as of commit d4bd43c the correct condition is now checked
earlier so we simply remove the expression altogether.


pabuhler:
Hi @vopatek , looks good but could you fix up the code format failure, see https://travis-ci.org/cisco/libsrtp/jobs/350356630 .

PR #409, opened 2018-04-09
Compatibilty with LibreSSL

strfry:
LibreSSL is a fork of OpenSSL 1.0 that is used by some distributions like Alpine Linux.

They basically keep the 1.0.1 API, but advertize a higher version number.

This patch deals with this issue by also checking for a LibreSSL-specific macro.

### DIFF
--- a/crypto/hash/hmac_ossl.c
+++ b/crypto/hash/hmac_ossl.c
@@ -79,7 +79,7 @@ static srtp_err_status_t srtp_hmac_alloc(srtp_auth_t **a,
 
 /* OpenSSL 1.1.0 made HMAC_CTX an opaque structure, which must be allocated
    using HMAC_CTX_new.  But this function doesn't exist in OpenSSL 1.0.x. */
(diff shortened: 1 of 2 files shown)


pabuhler:
Hi @strfry , did you mean to add the meson.build file?  If you did then please create a separate pull request for that. Adding a new build system requires someone committing to maintain that over time and so it would be good to have a separate PR for that.

strfry:
Sorry @pabuhler , that was an accident (or rather a misunderstanding of pull request semantics)
, not an attempt to sneak in a new build system ;)
I moved the offending commit into another branch, so i hope you can review this PR now :)

strfry:
Having automated tests for this sounds like a good idea.
But Travis only support Ubuntu as build environment, where libressl isn't easily available, as far as i know.
Testing against the OpenSSL 1.0 API is possible by installing libssl1.0-dev, that could be easily done by someone who is more experienced Travis.

pabuhler:
There is always docker ...

PR #414, opened 2018-05-30
Detect and handle broken "OPENSSL_cleanse".

fancycode:
On some platforms (experienced this on arm64 with OpenSSL 1.0.2g), the
function `OPENSSL_cleanse` is broken and doesn't clear the memory, but
fills it with data. This breaks `octet_string_set_to_zero`.

This patch checks if `OPENSSL_cleanse` is broken during configure and if
it is, will use `srtp_cleanse` instead. Also a testcase is added to
`datatypes_driver.c` to check if `octet_string_set_to_zero` is working
correctly.

### DIFF
--- a/configure
+++ b/configure
@@ -5565,6 +5565,60 @@ $as_echo "#define OPENSSL 1" >>confdefs.h
    USE_OPENSSL=1
 
 
+   { $as_echo "$as_me:${as_lineno-$LINENO}: checking if OPENSSL_cleanse is broken" >&5
+$as_echo_n "checking if OPENSSL_cleanse is broken... " >&6; }
+   if test "$cross_compiling" = yes; then :
+  { { $as_echo "$as_me:${as_lineno-$LINENO}: error: in \`$ac_pwd':" >&5
+$as_echo "$as_me: error: in \`$ac_pwd':" >&2;}
+as_fn_error $? "cannot run test program while cross compiling
+See \`config.log' for more details" "$LINENO" 5; }
+else
+  cat confdefs.h - <<_ACEOF >conftest.$ac_ext
+/* end confdefs.h.  */
+
+     #include <stdio.h>
+     #include <openssl/crypto.h>
+
+int
+main ()
+{
+
+     #define BUFFER_SIZE (16)
+     char buffer[BUFFER_SIZE];
+     int i;
+     for (i = 0; i < BUFFER_SIZE; i++) {
+       buffer[i] = i & 0xff;
+     }
+     OPENSSL_cleanse(buffer, BUFFER_SIZE);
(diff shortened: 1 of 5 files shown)


fancycode:
Output from the new test.

Before:
```
Buffer before: 000102030405060708090a0b0c0d0e0f
Buffer after: 001225394e647b93acc6e1fd1a385777
Buffer contents not zero at position 1 (is 18)
Aborted
```

After:
```
Buffer before: 000102030405060708090a0b0c0d0e0f
Buffer after: 00000000000000000000000000000000
```

crypto/test/datatypes_driver.c
```diff
@@ -228,3 +231,26 @@ void test_bswap(void)
     printf("bswapped octet string: %s\n",
            octet_string_hex_string((uint8_t *)&y, 8));
 }
+
+void test_set_to_zero(void)
+{
+#define BUFFER_SIZE 16
```
thisisG: Should probably have () around the numeric value.

configure.ac
```diff
@@ -254,6 +254,31 @@ if test "$enable_openssl" = "yes"; then
    HMAC_OBJS=crypto/hash/hmac_ossl.o
    AC_SUBST([USE_OPENSSL], [1])
 
+   AC_MSG_CHECKING([if OPENSSL_cleanse is broken])
+   AC_RUN_IFELSE([AC_LANG_PROGRAM([
+     #include <stdio.h>
+     #include <openssl/crypto.h>
+   ], [
+     static const int bufferSize = 16;
```
thisisG: Should probably use a define - this may not always compile (cleanly at least).

configure
```diff
@@ -5565,6 +5565,60 @@ $as_echo "#define OPENSSL 1" >>confdefs.h
...
+/* end confdefs.h.  */
+
+     #include <stdio.h>
+     #include <openssl/crypto.h>
+
+int
+main ()
+{
+
+     static const int bufferSize = 16;
```
thisisG: Should probably use a define - this may not always compile (cleanly at least).

fancycode:
Thanks for your review and patch! Applied and squashed the commits.

thisisG:
LGTM! :tada: 

Will leave it up for a bit in case there are other comments.

abelbeck:
Using release 2.3.0 containing this change, the test hard fails when `cross compiling` ...
```
checking if OPENSSL_cleanse is broken... configure: error: in `/home/dev/astlinux/trunk/output/build/libsrtp-2.3.0':
configure: error: cannot run test program while cross compiling
See `config.log' for more details
```
Possibly make `openssl_cleanse_broken` cache-able so the build environment can set it  when cross compiling ? aka `ac_cv_openssl_cleanse_broken` ?

PR #416, opened 2018-07-05
Coverity fixes in tests

pabuhler:
Coverity is now enabled for test source files.  This attempts to fix most of the problems reported.

### DIFF
--- a/test/rdbx_driver.c
+++ b/test/rdbx_driver.c
@@ -348,6 +348,9 @@ double rdbx_check_adds_per_second(int num_trials, unsigned long ws)
             ++failures;
     }
     timer = clock() - timer;
+    if (timer < 1) {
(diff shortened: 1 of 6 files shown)


test/rdbx_driver.c
```diff
@@ -348,6 +348,9 @@ double rdbx_check_adds_per_second(int num_trials, unsigned long ws)
             ++failures;
     }
     timer = clock() - timer;
+    if (timer < 1) {
```
paulej: This and the one other place where `t = clock() -t` is done will prevent a possible divide by zero.  However, it's possible that clock() might return -1 if there is an error.  Do we want to just return call exit() if clock will not function properly?

pabuhler: Potentially

PR #422, opened 2018-08-07
Double GCM

bifurcation:
This PR replaces #311 as the proposed way forward for PERC double-GCM encryption.

https://tools.ietf.org/html/draft-ietf-perc-double


crypto/cipher/aes_gcm_double.c
```diff
@@ -0,0 +1,688 @@
...
+ */
+static srtp_err_status_t srtp_aes_gcm_double_alloc(srtp_cipher_t **c,
+                                                int key_size,
+                                                int tlen)
+{
+    srtp_err_status_t err;
+    srtp_aes_gcm_double_ctx_t *dbl;
+
+    debug_print(srtp_mod_aes_gcm, "allocating cipher with key length %d",
+                key_size);
```
pabuhler: use srtp_mod_aes_gcm_double in debug_print functions in this file

crypto/cipher/aes_gcm_double.c
```diff
@@ -0,0 +1,688 @@
...
+        return (srtp_err_status_bad_param);
+    }
+
+    /* allocate the base structs */
+    *c = (srtp_cipher_t *)srtp_crypto_alloc(sizeof(srtp_cipher_t));
+    if (*c == NULL) {
+        return (srtp_err_status_alloc_fail);
+    }
+    memset(*c, 0x0, sizeof(srtp_cipher_t));
+
```
pabuhler: srtp_crypto_alloc returns memory that is already initialed to zero so the calls to memset are not required.

crypto/cipher/aes_gcm_double.c
```diff
@@ -0,0 +1,688 @@
...
+      debug_print(srtp_mod_aes_gcm_double, "error alloc inner: %d", err);
+      srtp_crypto_free(*c);
+      *c = NULL;
+      return err;
+    }
+
+    err = base_type->alloc(&dbl->outer, base_key_size_wsalt, GCM_AUTH_TAG_LEN);
+    if (err != srtp_err_status_ok) {
+      debug_print(srtp_mod_aes_gcm_double, "error alloc inner: %d", err);
+      srtp_crypto_free(*c);
```
pabuhler: change msg to error alloc outer

include/srtp.h
```diff
@@ -69,13 +69,18 @@ extern "C" {
 /*
  * SRTP_MAX_KEY_LEN is the maximum key length supported by libSRTP
  */
-#define SRTP_MAX_KEY_LEN 64
+#define SRTP_MAX_KEY_LEN 96
+
+/*
+ * SRTP_MAX_KEY_LEN is the maximum IV length supported by libSRTP
+ */
+#define SRTP_MAX_IV_LEN 24
```
pabuhler: Why is this in public header file? 
Also the comment is wrong.

include/srtp.h
```diff
@@ -1147,6 +1164,52 @@ void srtp_crypto_policy_set_aes_gcm_128_16_auth(srtp_crypto_policy_t *p);
...
+ * with the policy data structure.  You are encouraged to initialize
+ * policy elements with this function call.  Doing so may allow your
+ * code to be forward compatible with later versions of libSRTP that
+ * include more elements in the srtp_crypto_policy_t datatype.
+ *
+ * @return void.
+ *
+ */
+void srtp_crypto_policy_set_aes_gcm_256_double(srtp_crypto_policy_t *p);
+
```
pabuhler: update the function documentation to references these functions, ie add _double

include/srtp.h
bifurcation: It's in the public header file to be parallel to `SRTP_MAX_KEY_LEN`, `SRTP_MAX_TAG_LEN`, etc.  Do you think that because the IV is generated by libsrtp, it should go in the private header?  Happy to move it, just want to understand the rationale.

How is the comment wrong?

bifurcation:
@pabuhler Looks like CI passes now, if you'd like to give this a re-review.

include/srtp.h
pabuhler: SRTP_MAX_KEY_LEN  => SRTP_MAX_IV_LEN  in the comment

Yes that was my rational but I am ok with it staying here.

pabuhler:
@bifurcation, I have unblocked the commit but I am still reviewing the rest of the code so will provide a new review soon.  There are few places in the code that you have questions (as code comments), so I am trying to understand the problem correctly to give answers.

crypto/cipher/aes_gcm_double.c
```diff
@@ -0,0 +1,686 @@
...
+    ct_size += tag_size;
+
+    debug_print(srtp_mod_aes_gcm_double, "outer ciphertext: %s",
+                srtp_octet_string_hex_string(ct, ct_size));
+
+    /* Copy the ciphertext and tag to their respective buffers */
+    memcpy(buf, ct, pt_size);
+
+    c->tag_size = ct_size - pt_size;
+    memcpy(c->tag, ct + pt_size, c->tag_size);
```
paulej: This is copying the encrypted tag after the double encryption of the inner tag + the plaintext tag from the outer encryption, I think. Do we want that?  While there are two tags (inner and outer), should we not just report the outer tag value?  (I cannot recall where this is used, so maybe this makes sense.  I don't know.)

crypto/cipher/aes_gcm_double.c
```diff
@@ -0,0 +1,686 @@
...
+        return err;
+    }
+
+    ct_size += tag_size;
+
+    debug_print(srtp_mod_aes_gcm_double, "outer ciphertext: %s",
+                srtp_octet_string_hex_string(ct, ct_size));
+
+    /* Copy the ciphertext and tag to their respective buffers */
+    memcpy(buf, ct, pt_size);
```
paulej: Did you intend to use pt_size here? I think that's the original packet size.  I assume you want ct_size here?

srtp/srtp.c
```diff
@@ -1762,9 +1983,18 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
...
+     * double-size IV formation
      */
-    srtp_calc_aead_iv(session_keys, &iv, &est, hdr);
+    if (session_keys->rtp_cipher->algorithm == SRTP_AES_GCM_128_DOUBLE ||
+        session_keys->rtp_cipher->algorithm == SRTP_AES_GCM_256_DOUBLE) {
+        srtp_calc_double_aead_iv(session_keys, &iv_inner, &iv_outer, &est, hdr);
+        memcpy(iv, &iv_inner, SRTP_AEAD_SALT_LEN);
+        memcpy(iv + SRTP_AEAD_SALT_LEN, &iv_outer, SRTP_AEAD_SALT_LEN);
+    } else {
+        srtp_calc_aead_iv(session_keys, &iv_outer, &est, hdr);
```
paulej: Why was this code changed for non-double GCM use?

srtp/srtp.c
```diff
@@ -1762,9 +1983,18 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
...
+        memcpy(iv, &iv_outer, SRTP_AEAD_SALT_LEN);
```
paulej: Seems we get the same result, but with an extra memcpy

srtp/srtp.c
```diff
@@ -1875,20 +2109,30 @@ static srtp_err_status_t srtp_unprotect_aead(srtp_ctx_t *ctx,
...
```
paulej: Same question as before w.r.t. setting the IV for normal GCM.

PR #423, opened 2018-08-07
Travis nss build

pabuhler:
Add nss build to traivs on linux & osx.
nss is version 3.38 and built locally using the `make nss_build_all` command.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,6 +87,26 @@ matrix:
         - make
         - make runtest
 
+    # osx build with nss
+    - os: osx
+      env:
+        - TEST="osx (nss)"
+      osx_image: xcode8.2
+      before_install:
+            - wget https://ftp.mozilla.org/pub/security/nss/releases/NSS_3_38_RTM/src/nss-3.38-with-nspr-4.19.tar.gz
+            - tar -xzvf nss-3.38-with-nspr-4.19.tar.gz
+            - pushd nss-3.38/nss
+            - USE_64=1 make nss_build_all
+            - pushd ../dist/$(<../dist/latest)
+            - export NSS_DIR=$(pwd)
+            - echo $NSS_DIR
+            - popd
+            - popd
+      script:
+        - EXTRA_CFLAGS=-Werror ./configure --enable-nss --with-nss-dir=$NSS_DIR
+        - make
+        - make runtest
+
     # code format check
     - os: linux
       env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,31 @@ matrix:
         - make
         - make runtest
 
+    # linux build with nss
+    - os: linux
+      env:
+        - TEST="linux gcc (nss)"
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-6
+      before_install:
+            - wget https://ftp.mozilla.org/pub/security/nss/releases/NSS_3_38_RTM/src/nss-3.38-with-nspr-4.19.tar.gz
+            - tar -xzvf nss-3.38-with-nspr-4.19.tar.gz
+            - pushd nss-3.38/nss
+            - USE_64=1 make nss_build_all
+            - pushd ../dist/$(<../dist/latest)
+            - export NSS_DIR=$(pwd)
+            - echo $NSS_DIR
+            - popd
+            - popd
+      script:
+        - CC=gcc-6 EXTRA_CFLAGS=-Werror ./configure --enable-nss --with-nss-dir=$NSS_DIR
+        - make
+        - make runtest
+
     # default osx build with xcode (clang)
     - os: osx
       env:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -52,6 +52,31 @@ matrix:
...
+      env:
+        - TEST="linux gcc (nss)"
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - gcc-6
+      before_install:
+            - wget https://ftp.mozilla.org/pub/security/nss/releases/NSS_3_38_RTM/src/nss-3.38-with-nspr-4.19.tar.gz
```
bifurcation: This is too brute-force; the link will become stale almost immediately.

It seems like the better course would be to use the `libnss3` package in Ubuntu.  That will require some tweaks to the configure / build process to specify the include and library paths directly, rather than inferring them from a base.  Making these tweaks in the Makefile manually, this seems to work fine with the Ubuntu standard NSS installation.

Based on the [CircleCI docs](https://circleci.com/docs/1.0/ios-builds-on-os-x/), it looks like you might be able to do the same thing with Homebrew on MacOS.

pabuhler: I can look at that, using packages though will mean there is less control of the version that is supported.
Ideally when building against a system package will mean there is no need for --with-nss-dir .

pabuhler:
@bifurcation, now both system installs and local builds of NSS should work. Will this be ok ?

itollefsen:
Looking at the latest commit, "support system NSS packages", it'll work. Just some nitpicks, possibly for a later revision.

Regarding this snippet from `confiure.ac`:

```
if test "x$PKG_CONFIG" != "x" && test "$nss_skip_pkg_config" != "yes"; then
 PKG_CHECK_MODULES([nss], [nss],
   [CFLAGS="$CFLAGS $nss_CFLAGS"
     LIBS="$nss_LIBS $LIBS"])
else
 LIBS="-lnss3 -lnssutil3 -lnspr4 $LIBS"
fi
```

Ideally, the `else` part should  use `AC_CHECK_LIB`. Something along the lines of:

```
  AC_CHECK_LIB(
    [nspr4], [<function-used-from-nspr4>],
    [], [AC_MSG_WARN([can't find libnspr4])])
  AC_CHECK_LIB(
    [nssutil3], [<function-used-from-nssutil3>],
    [], [AC_MSG_WARN([can't find libnssutil3])])
  AC_CHECK_LIB(
    [nss3], [<function-used-from-nss3>],
    [], [AC_MSG_FAILURE([can't find useable libnss3])])
```

This will check that the libraries are present and useable, at the configuration stage, instead of failing the build at the linking stage. The end result will be the exact same `LIBS` modifications.

Note that the prerequisites use `AC_MSG_WARN` and only a failed `libnns3` check will result in a `AC_MSG_FAILURE` and abort.

This is the same pattern used by the OpenSSL checks. The idea is that you're only really interested in `libnss3`. If you can use it without the dependencies, then that's fine. You don't _have_ to have the dependencies if they are not needed.

The tricky part of this is to figure out which function to check for. It's probably quite easy for `libnss3` because that's the functions you're calling from this project. But for the dependencies, you would want to use a function that is needed by your code path.

One way of determining what you need is to leave them out and look at the undefined references you'd get from linking. And then add them back in one by one with one of those functions as the check.

The `pkg-config` approach is fine and will include them as long as the `nss.pc` file does the right thing. Meaning it will either have them listed in the various '.private' sections and we'll pick them up because we're using `pkg-config --static`, or it provides them unconditionally (because most folks don't know to use `--static`).

I also noticed that there are no header checks. Not for OpenSSL either, from which this is copied.

It would be something along the lines of:
```
AC_CHECK_HEADERS(
  [<nss-header-file>],
  [], [AC_MSG_FAILURE([can't find useable NSS headers])],
  [AC_INCLUDES_DEFAULT])
```
And similarly for OpenSSL:
```
AC_CHECK_HEADERS(
  [openssl/crypto.h],
  [], [AC_MSG_FAILURE([can't find useable OpenSSL headers])],
  [AC_INCLUDES_DEFAULT])
```

This is probably more important when `--with-nss-dir` is used without `pkg-config` then the other scenarios, to verify that the include path added to `CFLAGS` based on that value actually works. But it can be added unconditionally below the library checks (the `if.. else..` quoted above, and below `AC_SEARCH_LIBS` for OpenSSL), so that it will check for them always.

pabuhler:
@itollefsen , I added the AC_CHECK_´s only in the case where pkg-config is not used. If in the pkg-config case it does not add the correct libs then I think it is best that it fails.

itollefsen:
> I added the AC_CHECK_´s only in the case where pkg-config is not used. If in the pkg-config case it does not add the correct libs then I think it is best that it fails.

Makes sense, agreed.

However, did you now accidentally leave out `libnssutil3`? There's no `AC_CHECK_LIB` for it between the check for `libnspr4`and `libnss3`.

Other than that, this looks good.

pabuhler:
It seamed to link and run without libnssutil3 so left it out for now.

PR #428, opened 2018-09-28
enable valgrind in nss builds

pabuhler:
After enabling valgrind with nss , static leaks where reported from NSS_NoDB_Init().
According to https://wiki.mozilla.org/NSS_Library_Init the correct solution for libsrtp was to change to NSS_InitContext().

@bifurcation what do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,40 +9,44 @@ env:
 matrix:
   include:
 
-    # default linux build with gcc
+    # linux build
     - os: linux
       env:
-        - TEST="linux gcc"
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
-          packages:
-          - gcc-6
-      script:
-        - CC=gcc-6 EXTRA_CFLAGS=-Werror ./configure
-        - make
-        - make runtest
-
-    # linux build with openssl and gcc
-    - os: linux
-      env:
-        - TEST="linux gcc (openssl)"
+        - TEST="linux (gcc / valgrind)"
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
           packages:
             - gcc-6
+            - valgrind
+      script:
+        - CC=gcc-6 EXTRA_CFLAGS=-Werror ./configure
+        - make
+        - make runtest
+        - make runtest-valgrind
+
+    # linux build with openssl
+    - os: linux
+      env:
+        - TEST="linux openssl (gcc / valgrind)"
+      addons:
+        apt:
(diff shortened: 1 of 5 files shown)


bifurcation:
Actually, I think the right thing to do is probably to extend `srtp_crypto_kernel_init()` and `srtp_crypto_kernel_shutdown()` to do NSS init and shutdown.  If we do that, I think we can keep the simplicity of `NSS_NoDB_Init()`, at the cost of making sure we always call `NSS_Shutdown()`.

CC @mozkeeler in case she has thoughts

nils-ohlmeier:
Sorry I forgot that I had internally contacted @mozkeeler already and got an okay for the change. But I'll let @mozkeeler chime in here and provide feedback which route is better.

mozkeeler:
I once investigated using `NSS_InitContext` and `NSS_ShutdownContext` in gecko and found that it wasn't suitable for our purposes, although unfortunately I don't recall why. That said, this is a different situation and I don't see any reason why in theory this patch as-is wouldn't work. If you do have a single shutdown location and you know that everything that uses NSS is going to (or can be made to) go away before reaching that point, then in my experience `NSS_NoDB_Init` as needed and a single `NSS_Shutdown` is simple and effective. If the code has to coexist with other code that also uses NSS, then obviously that won't work (but hopefully the context-based solution will).

nils-ohlmeier:
We do have some changes coming up in Firefox which will result in this SRTP code co-existing inside a process where others parts also rely on NSS. I guess at that point me might have to add options to no longer initialize and shutdown NSS by itself. But I think we will cross that bridge once we get there.

pabuhler:
@mozkeeler I would say that as libsrtp is a library, there is no way to know if it will co-exist with any other NSS code in an application so using the context-based solution is the only valid solution. If there is no major overhead then keeping the way it is in this patch is the easiest solution, moving it out to srtp_crypto_kernel_init() and srtp_crypto_kernel_shutdown() is also valid but I would still argue that the context-based solution is the way to go. An alternative is not to call any NSS init functions from inside libsrtp and require that the application using libsrtp initialize NSS, this is what happens with openssl today.

pabuhler:
@bifurcation & @nils-ohlmeier, my intention with this PR was to enable valgrind on CI builds of NSS, not necessarily find the best solution for integrating NSS initialization. Based on the previous comments my change does not seam to be incorrect, so unless there is a strong objection I will merge this PR and if any issues with NSS initialization arise we can deal with it then.

PR #429, opened 2018-09-28
Rand for test only

pabuhler:
Reduce occurrence of rand() in code base and make it explicit that it is for testing only.

### DIFF
--- a/crypto/cipher/cipher.c
+++ b/crypto/cipher/cipher.c
@@ -199,7 +199,7 @@ void srtp_cipher_rand_for_tests(void *dest, uint32_t len)
 }
 
 /*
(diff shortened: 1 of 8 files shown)


crypto/cipher/cipher.c
```diff
@@ -195,7 +196,17 @@ static srtp_err_status_t srtp_cipher_rand(void *dest, uint32_t len)
         len--;
     }
 #endif
-    return srtp_err_status_ok;
+}
+
+/*
+ * A trivial platform independent 64 bit random number.
```
nils-ohlmeier: 64 bit?  Isn't this 32bit?

itollefsen:
+1

Issue #411, opened 2018-04-30
Failed to build shared library for iOS with bitcode support


AlexanderplUs:
I've previously successfully compile static library `libsrtp.a` for iOS with bitcode enabled but when I switch to shared library I get next error:
```
ld: -undefined and -bitcode_bundle (Xcode setting ENABLE_BITCODE=YES) cannot be used together
```

Here is my config:
```
CPPFLAGS="-arch armv7 -mcpu=cortex-a8 -mfpu=neon -mfloat-abi=softfp -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk -miphoneos-version-min=8.0 -DTARGET_OS_IPHONE=1 -fembed-bitcode"\
LDFLAGS="-arch armv7 -mcpu=cortex-a8 -mfpu=neon -mfloat-abi=softfp -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk -miphoneos-version-min=8.0 -DTARGET_OS_IPHONE=1 -fembed-bitcode" \
srtp/configure -prefix=output/armv7-apple-darwin --host=armv7-apple-darwin
```

I make shared library using `make shared_library`

I can make build work just removing `-undefined dynamic_lookup` key from linking parameters in `Makefile.in`:
```
--- Makefile.in.orig	2018-04-25 21:03:51.000000000 +0400
+++ Makefile.in	2018-04-30 15:44:01.000000000 +0400
@@ -87,7 +87,7 @@
 SHAREDLIBSUFFIX = $(SHAREDLIBSUFFIXNOVER)
 else ifeq (darwin,$(findstring darwin,@host@))
 SHAREDLIB_DIR = $(libdir)
-SHAREDLIB_LDFLAGS = -dynamiclib -twolevel_namespace -undefined dynamic_lookup \
+SHAREDLIB_LDFLAGS = -dynamiclib -twolevel_namespace \
 		-fno-common -headerpad_max_install_names -install_name $(libdir)/$@
 SHAREDLIBSUFFIXNOVER = dylib
 SHAREDLIBSUFFIX = $(SHAREDLIBVERSION).$(SHAREDLIBSUFFIXNOVER)
```

[...]

PR #432, opened 2018-10-18
Remove flags incompatible with -fembed-bitcode for iOS

ocrete:
This simple should fix issue #411, we've been using it in the GStreamer build for a while and I just notice it hadn't been uptreamed.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -116,8 +116,8 @@ SHAREDLIBSUFFIXNOVER = dll
 SHAREDLIBSUFFIX = $(SHAREDLIBSUFFIXNOVER)
 else ifeq (darwin,$(findstring darwin,@host@))
 SHAREDLIB_DIR = $(libdir)
-SHAREDLIB_LDFLAGS = -dynamiclib -twolevel_namespace -undefined dynamic_lookup \
-		-fno-common -headerpad_max_install_names -install_name $(libdir)/$@
+SHAREDLIB_LDFLAGS = -dynamiclib -twolevel_namespace -fno-common \
+		-install_name $(libdir)/$@
 SHAREDLIBSUFFIXNOVER = dylib
 SHAREDLIBSUFFIX = $(SHAREDLIBVERSION).$(SHAREDLIBSUFFIXNOVER)
 endif


itollefsen:
I think you inadvertently removed `-headerpad_max_install_names` in addition to the `-undefined dynamic_lookup` flag you meant to remove.

Removing `-undefined` is fine. It defaults to `error`. I'm not sure why it was set to `dynamic_lookup`in the first place. Could be historical reasons I guess.

But we want to keep `-headerpad_max_install_names` to reserve space in the header for changing that name later to the actual install location (if necessary).

itollefsen:
I think you inadvertently removed `-headerpad_max_install_names` in addition to the `-undefined dynamic_lookup` flag you meant to remove.

Removing `-undefined` is fine. It defaults to `error`. I'm not sure why it was set to `dynamic_lookup`in the first place. Could be historical reasons I guess.

But we want to keep `-headerpad_max_install_names` to reserve space in the header for changing that name later to the actual install location (if necessary).

itollefsen:
I think you inadvertently removed `-headerpad_max_install_names` in addition to the `-undefined dynamic_lookup` flag you meant to remove.

Removing `-undefined` is fine. It defaults to `error`. I'm not sure why it was set to `dynamic_lookup`in the first place. Could be historical reasons I guess.

But we want to keep `-headerpad_max_install_names` to reserve space in the header for changing that name later to the actual install location (if necessary).

pabuhler:
@ocrete  Can you confirm that removing headerpad_max_install_names was a oversight?

ocrete:
> @ocrete Can you confirm that removing headerpad_max_install_names was a oversight?

Tbh, I don't remember. I think removing it was required for the iOS builds to succeed. But one should probably test with the latest XCode as they may have improved things.

itollefsen:
Actually,  `-headerpad_max_install_names`  has no effect if you use `-fembed-bitcode`. Bitcode is "just" another IR representation, and it will be recompiled again for the target architecture. Hence it makes no sense to add space for binary patching. The linker will issue a warning if the two are combined. This will presumably fail if warnings are treated as errors.

If you want to submit apps to the app store today for any of the Apple platforms, you must include bitcode. Unfortunately, the current check is for whether the host platform is Darwin. That doesn't help at all in determining whether you want to embed bitcode or not.

Perhaps this could be solved by a configure option to enable bitcode? Which would then omit `-headerpad_max_install_names` etc.

pabuhler:
This should be resolved now as part of #472, just reopen if there are still issues.

PR #434, opened 2018-12-03
Fixed compatibility with Visual Studio 2010

jacgoudsmit:
From the history of the Visual Studio project files, it looks like compatibility with VS2010 was intended since they were added in 701bb796bfb8a512859f7147649e2b835a321ff2. This change accomplishes that Visual Studio 2010 or higher can be used to build libsrtp with no other changes.

- Removed the specified platform toolset "v140" from the project file because it's probably irrelevant anyway. The Platform Toolset determines which compiler etc. to use for building, and each version of Visual Studio has its own default. There is no reason to override the default, unless there is a known security bug that I'm not aware of.
- Modified the err.c and srtp.c files so that all declarations of variables appear before any executable code. VS2010 doesn't support this. No functional changes were made to the code.
- Modified the srtp_cipher_rand_for_tests( ) function so it doesn't try to return an error from a function that returns void.
- Modified .gitignore to ignore Visual Studio related files

### DIFF
--- a/crypto/cipher/cipher.c
+++ b/crypto/cipher/cipher.c
@@ -171,7 +171,7 @@ int srtp_cipher_get_key_length(const srtp_cipher_t *c)
  */
 void srtp_cipher_rand_for_tests(void *dest, uint32_t len)
 {
-#if 0 //defined(HAVE_RAND_S) 
+#if 0 // defined(HAVE_RAND_S) 
     uint8_t *dst = (uint8_t *)dest;
     while (len) {
         unsigned int val;
--- a/crypto/cipher/cipher.c
+++ b/crypto/cipher/cipher.c
@@ -171,7 +171,7 @@ int srtp_cipher_get_key_length(const srtp_cipher_t *c)
  */
 void srtp_cipher_rand_for_tests(void *dest, uint32_t len)
 {
-#if defined(HAVE_RAND_S)
+#if 0 //defined(HAVE_RAND_S) 
     uint8_t *dst = (uint8_t *)dest;
     while (len) {
         unsigned int val;
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -696,9 +696,7 @@ static srtp_err_status_t srtp_kdf_init(srtp_kdf_t *kdf,
 
     {
         srtp_err_status_t stat;
-        stat = srtp_crypto_kernel_alloc_cipher(cipher_id,
-                                               &kdf->cipher,
-                                               key_len,
+        stat = srtp_crypto_kernel_alloc_cipher(cipher_id, &kdf->cipher, key_len,
                                                0);
         if (stat)
             return stat;
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -88,7 +88,7 @@ static srtp_err_status_t srtp_validate_rtp_header(void *rtp_hdr,
     int rtp_header_len = octets_in_rtp_header + 4 * hdr->cc;
 
     if (*pkt_octet_len < octets_in_rtp_header)
-      return srtp_err_status_bad_param;
+        return srtp_err_status_bad_param;
 
     if (hdr->x == 1)
         rtp_header_len += octets_in_rtp_extn_hdr;
@@ -696,7 +696,10 @@ static srtp_err_status_t srtp_kdf_init(srtp_kdf_t *kdf,
 
     {
         srtp_err_status_t stat;
-        stat = srtp_crypto_kernel_alloc_cipher(cipher_id, &kdf->cipher, key_len, 0);
+        stat = srtp_crypto_kernel_alloc_cipher(cipher_id,
+                                               &kdf->cipher,
+                                               key_len,
+                                               0);
         if (stat)
             return stat;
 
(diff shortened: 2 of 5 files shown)


srtp/srtp.c
```diff
@@ -82,13 +82,14 @@ srtp_debug_module_t mod_srtp = {
...
 {
-    if (*pkt_octet_len < octets_in_rtp_header)
-        return srtp_err_status_bad_param;
-
     srtp_hdr_t *hdr = (srtp_hdr_t *)rtp_hdr;
 
     /* Check RTP header length */
     int rtp_header_len = octets_in_rtp_header + 4 * hdr->cc;
+
+    if (*pkt_octet_len < octets_in_rtp_header)
```
paulej: If the packet length is too short, the line above that reads into the header can cause an access violation.  We need to keep this line at the beginning of the function.

crypto/cipher/cipher.c
```diff
@@ -171,7 +171,7 @@ int srtp_cipher_get_key_length(const srtp_cipher_t *c)
  */
 void srtp_cipher_rand_for_tests(void *dest, uint32_t len)
 {
-#if defined(HAVE_RAND_S)
+#if 0 // defined(HAVE_RAND_S)
```
paulej: Do we not need to check to see if this is available for some platforms?

srtp/srtp.c
```diff
@@ -693,16 +694,20 @@ static srtp_err_status_t srtp_kdf_init(srtp_kdf_t *kdf,
         break;
     }
 
-    srtp_err_status_t stat;
-    stat = srtp_crypto_kernel_alloc_cipher(cipher_id, &kdf->cipher, key_len, 0);
-    if (stat)
-        return stat;
+    {
```
paulej: VS 2010 complains about variables defined further down?  If so, I'd be OK with moving "stat" to the top of the function.  I think it's unfortunate to introduce indentation just to address that issue.  (Plus, I think it's better to define variables at the top, anyway.  So, I'd personally prefer moving it.)

crypto/cipher/cipher.c
jacgoudsmit: - The rand( ) function is part of the standard C library and should be available on all platforms, so the code in the #else part is always valid. However the rand( ) function may not give enough entropy and is otherwise also not very good because it returns limited values (as indicated by the remarks in the code).
- rand_s( ) is "better" than rand( ), it probably has better entropy (citation needed) and generates a wider variety of numbers. It also may return an error if there is something wrong. I'm not familiar with it (I use OpenSSL's randomizer for my applications) but I suspect that it returns an error if the random number generator isn't initialized or doesn't have enough entropy yet.
- The code in the #if section correctly checks for the error but incorrectly breaks off the process of filling the buffer with random values. It tries to return an error code but the function is declared void, so on Visual Studio that results in a compiler warning C4098 "void function returning a value".

So imagine the case that srtp_cipher_rand_for_tests( ) gets called and the code in the #if section is executed and srand_s returns an error. This results in the buffer not being filled or only being partially filled (the latter is unlikely but technically possible depending on srand_s's behavior). And functions that call srtp_cipher_rand_for_tests (and of course don't know that something went wrong) will get a buffer that may have "random" bytes but not as random as they wanted (basically the buffer will be unchanged from what it was before).

I concluded that:
- The behavior of the code in the #if block is wrong
- It's possible that the code in the #if block is not usually tested (is HAVE_RAND_S only defined for Windows builds?) or the warning about the void function returning a value is ignored by engineers
- The function is obviously used for testing only, and apparently the engineers who wrote the code and know more about intricacies of SRTP have decided that at least in SOME cases, testing with rand( ) would be sufficient
- Fixing the code "the right way" to eliminate the warning would require that I either (A) change the function's return type and change all the places where it's called to check the return value, or (B) change the code so that if rand_s fails, it would fall back to the rand( ) code. And if it would be sufficient to fall back to the rand( ) code, why not use that in the first place?
- This problem is ultimately unrelated to the problem that I'm trying to fix (compatibility with Visual Studio 2010) and I'm not familiar with the code so I didn't want the size of the change to go out of hand. 

The easiest way to work around the problem was to eliminate the #if section. I intended to explain this in my commit comment but I couldn't think of short way to explain my way of thinking.

pabuhler:
Hi @jacgoudsmit ,
Given the current code in master alot of these changes are now redundant, I have created PR #456 as alternative. With #456 I have successfully built on vc 2010 and run the two test applications that are currently part of the cmake support. Can we now close this PR or are there things that #456 do not cover.

jacgoudsmit:
> Given the current code in master alot of these changes are now redundant, I have created PR #456 as alternative. With #456 I have successfully built on vc 2010 and run the two test applications that are currently part of the cmake support. Can we now close this PR or are there things that #456 do not cover.

Out of all the changes I requested, the only ones that remain absent from the latest "master" commit (9b77ac5e27121c8b2ede50e3c501fd2c5d7c97b2) are:

- The changes to .gitignore to prevent Visual Studio temporary files from being checked in, and
- The change to srtp2.vcxproj to delete the eight lines that specify the platform toolset as v141. Without those deletions, you can only build if you also have a later version of Visual Studio installed on your system. If the toolset version is specifically set to a particular version, it will try to use that version; otherwise it will use whatever the default is for the installed Visual Studio. VS2010 uses toolset v100. 

Both problems are fairly trivial. I'll close the pull request but hopefully the two changes can still be integrated.

Thanks Pascal!

PR #439, opened 2019-02-28
chacha20 poly1305

marakew:
how about add unsupported ChaCha20 Poly1305 ?


pabuhler:
Hi @marakew , before reviewing this change do you have any reference to some documentation that describes how  ChaCha20 Poly1305 should be used in a SRTP context? The fact that it is not standardized yet is not really an issue but it still needs to be documented.

marakew:
i saw how chacha20-poly1305 implemented in SSL as example in chrome
https://chromium.googlesource.com/chromium/src/net/third_party/nss/+/8d4d7db726f0c441cf0798ae0bc7c4d25b8b670a/patches/chacha20poly1305.patch
so i decide the chacha20-poly1305 very simular like aes-gcm
and iam reimplement existing aes-gcm srtp implementation
yes chacha20-poly1305 still not documented and this is futures
but i can't see any ploblem for use it
i add profile 

@@ -1180,6 +1190,8 @@ typedef enum {
    srtp_profile_null_sha1_32 = 6,
    srtp_profile_aead_aes_128_gcm = 7,
    srtp_profile_aead_aes_256_gcm = 8,

    srtp_profile_aead_chacha20_poly1305 = 16, // TODO unregister
} srtp_profile_t;

which  have not in standart
but i hope some one prepare rfc draft for this
before this any one can use this encryption as custom private implementation
i have plan to prepare patch for asterisk and linphone to testing this futures

pabuhler:
Hi @marakew, 
We have decided that we can not merge this in to libSRTP without having some kind of publicly available reference documentation. If you would like to create such public reference then we can help you get started. RFC 7714 exists today with a generic section for SRTP and AEAD operation but it contains a specific section for the GCM cipher. There would need to be equivalent documentation regarding chacha20 & poly1305. That RFC would be a good place to start.

In the mean time we could create a branch for this feature so people are free to experiment with it and it can be updated as documentation becomes available.
Would you like us to create such a branch ?

marakew:
can you create the branch or should i do this ?

pabuhler:
If you are ok with it going in to a branch for now then I will create it one and update the PR

PR #440, opened 2019-03-14
Rtp decoder update

pabuhler:
Update the rtp_decoder to changes in libsrtp from the last few years.
Includes better support for AES_GCM ciphers and AES_CM 192/256 ciphers.

### DIFF
--- a/test/rtp_decoder.c
+++ b/test/rtp_decoder.c
@@ -130,6 +130,28 @@ static struct srtp_crypto_suite srtp_crypto_suites[] = {
     {.can_name = NULL }
 };
 
+void rtp_decoder_srtp_log_handler(srtp_log_level_t level,
+                                  const char *msg,
+                                  void *data)
+{
+    char level_char = '?';
+    switch (level) {
+    case srtp_log_level_error:
+        level_char = 'e';
+        break;
+    case srtp_log_level_warning:
+        level_char = 'w';
+        break;
+    case srtp_log_level_info:
+        level_char = 'i';
+        break;
+    case srtp_log_level_debug:
+        level_char = 'd';
+        break;
+    }
+    fprintf(stderr, "SRTP-LOG [%c]: %s\n", level_char, msg);
+}
+
 int main(int argc, char *argv[])
 {
     char errbuf[PCAP_ERRBUF_SIZE];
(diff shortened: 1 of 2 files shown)


kyzer-davis:
Verified the following work properly with decoding lab captures.
- support decoding base64 AEAD_AES_GCM key
- support -s with AEAD_AES_(128/256)_GCM
- support AES_CM 192 & 256 with -s
- update debug options to use new srtp log api
- add -t to usage summary

test/rtp_decoder.c
```diff
@@ -207,20 +267,35 @@ int main(int argc, char *argv[])
...
+    if (scs.tag_size == 0) {
+        if (gcm_on) {
+            scs.tag_size = 16;
+        } else {
+            scs.tag_size = 10;
+        }
+    }
+
     if (gcm_on && scs.tag_size != 8 && scs.tag_size != 16) {
         fprintf(stderr, "error: GCM tag size must be 8 or 16 (%d)\n",
```
paulej: Size 8 is not allowed for GCM per the spec.  Should we allow it here?  That would mean we remove srtp_crypto_policy_set_aes_gcm_128_8_auth as an option.

kyzer-davis: As discussed I don't see anything that uses a tag size of 8 for GCM w/ SRTP or SDES. The only valid value is 16.

https://www.iana.org/assignments/sdp-security-descriptions/sdp-security-descriptions.xhtml

https://tools.ietf.org/html/rfc7714
> 1.  Introduction
>       AEAD_AES_128_GCM      AES-128 with a **16-octet authentication tag**
>       AEAD_AES_256_GCM      AES-256 with a **16-octet authentication tag**

pabuhler: Since the -s option with AEAD_AES_128_GCM or AEAD_AES_256_GCM will set the tag length to 16 and the support for tag of 8 has been there for a while then I don't think I would want to change that in this commit.  Will instead open a new issue to potentially remove srtp_crypto_policy_set_aes_gcm_128_8_auth and friends from srtp.h, and if that happens then rtp_decoder will also be changed to remove support for 8 byte tag.

pabuhler: #441 created to track this.

PR #444, opened 2019-05-15
Extend rtp_ecoder to support RTCP and multiplexed SSRC's 

pabuhler:
This makes i more useful for parsing packet capture files from direct captures.

### DIFF
--- a/test/rtp_decoder.c
+++ b/test/rtp_decoder.c
@@ -562,6 +562,9 @@ int main(int argc, char *argv[])
 
     pcap_loop(pcap_handle, 0, rtp_decoder_handle_pkt, (u_char *)dec);
 
+    fprintf(stderr, "SRTP packets decoded: %d\n", dec->srtp_cnt);
+    fprintf(stderr, "Packet decode errors: %d\n", dec->error_cnt);
+
     rtp_decoder_deinit_srtp(dec);
     rtp_decoder_dealloc(dec);
 
@@ -628,6 +631,8 @@ int rtp_decoder_init(rtp_decoder_t dcdr, srtp_policy_t policy)
     dcdr->start_tv.tv_usec = 0;
     dcdr->start_tv.tv_sec = 0;
     dcdr->frame_nr = -1;
+    dcdr->error_cnt = 0;
+    dcdr->srtp_cnt = 0;
     dcdr->policy = policy;
     dcdr->policy.ssrc.type = ssrc_specific;
     return 0;
@@ -656,6 +661,7 @@ void rtp_decoder_handle_pkt(u_char *arg,
                             const u_char *bytes)
 {
     rtp_decoder_t dcdr = (rtp_decoder_t)arg;
+    rtp_msg_t message;
     int pktsize;
     struct timeval delta;
     int octets_recvd;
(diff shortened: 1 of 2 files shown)


kyzer-davis:
Verified the new -m flag work as expected
- SRTP > RTP
- SRTCP > RTCP
- SRTP+SRTCP MUX > RTP+RCTP MUX

Verified output counters display as expected

test/rtp_decoder.c
```diff
@@ -671,32 +702,51 @@ void rtp_decoder_handle_pkt(u_char *arg,
...
-        return;
+    if (dcdr->mode == mode_rtp) {
+        rtp = 1;
+    } else if (dcdr->mode == mode_rtcp) {
+        rtp = 0;
+    } else {
+        rtp = 1;
+        if (octets_recvd >= 2) {
+            /* rfc5761 */
+            u_char pt = *(bytes + dcdr->rtp_offset + 1) & 0x7f;
```
thisisG: Maybe use `payload_type` to be a bit more explicit? Makes RFC perusing a bit easier. :-)

test/rtp_decoder.c
```diff
@@ -671,32 +702,51 @@ void rtp_decoder_handle_pkt(u_char *arg,
...
+            rtp = 64 <= pt && pt >= 95;
```
thisisG: Is not the `64 <= pt` check redundant here since the expression `pt >= 95` will be `false` for all values from 0 to 94 (assuming the range 0 -> 127 due to the `0x7f` mask).

From the RFC:

   Given these constraints, it is RECOMMENDED to follow the guidelines
   in the RTP/AVP profile [7] for the choice of RTP payload type values,
   with the additional restriction that payload type values in the range
   64-95 MUST NOT be used.

Seems like the check should allow anything outside the range 64-95 if I'm looking at the right part of the RFC.

`rtp = pt < 64 || pt > 95;`? Not sure, but something seems a bit off here.

paulej: I think to check for RTCP, it should be `pt >= 64 && pt <= 95`.

That said, if we're muxing RTCP with RTP, there might also be DTLS or STUN packets. Should we also check the first octet in the packet?  In addition to the PT value, the first octet of the packet should be `octet1 >= 128 && octet1 <= 191`.  If that first octet isn't in that range, it's neither RTP nor RTCP, so the packet should be skipped over entirely.

kyzer-davis: > If that first octet isn't in that range, it's neither RTP nor RTCP, so the packet should be skipped over entirely.

That makes sense to me, the packet captures I have been testing with have STUN packets along with the RTP+RTCP MUX.

pabuhler: @thisisG , you are right I will replace it with pt < 64 || pt > 95 .
@paulej , any packet that is not SRTP or SRTCP will simply fail the srtp_unprotect functions provided authentication is enabled. The purpose of this check was to ensure the correct srtp_unprotect function was called. Providing any other RTP or RTCP validation is possible but maybe outside the scope of this change?

pabuhler: @kyzer-davis, maybe you would like to confirm it still works for you after latest change?

kyzer-davis: Sure:
Tested a decode of SRTCP+SRTP MUX
Tested a decode of SRTP stream with a few different payload types on the same UDP source port. G711 RTP (0), RTP-NTE DTMF (101), Comfort Noise (13).

They both decoded without errors as per my manual checks.
The new rtp, rtcp, error counters also appear to be in line with my manual checks.

Issue #445, opened 2019-05-20
if statements in rtp_decoder.c:{372,386,406} check wrong field


jkougoulos:
in main() I noticed in the section 
> /* set up the srtp policy and master key */

in lines 372, 386, 406,  if statements use scs.key_size instead of scs.tag_size:
eg.
`   switch (scs.key_size) {`
`                case 128:`
`                    if (scs.key_size == 4) {`

paulej:
That does looks like an error.

pabuhler:
@jkougoulos thanks for reporting this, it should be fixed now.

PR #447, opened 2019-06-03
rtp_decoder fix checking against tag size

pabuhler:
The key size was incorrectly used to determine the
policy function to use. The correct variable to check
against was the tag size.

This fixes issues #445


thisisG:
LGTM!

PR #449, opened 2019-06-13
Make it easy to use lib in VS2012

Jopie64:
We needed to use this lib in a VC2012 project. I figured it would be a good idea to contribute the changes we made to accomplish this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,89 @@
+cmake_minimum_required (VERSION 2.8)
+
+project(srtp2)
+
+set(SOURCES_C
+	srtp/ekt.c
+	srtp/srtp.c
+)
+
+set(CIPHERS_SOURCES_C
+	crypto/cipher/aes.c
+	crypto/cipher/aes_icm.c
+	crypto/cipher/cipher.c
+	crypto/cipher/null_cipher.c
+)
+
+set(HASHES_SOURCES_C
+	crypto/hash/auth.c
+	crypto/hash/hmac.c
+	crypto/hash/null_auth.c
+	crypto/hash/sha1.c
+)
+
+set(KERNEL_SOURCES_C
+	crypto/kernel/alloc.c
+	crypto/kernel/crypto_kernel.c
+	crypto/kernel/err.c
+	crypto/kernel/key.c
+)
+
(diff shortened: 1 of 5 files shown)


pabuhler:
Hi, @Jopie64 thanks for the PR. The C89 changes look good and will likely merge them. When it comes to the cmake support I will just need to ensure that we would like the overhead of maintaining an extra build system. If the initial purpose is to replace the current visual studio projects then I think it is a good idea. @JonathanLennox  & @nils-ohlmeier, you two have made the most recent changes to the visual studio support, what do you think?

nils-ohlmeier:
I have only made changes to the visual studio project files to keep it working after/when merging the changes from Mozilla.
When it comes to choosing a build system Mozilla is building libsrtp with it's own build system any way. So I would vote for what ever is easiest to maintain from the perspective of the project. Cmake is a reasonable choice, but might not be a good fit if nobody in the project can maintain it.

Jopie64:
If it's maintaining like, adding and removing some files, then I think by looking at the CMakeLists.txt file you can see that this is not too hard. Even for someone who never heard of it. Also you don't need Visual Studio/Windows to maintain it.
Maybe this could weigh in the decision whether or not to replace the VS files with it.

pabuhler:
I think like @nils-ohlmeier  said most people have there own build system for libSRTP when integrating into  project. Replacing the current visual studio project files with a cmake solution sounds like a good idea to me. When can just add it for now and then remove the visual studio project files after the next release.

CMakeLists.txt
```diff
@@ -0,0 +1,89 @@
...
+	${HASHES_SOURCES_C}
+	${KERNEL_SOURCES_C}
+	${MATH_SOURCES_C}
+	${REPLAY_SOURCES_C}
+	${SOURCES_H}
+)
+
+target_include_directories(srtp2 PUBLIC crypto/include include)
+configure_file(config.hw ${CMAKE_CURRENT_SOURCE_DIR}/crypto/include/config.h COPYONLY)
+add_definitions(-D_LIB -DHAVE_CONFIG_H)
```
pabuhler: This is a very simple cmake file with no options. Will need to expand it to support at least openssl. But that can be done after.

Issue #450, opened 2019-06-26
replace visual studio project files with cmake


pabuhler:
Use and extend the cmake supported added in #449 to allow generation of visual studio project files, or other build systems.

The CMakeLists.txt files will need to be extended to support openssl and building of some test applications.

PR #455, opened 2019-07-08
cmake support

pabuhler:
This extends the cmake support added in #449.
It now enables cmake to be used cross platform and supports openssl and debug config options.
The idea is to support cmake as a complete alternative build system, replacing the checked in windows project files (#450).

cmake builds have been added to travis, including on windows. The ENABLE_OPENSSL option is also supported on windows but I did not manage to get travis to install openssl.

I am not a cmake expert so I am sure there are better ways of handling things but this is a start.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ matrix:
         - make
         - make runtest
         - make runtest-valgrind
+        - make distclean
         - mkdir build && cd build
         - cmake ..
         - make
@@ -46,6 +47,7 @@ matrix:
         - make
         - make runtest
         - make runtest-valgrind
+        - make distclean
         - mkdir build && cd build
         - cmake -DENABLE_OPENSSL=ON ..
         - make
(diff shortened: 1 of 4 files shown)


nils-ohlmeier:
LGTM as well. I might also be able to add similar support for NSS as well later.

pabuhler:
@nils-ohlmeier did not mean to offend anyone with skipping nss support initially :) just wanted to check that I could getting openssl working first. If you feel like adding nss support after that would be great!

kyleung-broadsoft:
It worked fine on Windows with OPENSSL enabled and custom OPENSSL directory to generate Visual Studio solution file and nmake makefile.

nils-ohlmeier:
> @nils-ohlmeier did not mean to offend anyone with skipping nss support initially :) just wanted to check that I could getting openssl working first. If you feel like adding nss support after that would be great!

No offense taken. Since the Firefox builds uses it's own build system it's totally Ok to merge this without NSS support. And I can look into adding NSS later in a follow up PR.

PR #456, opened 2019-07-10
vc 2010 support

pabuhler:
This is an alternative to PR #434 given that cmake is the preferred way of building on windows now (#450).

### DIFF
--- a/config_in_cmake.h
+++ b/config_in_cmake.h
@@ -103,3 +103,8 @@
 /* The size of `unsigned long long', as computed by sizeof. */
 @SIZEOF_UNSIGNED_LONG_LONG_CODE@
 
+/* Define to nothing, since the MSVC compiler doesn't support it.  */
+#cmakedefine HAVE_INLINE 1
+#ifndef HAVE_INLINE
+#define inline
+#endif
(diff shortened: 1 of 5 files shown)


config_in_cmake.h
```diff
@@ -103,3 +103,8 @@
 /* The size of `unsigned long long', as computed by sizeof. */
 @SIZEOF_UNSIGNED_LONG_LONG_CODE@
 
+/* Define to nothing, since the MSVC compiler doesn't support it.  */
```
jacgoudsmit: "the MSVC compiler doesn't support [inline]"

This remark is incorrect. The Microsoft Visual Studio compiler has supported the "__inline" keyword to compile C functions inline, since at least Visual Studio 6.0 (1997), possibly longer. Note the double-underscore prefix that's necessary in C source code; the "inline" keyword without underscores is only valid in C++.

Whether this has any influence on other code, I can't tell you. I'm not familiar enough with cmake or the libsrtp source tree at this time.

pabuhler: You are right, this is not exactly true. I had simply copied this from the config.hw file and since inline is supported in newer version of msvc then the cmake check would enabled it when supported and there would be no change for older versions. I now see that config.h_win32v7 defined inline as __inline. I will add a second check for __inline if inline is not supported and then use that. Will be good to get ride of these different hard coded config files.

PR #459, opened 2019-07-23
Add missing symbols to export into srtp.def

xhaakon:
srtp_protect_mki
srtp_unprotect_mki
srtp_protect_rtcp_mki
srtp_unprotect_rtcp_mki
srtp_set_stream_roc
srtp_get_stream_roc

### DIFF
--- a/srtp.def
+++ b/srtp.def
@@ -2,7 +2,9 @@ EXPORTS
 srtp_init
 srtp_shutdown
 srtp_protect
+srtp_protect_mki
 srtp_unprotect
+srtp_unprotect_mki
 srtp_create
 srtp_add_stream
 srtp_remove_stream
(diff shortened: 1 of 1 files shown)


pabuhler:
Hi, thanks for the PR. Quick question, are you using CMake on windows or the checked in project files?

I will test building a dll with the CMake support and see if it works as expect and maybe add other missing functions to the def file.

xhaakon:
Hi,

> are you using CMake on windows or the checked in project files?

Actually, neither of them. We use custom Meson files (available [here](https://github.com/xhaakon/mesonwraps/tree/master/libsrtp2)) so that srtp build integrates well with the rest of our project based on GStreamer.

The missing functions quickly became apparent to us because Meson by default exports only the symbols it finds in the .def file. So, your mileage may vary when building with CMake.

pabuhler:
I have added some more symbols to the def file and verified it builds with cmake.

PR #489, opened 2020-04-09
Update README.md

malwaremuser:
I cringed when I saw your UNIX command line in the function randhex() (line 390) -- just fixed it up.

Granted, it is now less efficient as a result of starting multiple commands in place of awk(1), but given you're reading from /dev/random I suspect your efficiency is shot anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ can use the following bash function to format the output of
 
 ~~~.txt
 function randhex() {
-   cat /dev/random | od --read-bytes=32 --width=32 -x | awk '{ print $2 $3 $4 $5 $6 $7 $8 $9 $10 $11 $12 $13 $14 $15 $16 }'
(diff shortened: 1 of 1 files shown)


paulej:
Why do you consider that better than awk?  If it's the individual field names, we we do this:

```bash
function randhex() {
   cat /dev/random | od --read-bytes=32 --width=32 -x | awk '{ for (i = 2; i <= 16; i++) { printf "%s", $i } printf "\n" }'
```

malwaremuser:
Yeah, that looks better -- it was more the long list of field names, which that _for_ command fixes, and without starting multiple external commands. Sorry, daft idea -- scrap it.

paulej:
How do you want to resolve this?

malwaremuser:
Hi Paul -- apologies for the delay. I was just trawling through github looking for something and remembered this.
Yeah, just kill it. I'll close it.

PR #491, opened 2020-05-18
Fix two-byte RTP header extension encryption

### LINKED ISSUE #490: Two-byte RTP header extension encryption does not work
Hi folks, I'm confused by the logical and with `0x1fff`. In order to just ignore the last 4 (application) bits, following [RFC 8285, sec 4.3](https://tools.ietf.org/html/rfc8285#section-4.3), shouldn't this logical and with `0xfff0`? Or am I missing something?

https://github.com/cisco/libsrtp/blob/90b3124060ef8ba257cbd933a5a0b96dbdcf1efa/srtp/srtp.c#L1426

lgrahl:
Also ignores the application bits as required by RFC 8285, sec 4.3
Fixes #490

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -1423,7 +1423,7 @@ static srtp_err_status_t srtp_process_header_encryption(
                 xtn_hdr_data++;
             }
         }
-    } else if ((ntohs(xtn_hdr->profile_specific) & 0x1fff) == 0x100) {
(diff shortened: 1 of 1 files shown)


lgrahl:
Ping. Anything remaining that needs addressing?

paulej:
Nope. Just need to hit merge.  @pabuhler Any objection?

PR #492, opened 2020-05-28
avoid non-existing compiler flag -O4

traud:


traud:
That change was introduced because of the compiler Clang. However, even GCC does not know that optimization level, see [Gentoo docs](https://wiki.gentoo.org/wiki/GCC_optimization#What_about_-O_levels_higher_than_3.3F) and `gcc -Q -O4 --help=optimizers`. GCC accepts `-O4` but turns that into level 3 internally. Minor nit because it does not make any difference. However, not to confuse other programmers, libSRTP should not use a then lowered optimization flag.

Nevertheless, I guess, the libSRTP team was about getting the fastest code. Therefore, the above proposal with `-Ofast`. An alternative would be to remove that check completely and directly use `-O3`.

pabuhler:
@itollefsen, would you care to comment on this?

itollefsen:
Using `-O4` with Clang used to turn on Link TIme Optimization (LTO), but they stopped that around version 3.5 I think. So if you want LTO these days, you'll have to add explicit flags for it.

I'm weary of `-Ofast` because it allows GCC to drop strict standard compliance. In particular, it turns on  `-ffast-math`. I don't think that's used today (I can't find it at least), so turning that on could potentially break something.

> An alternative would be to remove that check completely and directly use -O3.

I think this makes more sense, but don't drop the check. The check is there because not all compilers supports or understands GCC's command line options. Just leave `-O3` as the only option in the list.

And then drop `-fexpensive-optimizations` from the section below. That's turned on at `-O2` and higher anyway.

Note for later: Someone(TM) should look into whether `-funroll-loops` has a positive impact. It may just as well be the opposite. I'm guessing someone profiled it back when it was added, but wondering if it still provides any gain.

PR #495, opened 2020-06-09
Add support for the Meson build system

nirbheek:
Hi! I'm a [GStreamer](gstreamer.freedesktop.org/) developer, and [for the same reasons as OpenH264](https://github.com/cisco/openh264/pull/2955), we've written Meson build files for libsrtp. We use libsrtp in gstreamer for WebRTC DTLS.

The Meson build system is being adopted by projects such as GNOME, GTK, Mesa, GStreamer, etc. I am not sure if such a pull request is welcome, but it would be amazing if you could consider merging this :)

----
These Meson build files have feature-parity with the Autoconf/Make build files, the Visual Studio solution, and the CMake build files. They are also simpler, faster, and support all target platforms supported by libsrtp (that I know of). The following have been tested:

Linux, macOS, Windows (MinGW), Windows (MSVC), Windows (UWP/WinRT), Cross-Android (all arches), Cross-iOS (all arches), Cross-Linux, Cross-Windows (MinGW).

Meson supports a superset of the currently-supported target platforms, and no platform-specific code should be needed for those, so this also improves cross-platform support.

The build outputs should be ABI-compatible with the outputs by Autoconf, but this has not been verified in detail.

I've also updated the README, and (pasted from there) the steps to try this out are:

```
# Setup the build subdirectory
meson setup --prefix=/path/to/prefix builddir
# Build the project
meson compile -C builddir
# Run tests
meson test -C builddir
# Optionally, install
meson install -C builddir
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -263,7 +263,8 @@ matrix:
     - os: windows
       before_install:
         - choco install python --version 3.8.0
-        - pip3 install meson ninja
+        - choco install ninja
+        - pip3 install meson
       env:
         - TEST="windows"
         - PATH=/c/Python38:/c/Python38/Scripts:$PATH
--- a/.travis.yml
+++ b/.travis.yml
@@ -284,10 +284,12 @@ matrix:
         - meson compile -C builddir-gcc
         - meson test -v -C builddir-gcc
         # Test MSVC 64-bit build
-        - cmd.exe //C 'C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvarsall.bat' x86_amd64 \& set CC=cl \& meson setup --werror builddir-msvc \& meson compile -C builddir-msvc \& meson test -v -C builddir-msvc
+        - cmd.exe //C 'C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvarsall.bat' amd64 \& set CC=cl \& meson setup --werror builddir-msvc-amd64 \& meson compile -C builddir-msvc-amd64 \& meson test -v -C builddir-msvc-amd64
+        # Test MSVC 32-bit build
+        - cmd.exe //C 'C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvarsall.bat' x86 \& set CC=cl \& meson setup --werror builddir-msvc-x86 \& meson compile -C builddir-msvc-x86 \& meson test -v -C builddir-msvc-x86
         # Test MSVC 64-bit UWP build. This is a cross build because we cannot run UWP binaries natively.
         - |
-          cat > uwp-cross-file.txt <<EOF
+          cat > uwp-amd64-cross-file.txt <<EOF
           [host_machine]
           system = 'windows'
           cpu_family = 'x86_64'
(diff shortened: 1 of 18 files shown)


pabuhler:
Hi, I have nothing really against including meson besides the potential maintenance over head. There has been no real discussion on replacing autotools and the current cmake support was seen both as a way to replace hard coded windows project files and to provide a "cross platform" solution to reduce maintenance. If this was to be accepted and you would like to maintain some hope that it does not get broken then I would suggest adding at least a standard linux build to the travis build file.
If meson is gaining traction and it helps (more than one) other projects to have it here then I am ok with it. Lets see what others think?

nirbheek:
Thanks for looking at this @pabuhler :)

Sorry for not being clear, my intention is not to advocate for Meson to replace any of the build systems used by libsrtp. I'm an outsider to the project after all.

As for maintenance overhead, I'd be happy to maintain these build files. I can add a [`CODEOWNERS` file](https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax) with myself in it and I'll get email about all PRs that touch the build files. We're going to be using this in GStreamer, so we will be doing active maintenance too.

I'll also add some Travis CI builds, thanks for the suggestion!

nirbheek:
A datapoint in favour of the idea that this would help more than one project is that there's also a set of meson build files in [Meson's WrapDB](https://wrapdb.mesonbuild.com/libsrtp2) (code: https://github.com/mesonbuild/libsrtp2/tree/2.2.0) written independently by @xhaakon. 

@xhaakon would you also like to be in CODEOWNERS for the meson build files in libsrtp if this is merged?

nirbheek:
> I'll also add some Travis CI builds, thanks for the suggestion!

I've implemented Meson counterparts for all CI build jobs, and I added a new one specifically for testing the build on WinRT/UWP too. Tests seem to be passing everywhere they could be run.

CI now takes ~1-2min more than it used to because the big-endian mips test is the bottleneck and we run tests using Meson there too now.

pabuhler:
@nirbheek  thanks for adding the builds to travis, that would have taken me a bit of time to figure out. I think my position has not changed, this is fine for me as long as there is not to much maintenance and if that becomes an issue we can just bin it later. Will wait abit to see if any other people have comments.
@havardgraff I'll take your advice in to consideration ;-)

nirbheek:
@tp-m pointed out to me that it would be good to add a `make distcheck` target that checks whether the project version in the meson build files is correct. I'm not sure if I did it correctly, but I've added a rough check to the `make distribution` target with instructions on how to update it. It looks like this:

```
===================================================
Meson project version is 2.3.0 which is incorrect.
Please edit meson.build and change the 'version:'
field in the project() call to 2.4.0
===================================================
```

The `project()` call is the first line in the build file, so should be easy to find.

xhaakon:
> @xhaakon would you also like to be in CODEOWNERS for the meson build files in libsrtp if this is merged?

Nothing against that, feel free to add me.

nirbheek:
> Nothing against that, feel free to add me.

Done, thanks!

Also fixed a build failure on msvc-32-bit and added CI for windows UWP arm64.

nirbheek:
Gentle poke :)

pabuhler:
@nirbheek , sorry for the delay, have been out of the office for a few weeks. Will follow up now.

pabuhler:
@nirbheek  for me supporting meson seams non trivial as with all build systems there is lots to learn and when adapting to an existing projects there are always special case handling that require even more work. So we can see how this goes and if it does not cause to much work then great but if it becomes a problem we will have to review it.

nirbheek:
Many thanks for merging this 😄 I hope it works out for all of us. Please poke me when you have questions. Either on GitHub, or on FreeNode (`nirbheek`). I am on `#gstreamer` and `#mesonbuild`. And of course, since we're all in the CODEOWNERS file, we will get email for PRs to the meson build files.

Issue #479, opened 2020-01-12
autoconf cross-compile fails in 2.3.0 with PR #414


abelbeck:
Using release 2.3.0 containing the PR #414 change, the autoconf test hard fails when cross compiling ...
```
checking if OPENSSL_cleanse is broken... configure: error: in `/home/dev/astlinux/trunk/output/build/libsrtp-2.3.0':
configure: error: cannot run test program while cross compiling
See `config.log' for more details
```
There is no assumed case while cross compiling.

disa6302:
Is this going to be merged in anytime soon? 
I am running into this issue when I run configure on ARM platform.

disa6302:
Also, if this is not going to be merged anytime soon, is there a workaround?

abelbeck:
@disa6302 : The solution is to revert PR #414 ... this issue documents the problem.

disa6302:
Got it. Would the revert be made part of the release anytime soon?

abelbeck:
I do not have any permissions here, just a project that uses it, but it seems to me the OpenSSL version the PR #414 fixes is deprecated and no longer supported. So rather than fixing #414 it may be best to revert it at this point in time.

pabuhler:
Hi, sorry for not following up sooner, will review again now an give some feedback.

pabuhler:
This should now be merged. reopen if you still have problems

PR #496, opened 2020-06-15
Rework check for OPENSSL_cleanse()

ydroneaud:
Don't realize a runtime check for OPENSSL_cleanse() when not needed (newer OpenSSL versions), or not possible (when cross compiling, see #479).


pabuhler:
@itollefsen what do you think?  Seams to look ok to me.

pabuhler (approved):
verified behavior of config checks

ydroneaud:
Thanks

Issue #497, opened 2020-06-24
Potential inconsistencies in internal error handling


mfroeschl:
Dear libsrtp team,

I have noticed the following potential inconsistencies when handling `srtp_err_status_t` results of functions (or macros) such as `srtp_auth_update()` :

For example, in the function `srtp_protect_mki()`, the result of `srtp_auth_update()` is checked and in case of error the function aborts:

```
        /* run auth func over packet */
        status = srtp_auth_update(session_keys->rtp_auth, (uint8_t *)auth_start,
                                  *pkt_octet_len);
        if (status)
            return status;
```

([srtp.c#L2335](https://github.com/cisco/libsrtp/blob/72c7541b3eb3742fe6cae7f85d2a1b02bd19e03c/srtp/srtp.c#L2335))

In the function `srtp_unprotect_mki()` however, the result of `srtp_auth_update()` is not checked and in case of error the function continues:

```
        /* now compute auth function over packet */
        status = srtp_auth_update(session_keys->rtp_auth, (uint8_t *)auth_start,
                                  *pkt_octet_len - tag_len - mki_size);

        No check?

        /* run auth func over ROC, then write tmp tag */
        status = srtp_auth_compute(session_keys->rtp_auth, (uint8_t *)&est, 4,
                                   tmp_tag);
```

([srtp.c#L2605](https://github.com/cisco/libsrtp/blob/72c7541b3eb3742fe6cae7f85d2a1b02bd19e03c/srtp/srtp.c#L2605))

Can you please evaluate if this is intentional, or an oversight?
If it is indeed intentional, please kindly consider making this more obvious, e.g. by adding a comment.

Below is a list of similar locations:

- `srtp.c`: return value of `srtp_auth_start()` is sometimes checked, sometimes ignored
- `aes_gcm_ossl.c`: return value of `EVP_CIPHER_CTX_ctrl()` is sometimes checked, sometimes ignored
- `aes_gcm_ossl.c`: return value of `EVP_Cipher()` is sometimes checked, sometimes ignored
- `hmac_ossl.c`: return value of `HMAC_CTX_init()` is ignored
    - (Also, `HMAC_CTX_init()` seems to be deprecated in OpenSSL 1.1)

Thank you so much for your time!

pabuhler:
I think it is an oversight. For the unprotect function, the the function should fail regardless as it will compare the output of auth update with the pre calculated value, but for consistency and correctness I think we should add the checks.
The return from other calls should also be checked.

[...]

PR #498, opened 2020-07-07
Feature/additional error checks

mfroeschl:
Add additional error checks to `srtp.c` and `aes_gcm_ossl.c` as outlined in #497

### DIFF
--- a/crypto/cipher/aes_gcm_ossl.c
+++ b/crypto/cipher/aes_gcm_ossl.c
@@ -263,7 +263,10 @@ static srtp_err_status_t srtp_aes_gcm_openssl_set_aad(void *cv,
      */
     unsigned char dummy_tag[GCM_AUTH_TAG_LEN];
     memset(dummy_tag, 0x0, GCM_AUTH_TAG_LEN);
-    EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len, &dummy_tag);
+    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len,
+                             &dummy_tag)) {
+        return (srtp_err_status_algo_fail);
+    }
 
     rv = EVP_Cipher(c->ctx, NULL, aad, aad_len);
     if (rv != aad_len) {
@@ -322,7 +325,9 @@ static srtp_err_status_t srtp_aes_gcm_openssl_get_tag(void *cv,
     /*
      * Retreive the tag
      */
-    EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_GET_TAG, c->tag_len, buf);
+    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_GET_TAG, c->tag_len, buf)) {
+        return (srtp_err_status_algo_fail);
+    }
 
     /*
      * Increase encryption length by desired tag size
@@ -352,8 +357,10 @@ static srtp_err_status_t srtp_aes_gcm_openssl_decrypt(void *cv,
     /*
      * Set the tag before decrypting
      */
-    EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len,
-                        buf + (*enc_len - c->tag_len));
+    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len,
+                             buf + (*enc_len - c->tag_len))) {
+        return (srtp_err_status_auth_fail);
+    }
     EVP_Cipher(c->ctx, buf, buf, *enc_len - c->tag_len);
 
     /*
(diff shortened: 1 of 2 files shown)


crypto/cipher/aes_gcm_ossl.c
```diff
@@ -263,7 +263,10 @@ static srtp_err_status_t srtp_aes_gcm_openssl_set_aad(void *cv,
      */
     unsigned char dummy_tag[GCM_AUTH_TAG_LEN];
     memset(dummy_tag, 0x0, GCM_AUTH_TAG_LEN);
-    EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len, &dummy_tag);
+    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len,
+                             &dummy_tag)) {
```
mfroeschl, line 267: the line break was introduced by clang-format

crypto/cipher/aes_gcm_ossl.c
```diff
@@ -263,7 +263,10 @@ static srtp_err_status_t srtp_aes_gcm_openssl_set_aad(void *cv,
...
+        return (srtp_err_status_algo_fail);
```
mfroeschl, line 268: I chose the same return value as in line 273 below

crypto/cipher/aes_gcm_ossl.c
```diff
@@ -322,7 +325,9 @@ static srtp_err_status_t srtp_aes_gcm_openssl_get_tag(void *cv,
     /*
      * Retreive the tag
      */
-    EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_GET_TAG, c->tag_len, buf);
+    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_GET_TAG, c->tag_len, buf)) {
+        return (srtp_err_status_algo_fail);
```
mfroeschl, line 329: not sure if this is the best return value in this case, please check

crypto/cipher/aes_gcm_ossl.c
```diff
@@ -352,8 +357,10 @@ static srtp_err_status_t srtp_aes_gcm_openssl_decrypt(void *cv,
     /*
      * Set the tag before decrypting
      */
-    EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len,
-                        buf + (*enc_len - c->tag_len));
+    if (!EVP_CIPHER_CTX_ctrl(c->ctx, EVP_CTRL_GCM_SET_TAG, c->tag_len,
+                             buf + (*enc_len - c->tag_len))) {
+        return (srtp_err_status_auth_fail);
```
mfroeschl, line 362: I chose the same return value as in line 370 below

srtp/srtp.c
```diff
@@ -2604,6 +2604,8 @@ srtp_err_status_t srtp_unprotect_mki(srtp_ctx_t *ctx,
         /* now compute auth function over packet */
         status = srtp_auth_update(session_keys->rtp_auth, (uint8_t *)auth_start,
                                   *pkt_octet_len - tag_len - mki_size);
+        if (status)
+            return status;
```
mfroeschl, line 2608: If `srtp_auth_update()` above fails, existing code returns the `status`. But if `srtp_auth_compute()` below fails, existing code returns `srtp_err_status_auth_fail` rather than the original status code. So I was not sure whether to best return `status` or `srtp_err_status_auth_fail` here and arbitrarily chose the former. This applies to all changes in this file. Please tell me if you disagree.

mfroeschl:
Update: It seems adding the checks for `EVP_CIPHER_CTX_ctrl()` return values causes the `cipher_driver.c` test to fail:

https://travis-ci.org/github/cisco/libsrtp/jobs/705650070

```
Build done. Please run 'make runtest' to run self tests.
The command "make" exited with 0.
0.01s$ make runtest
Build done. Please run 'make runtest' to run self tests.
running libsrtp2 test applications...
crypto/test/cipher_driver -v >/dev/null
Makefile:46: recipe for target 'runtest' failed
make: *** [runtest] Error 11
```

`11` is `srtp_err_status_algo_fail`

pabuhler:
Hi, looks good so far. I think setting the dummy_tag with EVP_CTRL_GCM_SET_TAG was needed in older versions of openssl, testing with new version it does not seam to be needed. Regardless EVP_CTRL_GCM_SET_TAG can only be used when decrypting so it should only be used when c->dir == srtp_direction_decrypt, so we can protect the setting of the dummy_tag in an if check. I did a test and it seams to work. Will you make that change or if you prefer I can push that change to your branch.

mfroeschl:
@pabuhler Thank you for the quick response! I could confirm that after implementing your suggestion (only set the dummy tag when decrypting) the tests pass again. I will update my PR accordingly.

mfroeschl:
@pabuhler FYI, I have updated the PR with your suggestions (but forgot to send you a notification).

PR #502, opened 2020-07-29
remove visual studio build files

pabuhler:
This completes the transition from visual studio build
files to the cmake builds system for windows.
see #450


pabuhler:
@JonathanLennox since you have done a lot of the changes on these mscv files it would be good to get you to sign off on this.

PR #503, opened 2020-08-14
Support apple silicon build

pabuhler:
Initial support for building for Apple Silicon with cmake. This uses only internal crypto, no openssl support yet.

The HAVE_X86 flag should probably be removed but requires a bit of config clean up first.


crypto/include/datatypes.h
```diff
@@ -278,7 +278,7 @@ void octet_string_set_to_zero(void *s, size_t len);
 #define be64_to_cpu(x) bswap_64((x))
 #else /* WORDS_BIGENDIAN */
 
-#if defined(__GNUC__) && defined(HAVE_X86)
+#if defined(__GNUC__) && defined(HAVE_X86) || defined(__x86_64__)
```
paulej, line 281: Does && go with GNUC or with the || __x86_64__?  I assume this is intended to be `defined(__GNUC__) && (defined(HAVE_X86) || defined(__x86_64__))?

pabuhler: yip, I can add the parentheses to clarify

PR #504, opened 2020-08-25
temporarily disable meson uwp builds on windows

pabuhler:
Not sure what has changed but travis started failing.


pabuhler:
@nirbheek I am disabling meson uwp builds on windows for now as they started to fail. It was failing to find the ws2_32 library, I am not sure what has changed to trigger this. Feel free to fix and enable it again.

PR #505, opened 2020-08-25
Try to re-enable UWP CI

nirbheek:
Continuing from https://github.com/cisco/libsrtp/pull/504.


nirbheek:
Further investigation showed this was actually https://github.com/mesonbuild/meson/pull/7639. Will be fixed when 0.55.2 is released, and I'll re-enable this test then.

pabuhler:
@nirbheek thanks for following this up.

PR #506, opened 2020-09-10
Re-enable UWP now that Meson 0.55.2 has been released

nirbheek:
This should work now. Let's see.


nirbheek:
Seems to have worked 🙂

pabuhler:
thanks

PR #511, opened 2020-11-05
Add support for cryptex 

murillo128:
Implementation of cryptex as per https://github.com/juberti/cryptex

Discussion:
https://github.com/juberti/cryptex/issues/5


srtp/srtp.c
```diff
@@ -1727,10 +1731,29 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
...
+    }
     if (hdr->x == 1) {
-        xtn_hdr = (srtp_hdr_xtnd_t *)enc_start;
-        enc_start += (ntohs(xtn_hdr->length) + 1);
+        xtn_hdr = (srtp_hdr_xtnd_t *)(uint32_t *)hdr + uint32s_in_rtp_header +
+                  hdr->cc;
+        xtn_hdr_length = ntohs(xtn_hdr->length);
+        xtn_profile_specific = ntohs(xtn_hdr->profile_specific);
+    }
+    /* If no header extension is present cyrptex has no effect */
```
dbrgn, line 1745: Only a nitpick that doesn't add much to your efforts (I haven't read the spec yet), but here's a typo (cyrptex -> cryptex). Another one is farther below (line 2259).

murillo128: good catch! thanxs!

pabuhler:
hi, it would be good to get the CI build passing as soon as possible, are you able to fix that or doe you need some help?

srtp/srtp.c
```diff
@@ -1727,10 +1731,29 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
...
+        uint32_t *csrcs = (uint32_t *)hdr + uint32s_in_rtp_header;
+        /* Move CSRCS so it is contiguos with extension header block */
+        for (unsigned char i = 0; i < hdr->cc; ++i)
+            csrcs[i + 1] = csrcs[i];
+        enc_start = csrcs + 1;
+    } else {
+        enc_start = (uint32_t *)hdr + uint32s_in_rtp_header + hdr->cc;
+        if (hdr->x == 1) {
+            enc_start += (xtn_hdr_length + 1);
+        }
```
JonathanLennox, line 1756: Am I missing something, or does this not do AAD over the header extension header (the "defined by profile" and length fields) if CSRCs are present?  I think the first CSRC ends up being put in the AAD instead.

srtp/srtp.c
```diff
@@ -1812,6 +1835,23 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
     if (status) {
         return srtp_err_status_cipher_fail;
     }
+
+    if (stream->use_cryptex && xtn_hdr) {
+        uint32_t *csrcs = (uint32_t *)hdr + uint32s_in_rtp_header;
+        /* Restore CSRCS to its original position */
+        for (unsigned char i = 0; i < hdr->cc; ++i)
+            csrcs[i] = csrcs[i + 1];
+        /* Restore extension header and change profiles by crytex values*/
```
JonathanLennox, line 1844: Typo, "cryptex"

srtp/srtp.c
```diff
@@ -2169,10 +2246,28 @@ srtp_err_status_t srtp_protect_mki(srtp_ctx_t *ctx,
...
-            enc_start += (ntohs(xtn_hdr->length) + 1);
+            xtn_hdr = (srtp_hdr_xtnd_t *)(uint32_t *)hdr +
+                      uint32s_in_rtp_header + hdr->cc;
+            xtn_hdr_length = ntohs(xtn_hdr->length);
+            xtn_profile_specific = ntohs(xtn_hdr->profile_specific);
+        }
+        /* If no header extension is present cryptex has no effect */
+        if (stream->use_cryptex && hdr->x) {
+            uint32_t *csrcs = (uint32_t *)hdr + uint32s_in_rtp_header;
+            /* Move CSRCS so it is contiguos with extension header block */
```
JonathanLennox, line 2262: Typo, "contiguous"

srtp/srtp.c
```diff
@@ -1959,6 +2018,22 @@ static srtp_err_status_t srtp_unprotect_aead(srtp_ctx_t *ctx,
         return status;
     }
 
+    if (use_cryptex) {
+        uint32_t *csrcs = (uint32_t *)hdr + uint32s_in_rtp_header;
+        /* Restore CSRCS to its original position */
+        for (unsigned char i = 0; i < hdr->cc; ++i)
+            csrcs[i] = csrcs[i + 1];
+        /* Restore extension header and change profiles by crytex values*/
```
JonathanLennox, line 2026: Typo, "cryptex"

srtp/srtp.c
murillo128: Yes, that is incorrect. First CSRC should be overwritten by "defined by profile" and length fields so they are included on the AAD, right?

JonathanLennox: Yes, and then that field should be put back in its proper place when the CSRCs are moved back.

murillo128: Moving back the CSRCs and restoring the defined by profile + len is done after encryption below. I will have to replace the 0xBEDE and 0x100X by the crytpex  one here so they correct values are used in the authentication. Currently it is done after authentication.

murillo128: should be fixed now

fluffy:
Great to have this on a branch but I would like to see the IETF draft to get further along before we merge it to the main branch.

murillo128:
@fluffy I agree. My main intent was to get feedback on the changes and spot any issue on the implementation that could impact the spec.

Would be great if it could be thoroughly reviewed so we can confirm that the test vectors can be incorporated into the rfc and be ready to merge the branch as soon as the rfc is ready.

fluffy:
Given the IETF draft is about at WGLC, we should merge this in once we are happy with the tests.

pabuhler:
> Given the IETF draft is about at WGLC, we should merge this in once we are happy with the tests.

@murillo128  are you able to fix the compile issues? Maybe rebase on master as well?

@paulej  have you thought to review these changes ?

murillo128:
I'll do it later this month when I am back from vacations

paulej (approved):
Looks good. I appreciate that there is test code to go with the implementation, because I'll admit I didn't check every field to ensure the math is right.

srtp/srtp.c
```diff
@@ -2169,10 +2261,42 @@ srtp_err_status_t srtp_protect_mki(srtp_ctx_t *ctx,
...
+            } else if (xtn_profile_specific == 0x1000) {
+                xtn_hdr_profile_and_value =
+                    htonl(0xc2de << 16 | xtn_hdr_length);
+            } else {
+                return srtp_err_status_parse_err;
+            }
+            /* Get CSRCs block position or profile if no CSRCs */
+            uint32_t *csrcs = (uint32_t *)hdr + uint32s_in_rtp_header;
+            /* Move CSRCS so block is contiguous with extension header block */
+            for (unsigned char i = hdr->cc; i > 0; --i)
```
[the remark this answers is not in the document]
murillo128: just thought it would be easier to review the change and check it was spec compliant, I can change it to memcpy or memmove if needed

murillo128:
I have also rebased to master and fixed the CI errors of the previous version. Could anyone run the github workflows to check if everything is correct now?

murillo128:
fixed format and mem leak on deallocating recv sessions on tests

fluffy:
@bifurcation - Did you ever ger to look at this ?

bifurcation:
Yes, that's what led to this: https://github.com/juberti/cryptex/issues/31

hakarim740-com-ra:
When will this be merged?

pabuhler:
hmm,  I see this made it to an RFC now, https://datatracker.ietf.org/doc/html/rfc9335 , this was not the case when this code was committed or originally reviewed. I am unsure if the current implementation follows the RFC or not, would need some input from @murillo128 on that.
Either way this code needs to be updated before it can be merge.
Do you have a use for this code?

hakarim740-com-ra:
Mainly to be used with WebRTC

srtp/srtp.c
```diff
@@ -1723,10 +1728,41 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
...
+                  hdr->cc;
+        xtn_hdr_length = ntohs(xtn_hdr->length);
+        xtn_profile_specific = ntohs(xtn_hdr->profile_specific);
+    }
+    /* If no header extension is present cryptex has no effect */
+    if (stream->use_cryptex && hdr->x) {
+        /* Change profiles by cryptex values */
+        if (xtn_profile_specific == 0xbede) {
+            xtn_hdr_profile_and_value = htonl(0xc0de << 16 | xtn_hdr_length);
+        } else if (xtn_profile_specific == 0x1000) {
```
bifurcation, line 1747: This is incorrect.  The low-order bits don't have to be zero.

cf. https://github.com/cisco/libsrtp/blob/main/srtp/srtp.c#L1638

srtp/srtp.c
```diff
@@ -1723,10 +1728,41 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
...
+            xtn_hdr_profile_and_value = htonl(0xc2de << 16 | xtn_hdr_length);
+        } else {
+            return srtp_err_status_parse_err;
+        }
+        /* Get CSRCs block position or profile if no CSRCs */
+        uint32_t *csrcs = (uint32_t *)hdr + uint32s_in_rtp_header;
+        /* Move CSRCS so block is contiguous with extension header block */
+        for (unsigned char i = hdr->cc; i > 0; --i)
+            csrcs[i] = csrcs[i - 1];
```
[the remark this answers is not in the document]
bifurcation: Or even better, `memmove`.

bifurcation (commented):
I would hold on merging this until a few more folks chime in saying they'll use it.  Otherwise it's unnecessary complexity.

murillo128:
If this gets merged and upstreamed by libwebrtc  I will provide a patch for adding support to  cryptex in libwebrtc

pabuhler:
ok, I can update it to work on top of the v3 work if you like, it will give me a chance to better review the code.

include/srtp.h
```diff
@@ -351,6 +351,8 @@ typedef struct srtp_policy_t {
     uint8_t *enc_xtn_hdr;       /**< List of header ids to encrypt.      */
     size_t enc_xtn_hdr_count;   /**< Number of entries in list of header */
                                 /**<  ids.                               */
+    unsigned int use_cryptex;   /**< Encrypt header block and CSRCS with */
```
murillo128, line 354:
```suggestion
    bool use_cryptex;   /**< Encrypt header block and CSRCS with */
```

include/srtp_priv.h
```diff
@@ -125,6 +125,8 @@ typedef struct srtp_stream_ctx_t_ {
     uint8_t *enc_xtn_hdr;
     size_t enc_xtn_hdr_count;
     uint32_t pending_roc;
+    unsigned int use_cryptex;
```
murillo128, line 128:
```suggestion
    bool use_cryptex;
```

srtp/srtp.c
```diff
@@ -1883,11 +1892,48 @@ static srtp_err_status_t srtp_protect_aead(srtp_ctx_t *ctx,
...
+        xtn_hdr_length = ntohs(xtn_hdr->length);
+        xtn_profile_specific = ntohs(xtn_hdr->profile_specific);
+    }
+  
+    /* If no header extension is present cryptex has no effect */
+    if (stream->use_cryptex && hdr->x) {
+        /* Change profiles by cryptex values */
+        if (xtn_profile_specific == 0xbede) {
+            xtn_hdr_profile_and_value = htonl(0xc0de << 16 | xtn_hdr_length);
+        } else if (xtn_profile_specific == 0x1000) {
```
murillo128, line 1917:
```suggestion
        } else if ((xtn_profile_specific & 0xfff0) == 0x1000) {
```

srtp/srtp.c
```diff
@@ -2019,7 +2086,12 @@ static srtp_err_status_t srtp_unprotect_aead(srtp_ctx_t *ctx,
     v128_t iv;
     srtp_err_status_t status;
     size_t tag_len;
-    size_t aad_len;
+    size_t int aad_len;
+    srtp_hdr_xtnd_t *xtn_hdr = NULL;
+    int xtn_hdr_length = 0;
+    int xtn_profile_specific = 0;
+    unsigned int use_cryptex = 0;
```
murillo128, line 2093:
```suggestion
    bool use_cryptex = false;
```

srtp/srtp.c
```diff
@@ -2019,7 +2086,12 @@ static srtp_err_status_t srtp_unprotect_aead(srtp_ctx_t *ctx,
     v128_t iv;
     srtp_err_status_t status;
     size_t tag_len;
-    size_t aad_len;
+    size_t int aad_len;
+    srtp_hdr_xtnd_t *xtn_hdr = NULL;
+    int xtn_hdr_length = 0;
```
murillo128, line 2091:
```suggestion
    size_t xtn_hdr_length = 0;
```

srtp/srtp.c
```diff
@@ -2213,6 +2333,11 @@ srtp_err_status_t srtp_protect(srtp_t ctx,
     srtp_stream_ctx_t *stream;
     size_t prefix_len;
     srtp_session_keys_t *session_keys = NULL;
+    uint8_t *mki_location = NULL;
+    int advance_packet_index = 0;
+    int xtn_hdr_length = 0;
```
murillo128, line 2338:
```suggestion
    size_t xtn_hdr_length = 0;
```

srtp/srtp.c
```diff
@@ -2524,6 +2705,10 @@ srtp_err_status_t srtp_unprotect(srtp_t ctx,
     bool advance_packet_index = false;
     uint32_t roc_to_set = 0;
     uint16_t seq_to_set = 0;
+    int xtn_hdr_length = 0;
```
murillo128, line 2708:
```suggestion
    size_t xtn_hdr_length = 0;
```

srtp/srtp.c
```diff
@@ -2524,6 +2705,10 @@ srtp_err_status_t srtp_unprotect(srtp_t ctx,
     bool advance_packet_index = false;
     uint32_t roc_to_set = 0;
     uint16_t seq_to_set = 0;
+    size_t xtn_hdr_length = 0;
+    int xtn_profile_specific = 0;
+    unsigned int use_cryptex = 0;
```
murillo128, line 2710:
```suggestion
    bool use_cryptex = 0;
```

murillo128:
i have rebased the PR and addressed the comments. could we run the tests? not completly sure I did everything correctly

JonathanLennox:
If you want to test interop, once you have the WebRTC patch, Jitsi Meet's Videobridge supports cryptex, and I should be able to configure an instance to negotiate it in our WebRTC client.

include/srtp_priv.h
```diff
@@ -125,6 +125,8 @@ typedef struct srtp_stream_ctx_t_ {
     uint8_t *enc_xtn_hdr;
     size_t enc_xtn_hdr_count;
     uint32_t pending_roc;
+    bool use_cryptex;
+    struct srtp_stream_ctx_t_ *next; /* linked list of streams */
```
pabuhler, line 129: next was removed, I guess this slipped in while merging ?

pabuhler: There have been a lot of changes in main, are you sure the merge went well ? All builds are failing.
Also the input buffer to the protect/unprotect functions is now const, not sure you should just cast that away and move the memory around.

murillo128: i think I have been too optimistic about my rebasing 😅

Wil try to get a look next week

pabuhler: Yeah, I had a quick attempt a few weeks ago and understood that this was not an easy merge :( . Let me know if you want to discuss anything.

pabuhler:
@murillo128  I have created a new PR #724 , that tries to merge your changes in to main. It takes a slightly different approach given the non in-place io support that is in main now.
With the new API the input buffer is always const, this means that modifying it in-place to create a continues memory block to encrypt is maybe not feasible. The new code will encrypt the csrc list first before encrypting the header extensions and payload in one block. The code also does not currently support encrypting csrc's in gcm mode as today in gcm mode there can only be one call to encrypt. So something needs to change if cryptex is to be supported.
 The two options I see are:
1. Add support to call encrypt/decrypt multiple times in gcm mode, I think most backends can support this.
2. Change the input buffers from const to non const.

If option 1 is done then it could still be possible to modify the input to supportone call to encrypt when it is detected that this is in-place io.
I prefer not to remove the const from the input buffers but it is something that could be discussed. I am not sure of the performance penalty of call encrypt twice, it would probably depend alot on the backend.

murillo128:
@pabuhler thank you very much for taking the time to work on this!

Removing the `const` is not my preferred option too, would be great if the gcm apis could be updated to allow multiple calls. 

On my experience, some backends (expecially OpenSSL 3.0) have a huge overhead when making multiple calls to the api, but we are already having that performance issue anyway: https://github.com/cisco/libsrtp/issues/645

pabuhler:
> @pabuhler thank you very much for taking the time to work on this!
> 
> Removing the `const` is not my preferred option too, would be great if the gcm apis could be updated to allow multiple calls.
> 
> On my experience, some backends (expecially OpenSSL 3.0) have a huge overhead when making multiple calls to the api, but we are already having that performance issue anyway: #645

OK, I will tryout gcm api supporting multiple calls to verify that it is possible. Then at least cryptex can be "functionaly" complete and can look at optimizations afterwards. It is not sure that the performance problems in #645 will be such and issue when making multiple calls to openssl if IV has not changed but time will tell.

pabuhler:
@murillo128, I have updated the code in #724,  I have abandoned the attempt to support calling encrypt/decrypt multiple times when using a gcm cipher as it is not well support by the different backends. Can look in to this again later.

So the current state is that if the io is in-place and there are CC's in the header then the input buffer will be rearranged similar to your original PR, this is supported by both icm and gcm. If there are CC's in the header and io is not in-place then there will multiple calls to encrypt or decrypt, this is only supported by icm. If there are no CC's in the header then the start offset is just updated and there is a single call to encrypt or decrypt, this is supported by both icm and gcm.

Short summary is if you want to use cryptex then you should use in-place io.

Do you think could merge that code in to this PR and it could get reviewed and hopefully merged?
There are a couple of failing CI builds in #724 which I will follow up on. I do not think they are cryptex specific so will try to reproduce and fix independently.

pabuhler:
@murillo128 have you had a chance to look at my proposal? I am thinking to merge it in to your branch soon so it can be reviewed and tested.
@JonathanLennox I am not sure how to test with WebRTC, but if some one could provide a pcap and a key I could run some tests against that.

pabuhler:
@murillo128  I have force pushed my changes to your branch, overwriting your recent merging with main attempt, I hope that is ok. I have made a local copy of your changes if you would like it reverted.

It would be could to get this reviewed again and then it can be merged. There are some limitations as described earlier but I think it is acceptable for now.

srtp/srtp.c
```diff
@@ -2082,6 +2292,19 @@ static srtp_err_status_t srtp_unprotect_aead(srtp_ctx_t *ctx,
...
+    status = srtp_cryptex_unprotect_init(stream, hdr, srtp, rtp, &cryptex_inuse,
+                                         &cryptex_inplace, &enc_start);
+    if (status) {
+        return status;
+    }
+
+    if (cryptex_inuse && !cryptex_inplace && hdr->cc) {
+        debug_print0(mod_srtp,
+                     "unsupported cryptex mode, AEAD, CC and not inplace io");
+        return srtp_err_status_bad_param;
```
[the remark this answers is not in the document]
JonathanLennox: Oh, I guess the problem is that `srtp_cipher_encrypt` is a one-and-done API, not an incremental API.

This could be changed, but it'd be a big enough change that it probably shouldn't be part of this PR just for non-in-place cryptex-with-CC.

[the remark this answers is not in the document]
pabuhler: srtp_cipher_encrypt API works ok incrementally for normal AES but does not work for AES GCM. I made an attempt to turn it into an incremental API but had problems with the different back ends, some just did not support it or required it to be in AES block size increments :( . Like you say we can try again at some point after this PR so it does not mix things up to much.

srtp/srtp.c
```diff
@@ -121,6 +127,182 @@ static size_t srtp_get_rtp_xtn_hdr_len(const srtp_hdr_t *hdr,
...
+    const uint8_t *rtp,
+    const uint8_t *srtp,
+    bool *inuse,
+    bool *inplace,
+    size_t *enc_start)
+{
+    if (stream->use_cryptex && stream->rtp_services & sec_serv_conf) {
+        if (hdr->cc && hdr->x == 0) {
+            /* Cryptex can only encrypt CSRCS if header extension is present */
+            return srtp_err_status_parse_err;
```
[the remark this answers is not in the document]
pabuhler: Any suggestions?
Something very specific like
srtp_err_status_cryptex_rtp_header_extension_required
or a little more generic
srtp_err_status_cryptex_unsupported

pabuhler:
Hi @JonathanLennox, 
Thank you for your review, I have pushed a commit that should address your comments.

include/srtp.h
```diff
@@ -216,6 +216,7 @@ typedef enum {
     srtp_err_status_pkt_idx_adv = 27,   /**< packet index advanced, reset    */
                                         /**< needed                          */
     srtp_err_status_buffer_small = 28,  /**< out buffer is too small         */
+    srtp_err_status_cryptex_unsupported = 29, /**< out buffer is too small */
```
JonathanLennox, line 219: Comment needs updating

srtp/srtp.c
```diff
@@ -121,6 +127,182 @@ static size_t srtp_get_rtp_xtn_hdr_len(const srtp_hdr_t *hdr,
...
+    }
+
+    if (inplace) {
+        srtp_cryptex_adjust_buffer(hdr, srtp);
+    } else {
+        if (hdr->cc) {
+            uint8_t *cc_list = srtp + octets_in_rtp_header;
+            size_t cc_list_size = hdr->cc * 4;
+            /* CSRCs are in dst header already, enc in place */
+            srtp_err_status_t status = srtp_cipher_encrypt(
```
paulej, line 216: Maybe we should talk about this, but it's not clear to me how this works. The "inplace" version, which is used with AEAD, makes sense since it moves the CSRCs to form the plaintext as described in 6.2/RFC9335. However, this seems to encrypt the CC list separately. I'm confused.

srtp/srtp.c
```diff
@@ -121,6 +127,182 @@ static size_t srtp_get_rtp_xtn_hdr_len(const srtp_hdr_t *hdr,
...
+                                                srtp_cipher_t *rtp_cipher)
+{
+    if (inplace) {
+        srtp_cryptex_adjust_buffer(hdr, rtp);
+    } else {
+        if (hdr->cc) {
+            uint8_t *cc_list = rtp + octets_in_rtp_header;
+            size_t cc_list_size = hdr->cc * 4;
+            /* CSRCs are in dst header already, enc in place */
+            srtp_err_status_t status = srtp_cipher_decrypt(
```
paulej, line 278: Same point of confusion here.

srtp/srtp.c
```diff
@@ -121,6 +127,182 @@ static size_t srtp_get_rtp_xtn_hdr_len(const srtp_hdr_t *hdr,
...
+    const srtp_hdr_t *hdr,
+    const uint8_t *rtp,
+    const uint8_t *srtp,
+    bool *inuse,
+    bool *inplace,
+    size_t *enc_start)
+{
+    if (stream->use_cryptex && (stream->rtp_services & sec_serv_conf)) {
+        if (hdr->cc && hdr->x == 0) {
+            /* Cryptex can only encrypt CSRCS if header extension is present */
```
paulej, line 173: s/CSRCS/CSRCs/

include/srtp.h
```diff
@@ -216,6 +216,7 @@ typedef enum {
     srtp_err_status_pkt_idx_adv = 27,   /**< packet index advanced, reset    */
                                         /**< needed                          */
     srtp_err_status_buffer_small = 28,  /**< out buffer is too small         */
+    srtp_err_status_cryptex_err = 29,   /**< unsupported cryptex operation   */
```
paulej, line 219: Rather than adding this error, why not just disable extension encryption of extensions are not present?

srtp/srtp.c
paulej: Talked with @pabuhler. I now understand what I was missing.

include/srtp.h
pabuhler: Want to avoid people thinking the csrc's will be protected but are actually not as they did not include a header extension, unless I misunderstood you comment.

srtp/srtp.c
```diff
@@ -121,6 +127,182 @@ static size_t srtp_get_rtp_xtn_hdr_len(const srtp_hdr_t *hdr,
...
+        (const srtp_hdr_xtnd_t *)(rtp + srtp_get_rtp_hdr_len(hdr));
+    return ntohs(xtn_hdr->profile_specific);
+}
+
+static void srtp_cryptex_adjust_buffer(const srtp_hdr_t *hdr, uint8_t *rtp)
+{
+    uint8_t tmp[4];
+    uint8_t *ptr = rtp + srtp_get_rtp_hdr_len(hdr);
+    memcpy(tmp, ptr, 4);
+    for (size_t i = hdr->cc; i > 0; --i) {
```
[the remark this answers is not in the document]
pabuhler: I am not sure why the original code was change to not use memmove() will change back

include/srtp.h
paulej: OK, that's a valid argument. If the only reason for the error is that extension encryption is requested, but extensions were not given, perhaps that should be in the description. Like, "extensions required for cryptex".

paulej (approved):

lminiero:
Which libsrtp version will be the first one to include cryptex support? I don't see it referenced in any of the `CHANGES` files, neither on the `main` not on the `2_x_dev` branch. It would help planning support for the feature in applications in a conditional way.

lminiero:
Thanks!

pabuhler:
@fippo why would you want to conditionally enable at compile time? It is a config option when creating a session, it should have no effect if not enabled.
Add compile time config is an option I guess but not sure I understand the use case.

PR #512, opened 2020-11-17
Mbedtls support

ycyang1229:
Recently focusing on the implementation of webrtc on rtos-based embedded devices, so adding the support of mbedtls. Hope this can help someone using this lib with mbedtls. thanks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ set(ENABLE_DEBUG_LOGGING OFF CACHE BOOL "Enable debug logging in all modules")
 set(ERR_REPORTING_STDOUT OFF CACHE BOOL "Enable logging to stdout")
 set(ERR_REPORTING_FILE "" CACHE FILEPATH "Use file for logging")
 set(ENABLE_OPENSSL OFF CACHE BOOL "Enable OpenSSL crypto engine")
-set(ENABLE_MBEDTLS ON CACHE BOOL "Enable MbedTLS crypto engine")
+set(ENABLE_MBEDTLS OFF CACHE BOOL "Enable MbedTLS crypto engine")
 set(TEST_APPS ON CACHE BOOL "Build test applications")
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ project(libsrtp2 LANGUAGES C)
 
 set(PACKAGE_VERSION 2.4.0)
 set(PACKAGE_STRING "${CMAKE_PROJECT_NAME} ${PACKAGE_VERSION}")
-
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR};${CMAKE_MODULE_PATH}")
 include(TestBigEndian)
 include(CheckIncludeFile)
 include(CheckFunctionExists)
@@ -54,15 +54,27 @@ set(ENABLE_DEBUG_LOGGING OFF CACHE BOOL "Enable debug logging in all modules")
 set(ERR_REPORTING_STDOUT OFF CACHE BOOL "Enable logging to stdout")
 set(ERR_REPORTING_FILE "" CACHE FILEPATH "Use file for logging")
 set(ENABLE_OPENSSL OFF CACHE BOOL "Enable OpenSSL crypto engine")
+set(ENABLE_MBEDTLS ON CACHE BOOL "Enable MbedTLS crypto engine")
 set(TEST_APPS ON CACHE BOOL "Build test applications")
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
+
+if(ENABLE_OPENSSL AND ENABLE_MBEDTLS)
+  message(FATAL_ERROR "ssl conflict. can not enable openssl and mbedtls simultaneously.")
+endif()
 
 if(ENABLE_OPENSSL)
   find_package(OpenSSL REQUIRED)
   include_directories(${OPENSSL_INCLUDE_DIR})
+  set(OPENSSL ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
+  set(GCM ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
 endif()
-set(OPENSSL ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
-set(GCM ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
+
+if(ENABLE_MBEDTLS)
+  find_package(MbedTLS REQUIRED)
+  include_directories(${MBEDTLS_INCLUDE_DIRS})
+  set(MBEDTLS ${ENABLE_MBEDTLS} CACHE BOOL INTERNAL)
+  set(GCM ${ENABLE_MBEDTLS} CACHE BOOL INTERNAL)
+endif()
 
 set(CONFIG_FILE_DIR ${CMAKE_CURRENT_BINARY_DIR})
 include_directories(${CONFIG_FILE_DIR})
@@ -85,6 +97,11 @@ if(ENABLE_OPENSSL)
     crypto/cipher/aes_icm_ossl.c
     crypto/cipher/aes_gcm_ossl.c
   )
+elseif(ENABLE_MBEDTLS)
+  list(APPEND CIPHERS_SOURCES_C
+    crypto/cipher/aes_icm_mbedtls.c
+    crypto/cipher/aes_gcm_mbedtls.c
+  )
 else()
   list(APPEND  CIPHERS_SOURCES_C
     crypto/cipher/aes.c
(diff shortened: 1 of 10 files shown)


fluffy:
I'd like to add the MBed support but I want to make sure we get a few people to review this carefully.

ycyang1229:
> I'd like to add the MBed support but I want to make sure we get a few people to review this carefully.

Thanks a lot. It is nice to have someone to review this work.

crypto/hash/hmac_mbedtls.c
```diff
@@ -0,0 +1,253 @@
...
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include "auth.h"
+#include "alloc.h"
+#include "err.h" /* for srtp_debug */
+#include "mbedtls/md.h"
```
pabuhler, line 51: this should maybe use angle brackets for consistency, ie #inlcude <mbedtls/md.h>

pabuhler (approved):
As far as I can tell this is ok. I have done testing and it works on Ubuntu for me. The majority of the change is the same as the openssl/nss integration.
As a follow up note we should look at converting more tests to the cmake build, supporting vagrind in travis for cmake and finally move the test vectors in to a common file that can be shared between back end implementations.

crypto/hash/hmac_mbedtls.c
ycyang1229: correcting it, thanks.

ycyang1229:
> As far as I can tell this is ok. I have done testing and it works on Ubuntu for me. The majority of the change is the same as the openssl/nss integration.
> As a follow up note we should look at converting more tests to the cmake build, supporting vagrind in travis for cmake and finally move the test vectors in to a common file that can be shared between back end implementations.

Cool, if anything I can help, please let me know. big thanks.

config_in_cmake.h
```diff
@@ -21,6 +21,9 @@
 /* Define this to use AES-GCM. */
 #cmakedefine GCM 1
 
+/* Define this to use MBEDTLS. */
+#cmakedefine MBEDTLS 1
```
bifurcation, line 25: I would reorder these to put `MBEDTLS` just after `OPENSSL`.

crypto/hash/hmac_mbedtls.c
```diff
@@ -0,0 +1,253 @@
...
+        result[i] = hash_value[i];
+    }
+
+    debug_print(srtp_mod_hmac, "output: %s",
+                srtp_octet_string_hex_string(hash_value, tag_len));
+
+    return srtp_err_status_ok;
+}
+
+/* begin test case 0 */
```
bifurcation, line 201: It seems like the test cases should be common across the various implementations.  The test data should hash to the same thing regardless of whether you're using OpenSSL, mBedTLS, etc.  Suggest factoring the tests out to a different file.  I haven't looked at the encryption code, but I suspect the same comments apply.  Perhaps we could just make a `crypto_test_cases.c` file that contains all of them.  Then the individual methods can just point to the one that makes sense.

crypto/include/cipher_types.h
```diff
@@ -74,6 +74,9 @@ extern srtp_debug_module_t srtp_mod_aes_icm;
 #ifdef OPENSSL
 extern srtp_debug_module_t srtp_mod_aes_gcm;
 #endif
+#ifdef MBEDTLS
```
bifurcation, line 77: This is getting wordy (as opposed to just doing `#if defined(OPENSSL) || defined(MBEDTLS) || defined(NSS)`).  And don't we have a `GCM` variable anyway?  Seems like we should use `#ifdef GCM`, or if that doesn't work, the `#if defined || defined || defined` as above.

bifurcation (commented):
Here are a few comments from me.  I'm not an mBedTLS expert, though, so I have asked @hannestschofenig to review.  He tells me he'll take a look tomorrow.

paulej (approved):

CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ project(libsrtp2 LANGUAGES C)
 
 set(PACKAGE_VERSION 2.4.0)
 set(PACKAGE_STRING "${CMAKE_PROJECT_NAME} ${PACKAGE_VERSION}")
-
+set(CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR};${CMAKE_MODULE_PATH}")
```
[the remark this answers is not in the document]
ycyang1229: I added this for looking for FindMbedTLS.cmake, and it will be used for searching mbedtls related header files and libs. As I remember, I need to use such a cmake files to search mbedtls related stuff. Did mistake it?

CMakeLists.txt
```diff
@@ -54,15 +54,27 @@ set(ENABLE_DEBUG_LOGGING OFF CACHE BOOL "Enable debug logging in all modules")
...
 if(ENABLE_OPENSSL)
   find_package(OpenSSL REQUIRED)
   include_directories(${OPENSSL_INCLUDE_DIR})
+  set(OPENSSL ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
+  set(GCM ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
+endif()
+
+if(ENABLE_MBEDTLS)
+  find_package(MbedTLS REQUIRED)
+  include_directories(${MBEDTLS_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
ycyang1229: The same reason as the above stuff, FindMbedTLS.cmake.

config_in_cmake.h
ycyang1229: Sure, I will change the order.

crypto/hash/hmac_mbedtls.c
ycyang1229: This idea came up when I was coding. But I saw the code inside openssl, I left it as the same as the part of openssl. Do you want to add this part, or cut another pr?

pabuhler: I have a local change that does this now, I will create a PR after this is merged.

crypto/include/cipher_types.h
pabuhler: If you make this change then great other wise we can do it in a separate clean up PR.

hannestschofenig:
Thanks for adding Mbed TLS support to libsrtp!

I took a look at the code, as requested by Richard. The Mbed TLS related code looks good to me. I cannot comment on the requirements for tests, documentation and coding style. 

(Note that we have been working on a new API, called the PSA Crypto API, which allows hardware crypto to be used conveniently. I understand that you are using our current API instead but in the future there may be benefit to also take a look at the new API. You can find the details of the PSA Crypto API at  https://armmbed.github.io/mbed-crypto/PSA_Cryptography_API_Specification.pdf and I have placed examples at https://github.com/ARMmbed/mbedtls/pull/3833)

Just for my understanding: Are you expecting this code to be used in combination with the Mbed TLS code (with MBEDTLS_SSL_DTLS_SRTP enabled) so that the key exchange is done based on the code in https://github.com/ARMmbed/mbedtls/ and then for SRTP you are using the code from this library?

crypto/include/cipher_types.h
ycyang1229: I did not go through the nss part in detail, and see the nss in the cmakefiles either. So I am not sure it includes the gcm or not. Went through the nss part quickly, and it should have the gcm part. I changed it as the second choice #if (defined(OPENSSL) || defined(MBEDTLS) || defined(NSS)).

ycyang1229:
> Thanks for adding Mbed TLS support to libsrtp!
> 
> I took a look at the code, as requested by Richard. The Mbed TLS related code looks good to me. I cannot comment on the requirements for tests, documentation and coding style.
> 
> (Note that we have been working on a new API, called the PSA Crypto API, which allows hardware crypto to be used conveniently. I understand that you are using our current API instead but in the future there may be benefit to also take a look at the new API. You can find the details of the PSA Crypto API at https://armmbed.github.io/mbed-crypto/PSA_Cryptography_API_Specification.pdf and I have placed examples at [ARMmbed/mbedtls#3833](https://github.com/ARMmbed/mbedtls/pull/3833))
> 
> Just for my understanding: Are you expecting this code to be used in combination with the Mbed TLS code (with MBEDTLS_SSL_DTLS_SRTP enabled) so that the key exchange is done based on the code in https://github.com/ARMmbed/mbedtls/ and then for SRTP you are using the code from this library?

Sure, I would love to take a look, and it is my plesaure. Yes, I took this pr(https://github.com/ARMmbed/mbedtls/pull/3235/files) in the beginning. Big thanks.

pabuhler:
@ycyang1229  the failed build is related to some meson issues in travis that started yesterday, so I think you just need to ignore that for now.

ycyang1229:
> @ycyang1229 the failed build is related to some meson issues in travis that started yesterday, so I think you just need to ignore that for now.

Thanks for immediate response. I just tried to see why. :D

bifurcation (approved):
Thanks for the review, @hannestschofenig.  I filed #517 to track the PSA upgrade.

Given the responses on my other comments, this looks good to merge.

PR #516, opened 2021-01-08
temporary disable meson ci builds on mips

pabuhler:
disabling build until some one can look in to it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -236,9 +236,9 @@ matrix:
           exe_wrapper = 'qemu-mips-static'
           pkgconfig   = 'false'
           EOF
-        - sudo docker exec --tty mipsX meson setup --cross-file mips-cross-file.txt -Ddefault_library=static -Db_staticpic=false builddir
-        - sudo docker exec --tty mipsX meson compile -C builddir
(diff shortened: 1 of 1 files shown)


tp-m:
Do you have a link to the failure that needs fixing here?

pabuhler:
sorry I should have included one in the original commit, I think this is an example of it, https://travis-ci.org/github/cisco/libsrtp/jobs/753516746 , it just started to fail one or two days a go.

tp-m:
Thanks.

I think the new meson 0.56.1 release which hit servers a few days ago [broke python 3.5.x compatibility](https://github.com/mesonbuild/meson/issues/8161).

This should be fixed now with the new 0.56.2 release that was rolled yesterday to fix this.

pabuhler:
ok, I guess we wait abit and re - enable. Looking to move to git hub actions for ci building, will likely add meson as a separate build check that can be ignored if not always passing,  this is I think the third time it has broken for external reasons.

nirbheek:
Regardless of what you decide to do, we would really appreciate if you could CC tim or myself when such things happen. We would prefer to fix the issue quickly if possible.

PR #520, opened 2021-01-11
use nss based hmac when nss enabled

pabuhler:
This was a missing part of the nss support.

Disable sha1_driver if USE_EXTERNAL_CRYPTO is enabled. The sha1_driver is used to test internal sha1 implementation but that is only used when there is no external hmac implementation.

### DIFF
--- a/crypto/hash/hmac_nss.c
+++ b/crypto/hash/hmac_nss.c
@@ -92,7 +92,8 @@ static srtp_err_status_t srtp_hmac_alloc(srtp_auth_t **a,
         return srtp_err_status_alloc_fail;
     }
 
-    hmac = (srtp_hmac_nss_ctx_t *)srtp_crypto_alloc(sizeof(srtp_hmac_nss_ctx_t));
+    hmac =
+        (srtp_hmac_nss_ctx_t *)srtp_crypto_alloc(sizeof(srtp_hmac_nss_ctx_t));
     if (hmac == NULL) {
         NSS_ShutdownContext(nss);
         srtp_crypto_free(*a);
@@ -184,7 +185,7 @@ static srtp_err_status_t srtp_hmac_init(void *statev,
 
     SECItem key_item = { siBuffer, (unsigned char *)key, key_len };
     sym_key = PK11_ImportSymKey(slot, CKM_SHA_1_HMAC, PK11_OriginUnwrap,
-                            CKA_SIGN, &key_item, NULL);
+                                CKA_SIGN, &key_item, NULL);
     PK11_FreeSlot(slot);
 
     if (!sym_key) {
@@ -246,7 +247,8 @@ static srtp_err_status_t srtp_hmac_compute(void *statev,
         return srtp_err_status_auth_fail;
     }
 
-    if (PK11_DigestFinal(hmac->ctx, hash_value, &len, SHA1_DIGEST_SIZE) != SECSuccess) {
+    if (PK11_DigestFinal(hmac->ctx, hash_value, &len, SHA1_DIGEST_SIZE) !=
+        SECSuccess) {
         return srtp_err_status_auth_fail;
     }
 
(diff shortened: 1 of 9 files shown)


crypto/hash/hmac_nss.c
```diff
@@ -0,0 +1,284 @@
...
+    /* check output length - should be less than 20 bytes */
+    if (out_len > SHA1_DIGEST_SIZE) {
+        return srtp_err_status_bad_param;
+    }
+
+    /* Initialize NSS equiv of NSS_NoDB_Init(NULL) */
+    nss = NSS_InitContext("", "", "", "", NULL,
+                          NSS_INIT_READONLY | NSS_INIT_NOCERTDB |
+                              NSS_INIT_NOMODDB | NSS_INIT_FORCEOPEN |
+                              NSS_INIT_OPTIMIZESPACE);
```
nils-ohlmeier, line 85: This makes me wonder if in the future we should move this into one shared space where NSS gets inited only ones, which I think should be sufficient.

crypto/hash/hmac_nss.c
```diff
@@ -0,0 +1,284 @@
...
+    if (!nss) {
+        return srtp_err_status_auth_fail;
+    }
+
+    *a = (srtp_auth_t *)srtp_crypto_alloc(sizeof(srtp_auth_t));
+    if (*a == NULL) {
+        return srtp_err_status_alloc_fail;
```
nils-ohlmeier, line 92: Why is this not shutting down the NSS context, like below in line 97?

nils-ohlmeier (commented):
Thank you for adding this. Awesome!

Probably should also be reviewed by experts like @bifurcation

crypto/hash/hmac_nss.c
pabuhler: Makes sense, I had just followed the pattern from aes_icm_nss.c & aes_gcm_nss.c . We could make a shared initialization state between these 3 files. Perhaps @bifurcation  know if this is a good idea or not.

crypto/hash/hmac_nss.c
pabuhler: oversight on my part, thanks

crypto/hash/hmac_nss.c
bifurcation: This doesn't seems super compelling to me.  In a C++ world, we would just do this with RAII, but in C, we'll have `NSS_ShutdownContext` calls scattered around anyway.  So all you would really gain is a convenient abbreviation for this and the call in `_dealloc`.

crypto/hash/hmac_nss.c
```diff
@@ -0,0 +1,287 @@
...
+        hmac->key = NULL;
+    }
+
+    PK11SlotInfo *slot = PK11_GetBestSlot(CKM_SHA_1_HMAC, NULL);
+    if (!slot) {
+        return srtp_err_status_bad_param;
+    }
+
+    SECItem key_item = { siBuffer, (unsigned char *)key, key_len };
+    sym_key = PK11_ImportSymKey(slot, CKM_SHA_1_HMAC, PK11_OriginUnwrap,
```
bifurcation, line 188: This stuff actually seems a little more useful to make common across the NSS modules.  It seems like you could have a function of the following form that would wrap these lines:

```
PK11SymKey* srtp_nss_import_key(CK_METHOD method, const uint8_t *key, size_t key_len);
```

If you take that together with startup / shutdown methods, maybe that's enough for another `.c/.h`?

crypto/hash/hmac_nss.c
bifurcation: I'm actually not sure each algorithm needs to hold its own NSS handle.  You could probably do a simple ref-counted thing, where you have a global NSS instance, start it up for the first user, and shut it down once when the refcount gets back to zero.  That would obviously need to be in its own module.

bifurcation (approved):
A couple comments inline, but overall this lgtm.

Tagging @jcjones here in case he has a chance to look, as he has more recent NSS experience than me.

pabuhler:
I have merged this now, the intention of this PR was to add the missing hmac support for NSS. If the entire NSS code can/should be refactored or cleaned up then perhaps there should be an issues created based off the discussion here. It would be best if it was driven by some one familiar with NSS or that actively uses the NSS backend.

Issue #523, opened 2021-01-27
Test failures with NSS 3.60.1


bifurcation:
@pabuhler reports that after upgrading to NSS 3.60.1, tests exercising AES-GCM with NSS fail.  Creating this bug to capture details.

cc @martinthomson

pabuhler:
The PK11_Encrypt function is returning -1 when running the cipher self test, will investigate some more.

martinthomson:
Opened https://bugzilla.mozilla.org/show_bug.cgi?id=1689209

pabuhler:
There was an API change in NSS that causes this, https://bugzilla.mozilla.org/show_bug.cgi?id=1637488#c3 has details and possible solutions. PR is coming.

martinthomson:
Happy to provide review if you decide to use the new parameter structs.

pabuhler:
@martinthomson thanks for the quick help yesterday. I have done the simplest fix for now and you are welcome to review that.
As I understood the other options introduce nss version requirements. As I do not know enough about what versions of nss are in use and there does not seam to be anything to gain by adding these version requirements I have chosen to go with this fix for now.
If you have a better option then we can look at doing that.

martinthomson:
Kevin was the one who managed to remember the change :)  Happy to facilitate of course.

PR #524, opened 2021-01-28
support running against nss >= 3.52

pabuhler:
Fix for #523

This is described in more detail at:
https://bugzilla.mozilla.org/show_bug.cgi?id=1637488#c3

Have decided to go with using NSS_PKCS11_2_0_COMPAT for now
as it provides the most fexibility.
Consider moving to the new PK11_AEADOp api in the future but
that will put a requirment on nss version at both compile and
run time.

### DIFF
--- a/crypto/hash/hmac_nss.c
+++ b/crypto/hash/hmac_nss.c
@@ -42,6 +42,9 @@
 #include "alloc.h"
 #include "err.h" /* for srtp_debug */
 #include "auth_test_cases.h"
+
+#define NSS_PKCS11_2_0_COMPAT 1
+
 #include <nss.h>
 #include <pk11pub.h>
 
--- a/crypto/include/aes_gcm.h
+++ b/crypto/include/aes_gcm.h
@@ -85,6 +85,8 @@ typedef struct {
 
 #ifdef NSS
 
+#define NSS_PKCS11_2_0_COMPAT 1
+
 #include <nss.h>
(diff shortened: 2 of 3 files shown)


martinthomson (approved):
The alternative change, which would be to deal with the PKCS#11 change, might be better.  I would recommend opening an issue to track that.

crypto/hash/hmac_nss.c
```diff
@@ -42,6 +42,9 @@
 #include "alloc.h"
 #include "err.h" /* for srtp_debug */
 #include "auth_test_cases.h"
+
+#define NSS_PKCS11_2_0_COMPAT 1
```
bifurcation, line 46: Do we need this flag outside of the GCM case?

bifurcation (approved):
I would prefer an approach here that's more future-oriented.  For example, perhaps we could offer consumers a switch to enable old NSS, and map things over internally to libsrtp.

If that's too hard, this is a fine hack.  But as @martinthomson suggests, we should file an issue to move to the new API.

crypto/hash/hmac_nss.c
pabuhler: likely not i was just being consistent with setting it before all includes

pabuhler:
#525 is a follow on issues. Until there is value in placing the burden of a version check (build/runtime) on uses of libsrtp I think  this is the best solution.

PR #529, opened 2021-03-31
Remove EKT

### LINKED ISSUE #384: Remove ekt.h & ekt.c
The current EKT implementation in master is incorrect so it should either be fixed or removed. The problem is that it is in the public api so would need to wait with removing it until next major version, so this is a place holder task.

bifurcation:
The EKT code in libsrtp has long been incorrect and unusable.  This PR removes it from the code base.

Fixes #384


paulej (approved):

pabuhler (changes requested):
I absolutely agree, with removing EKT from the lib, but it is part of public API even if it was not working so simply removing it could be problematic and the reason that #384 was still open. There are binary distributions of libsrtp so we need to ensure that any change to the API needs to be sure it is ABI compatible. I think removing members from srtp_policy_t is problematic. Lets just be sure we dont break things be fore merging.

paulej:
Who would be relying on this public API given it doesn't function?  It seems more confusing keeping it than it, especially since there is a new RFC for EKT.

pabuhler:
@paulej  hopefully no one and removing the header file to force that fact that it is broken is a good thing. Removing the header may result in the need for some application to recompile with he new version if they where including it for some reason and that is fine. The problem I think is the removal of  "srtp_ekt_policy_t ekt;" this will change the size and offsets of the srtp_policy_t struct which would be a problem if an application uses a prebuilt distribution of libsrtp, such a change should be done in a major version only.

bifurcation:
@pabuhler The point about changing the ABI is a good one.  I've updated the PR so that there's a `deprecated_ekt` field in the `srtp_policy_t` struct where the `ekt` field was before.  In order to keep the file deleted, I changed the type of the field to `void*`, but pointers are pointers, so it should maintain the layout of the rest of the struct.

pabuhler:
@bifurcation there are two failed builds, one is clang format and that should be fixed, the other can probably be ignored for now.

bifurcation:
@pabuhler Looks like all the builds are passing now except for the one where the tooling is broken.

PR #530, opened 2021-04-02
Rust migration

bifurcation:
Since libsrtp handles packets directly off the network, before any authentication has been done, it should have stronger protections against memory corruption.  This PR proposes a way to work toward a migration to Rust:

* Establish a `libsrtp_rs` library within the libsrtp code base
* Enable optional usage of this library through config parameters
* Gradually replicate functionality from the C code base to the Rust library, and call via FFI from what remains as C

This strategy assures that API and test invariants are maintained through the migration.  It also lets consumers select when they want to move over.  In the long run, of course, the goal would be to have a pure-Rust library with a C API to support legacy consumers.

At this point, Rust integration is only supported with the CMake build, and requires that the [corrosion](https://github.com/AndrewGaspar/corrosion) CMake extension be installed.  To build with Rust:

```sh
$ cmake -B build -DENABLE_RUST=ON
$ cmake --build build
```


libsrtp_rs/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+authors = ["Richard Barnes <rlb@ipv.sx>"]
+edition = "2018"
+
+[dependencies]
+
+[build-dependencies]
+cbindgen = "0.18.0"
+
+[lib]
+crate-type=["staticlib"]
```
alex, line 13:
```suggestion
crate-type = ["staticlib"]
```

libsrtp_rs/include/libsrtp_rs/libsrtp_rs.h
```diff
@@ -0,0 +1 @@
+void rust_noop();
```
alex, line 1: Was this hand generated or with cbindgen? I see you depend on cbindgen, but I don't see it used anywhere.

alex (commented):
I imagine for real use cases you'll want to use bindgen to generate Rust to call into libsrtp headers (or at least share data structures).

libsrtp_rs/include/libsrtp_rs/libsrtp_rs.h
bifurcation: This was hand-generated.  I had something weird with my rust setup that caused cbindgen not to work, so I hand-wrote something to get it going and forgot to remove the dependency.

nils-ohlmeier:
In general I totally agree with the idea of having as much SRTP code written in Rust.

I would also strongly suggest to use bindgen as much as possible, to avoid problems by hand written binding code.

Is the end goal here to have a pure Rust SRTP crate and another crate with the bindings for C?
That setup has the advantage that pure Rust implementations do not have to pull in the bindings code if they don't need it.

bifurcation:
@nils-ohlmeier Yes, that's the end goal I had in mind, a Rust crate with C bindings on top for those who need them.  The idea of gradually migrating over libsrtp code was to ensure correctness as the migration is done.  

There is some tension here between idiomatic Rust and compatibility with C (including other parts of libsrtp).  For example, libsrtp's crypto agility is done with function pointer tables, which might not be how you would do it in Rust.

Logistically, since this is going to take a little while it would probably be best to create a feature branch and land PRs to that.

sdroege:
> a Rust crate with C bindings on top

You might want to also take a look at [cargo-c](https://github.com/lu-zero/cargo-c), which adds features to cargo to build a proper C-compatible shared/static library, generates headers via [cbindgen](https://github.com/eqrion/cbindgen) from your `extern "C"` Rust API and installs a pkg-config file together with the library and headers.

Sean-Der:
A rust implementation exists today https://github.com/webrtc-rs/srtp it was ported from https://github.com/pion/webrtc 

Supports `SRTP_AES128_CM_SHA1_80` and `SRTP_AEAD_AES_128_GCM`. Is it missing any features that stop it from being useful today?

bifurcation:
Closing this while I come up with a more thoroughly worked-out proposal.

PR #531, opened 2021-04-05
Remove dead code

bifurcation:
As I was looking into migrating things to Rust, I noticed there are a few chunks of around.  In addition to the ones I noticed in the SHA-1 code, I scanned for `#if 0` blocks and used `cflow -r` to find unused functions.


crypto/include/auth.h
```diff
@@ -156,16 +156,6 @@ srtp_err_status_t srtp_auth_type_self_test(const srtp_auth_type_t *at);
...
-/*
- * srtp_replace_auth_type(ct, id)
- *
- * replaces srtp's kernel's auth type implementation for the auth_type id
- * with a new one passed in externally.  The new auth type must pass all the
- * existing auth_type's self tests as well as its own.
- */
-srtp_err_status_t srtp_replace_auth_type(const srtp_auth_type_t *ct,
-                                         srtp_auth_type_id_t id);
-
```
JonathanLennox, line 168: The `replace` functions are API, not dead code - they're to allow a third-party crypto implementation to be injected.

crypto/math/datatypes.c
```diff
@@ -302,19 +277,14 @@ void v128_left_shift(v128_t *x, int shift)
 
 int bitvector_get_bit(const bitvector_t *v, int bit_index)
 {
-    return _bitvector_get_bit(v, bit_index);
+    return /bitvector_get_bit(v, bit_index);
```
JonathanLennox, line 280: ??? Does this compile?

bifurcation: It does, but only because it's within `#ifndef DATATYPES_USE_MACROS`, and that doesn't appear to be the case for either my local build (macOS, CMake, default config) or any of the CI builds.  Maybe we should remove that option?

crypto/include/auth.h
bifurcation: That's a little surprising to me; I didn't realize the `crypto/include` headers were considered public API.

JonathanLennox: Just `crypto/include/cipher.h`, `crypto/include/auth.h`, and `crypto/include/crypto_types.h` are public - see the `install` targets in `Makefile.in` or `CMakeLists.txt`.

crypto/math/datatypes.c
JonathanLennox: `DATATYPES_USE_MACROS` is defined unconditionally in `datatypes.h`, so yeah, removing its `#ifndef` conditions is probably a good idea.

pabuhler:
@bifurcation  could you fix the clang format errors

bifurcation:
@pabuhler Looks like all builds are passing now except for the broken mips/meson build.

PR #546, opened 2021-07-23
Improve OpenSSL KDF check

nazar-pc:
Without this change libsrtp2 usage with OpenSSL as subproject will fail with:
> subprojects/libsrtp2/meson.build:134:10: ERROR: Dependencies must be external dependencies

I had to include 2 additional conditions:
* make sure dependency is not internal, otherwise we know `cc.has_function()` will fail as error above suggests
* check if user explicitly disabled `crypto-library-kdf` option (in which case there is no need to check anything and configuration incorrectly didn't take that into account)

P.S. OpenSSL as subproject/wrap will hopefully soon be a thing, see https://github.com/mesonbuild/wrapdb/pull/113

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -133,7 +133,7 @@ if crypto_library == 'openssl'
   # forks of OpenSSL: https://github.com/cisco/libsrtp/issues/458
   if (
     openssl_dep.type_name() != 'internal' and
-    get_option('crypto-library-kdf').disabled() != true and
+    not get_option('crypto-library-kdf').disabled() and
     cc.has_function('kdf_srtp', dependencies: openssl_dep)
   )
     cdata.set('OPENSSL_KDF', true)
--- a/meson.build
+++ b/meson.build
@@ -131,7 +131,11 @@ if crypto_library == 'openssl'
   use_openssl = true
   # NOTE: This is not available in upstream OpenSSL yet. It's only in 'certain'
   # forks of OpenSSL: https://github.com/cisco/libsrtp/issues/458
-  if cc.has_function('kdf_srtp', dependencies: openssl_dep)
+  if (
+    openssl_dep.type_name() != 'internal' and
+    get_option('crypto-library-kdf').disabled() != true and
+    cc.has_function('kdf_srtp', dependencies: openssl_dep)
+  )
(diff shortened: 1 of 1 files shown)


nazar-pc:
Can someone review this, please? It is a pretty trivial, but also important change

paulej:
Looks beautiful, but does it work? I have no idea. Who is the Meson expert?

meson.build
```diff
@@ -131,7 +131,11 @@ if crypto_library == 'openssl'
   use_openssl = true
   # NOTE: This is not available in upstream OpenSSL yet. It's only in 'certain'
   # forks of OpenSSL: https://github.com/cisco/libsrtp/issues/458
-  if cc.has_function('kdf_srtp', dependencies: openssl_dep)
+  if (
+    openssl_dep.type_name() != 'internal' and
+    get_option('crypto-library-kdf').disabled() != true and
+    cc.has_function('kdf_srtp', dependencies: openssl_dep)
+  )
```
tp-m, line 138: Looks fine to me, but
- why do we have to exclude openssl as meson subproject here? I mean, is there a technical reason or is it just a shortcut because you know your openssl wrap doesn't provide this? (But it could in future, if a new version of openssl starts shipping this, or someone makes a modified subproject with it included)
- use `not get_option(..).disabled()`

nazar-pc: As I mentioned in the very first comment `cc.has_function()` doesn't work with dependencies that are subprojects.
Applied second suggestion.

tp-m: Ah right, of course. Sorry, only started reading from the most recent email onwards :)

nazar-pc:
@tp-m can you confirm this looks good to you?

As to whether it works, it definitely does in https://github.com/versatica/mediasoup/pull/622 where CI runs on Windows, Linux and macOS with various compilers and all builds/runs fine. And there is really nothing so significant here to break anything.

tp-m:
> @tp-m can you confirm this looks good to you?

Yes, looks good to me.

Looking forward to be able to use this in GStreamer as well!

pabuhler:
OpenSSL KDF check should probably be removed, I do not know of any actual use of OpenSSL KDF with libSRTP but that is a different issue.

PR #552, opened 2021-08-15
Remove tabs, fix typos, etc.

paulej:


bifurcation:
Shouldn’t this just apply clang-format?

paulej:
That might change a lot of lines. I just wanted to fix the whitespace, which doesn't look so disruptive. Also, that won't work on .md and .css files.


pabuhler:
speaking of clang-format   we should probably update the version in use.

pabuhler:
@paulej have you run clang format after making the changes to source files ?  I have added clang format as a git hub action now so if you rebase or merge from master the format check will run on this PR

paulej:
@pabuhler No. I didn't want to create formatting changes that would look extensive. I just wanted to remove the inconsistent use of tabs.

PR #562, opened 2021-09-13
Fix typo introduced by PR 559

lminiero:
This PR fixes a typo introduced by #559, where for `SRTP_AES_GCM_256` the returned length is `SRTP_AES_ICM_256_KEY_LEN_WSALT` instead of `SRTP_AES_GCM_256_KEY_LEN_WSALT`, which causes problems when SRTP sessions are created using AES-GCM-256.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -788,7 +788,7 @@ static inline int full_key_length(const srtp_cipher_type_t *cipher)
     case SRTP_AES_GCM_128:
         return SRTP_AES_GCM_128_KEY_LEN_WSALT;
     case SRTP_AES_GCM_256:
-        return SRTP_AES_ICM_256_KEY_LEN_WSALT;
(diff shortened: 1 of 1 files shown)


paulej (approved):

pabuhler:
strange there was no test to catch this :(

lminiero:
> strange there was no test to catch this :(

@pabuhler not sure, but my guess is this would have probably been tricky to catch using unit tests. We only noticed the issue when we were notified by someone using 2.4.1 about problems with `srtp_unprotect` on incoming RTCP on the server, but there were no reported failures in `srtp_protect` for outgoing RTP or RTCP for instance: when we worked on replicating the issue, we noticed media wouldn't work either, and that was because `srtp_protect` would "succeed", but then the browser would not render anything, because of failures to unprotect on that end.

PR #572, opened 2021-10-13
rtp_decoder support for MKI and RCCm3 ROC field

paulr227:
MKI: support for MKI was added to libsrtp in 2.1.0, however the support wasn't extended to rtp_decder. These changes update rtp_decoder to use the srtp_unprotect_mki() and srtp_unprotect_rtcp_mki() APIs and to accept a new argument for specifying the MKI of the master key passed in the stream policy provided to libsrtp.

RCCm3: is one of the three integrity transforms specified in RFC4771 to support synchronisation of sender and receiver ROC by inclusion of the sender's ROC value in the optional authentication field at the end of the SRTP packet. RCCm3 is the only unauthenticated transform of the three, and therefore the only one that is supported without changes within libsrtp itself. A new rtp_decoder argument is introduced to specify the ROC transmission rate parameter. Any future support for RCCm1 and RCCm2 would be compatible with this argument.

New arguments introduced:
       -i <mki>  sets the MKI given in hexadecimal
       -n <rate> RFC4771 ROC transmission rate (assumes RCCm3)

### DIFF
--- a/test/rtp_decoder.c
+++ b/test/rtp_decoder.c
@@ -590,7 +590,8 @@ int main(int argc, char *argv[])
                 octet_string_hex_string(key + key_octets, salt_octets));
 
         /* Extract MKI */
-        mki_size = hex_string_to_octet_string(mki, input_mki, strlen(input_mki));
+        mki_size =
+            hex_string_to_octet_string(mki, input_mki, strlen(input_mki));
         mki_size /= 2;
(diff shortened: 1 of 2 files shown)


pabuhler:
Closing PR for now, can be reopened if needed.
MIK support is now been added to rtp_decoder.
If RCCm3 is desired then it should be added as a feature to core libSRTP.

PR #573, opened 2021-10-13
Export CMake Targets

SE2Dev:
The lack of any properly exported CMake target makes it somewhat difficult for libraries like `libdatachannel` to have a proper cmake configuration for consuming libSRTP.

This pull request effectively allows developers to consume libSRTP as seen here:
```cmake
find_library(libSRTP REQUIRED)
# ...
target_link_libraries(server PRIVATE libSRTP::srtp2)
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -243,7 +243,7 @@ set_target_properties(srtp2 PROPERTIES VERSION ${CMAKE_PROJECT_VERSION})
 target_include_directories(srtp2 PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/crypto/include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-  $<INSTALL_INTERFACE:/include>
+  $<INSTALL_INTERFACE:include>
 )
 if(ENABLE_OPENSSL)
   target_include_directories(srtp2 PRIVATE ${OPENSSL_INCLUDE_DIR})
@@ -261,7 +261,7 @@ if(WIN32)
 endif()
 
 install(TARGETS srtp2 DESTINATION lib
-  EXPORT libSRTPTargets 
+  EXPORT libSRTPTargets
 )
 
 install(FILES include/srtp.h crypto/include/auth.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,7 +240,11 @@ add_library(srtp2
 
 set_target_properties(srtp2 PROPERTIES VERSION ${CMAKE_PROJECT_VERSION})
 
-target_include_directories(srtp2 PUBLIC crypto/include include)
+target_include_directories(srtp2 PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/crypto/include>
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  $<INSTALL_INTERFACE:/include>
+)
 if(ENABLE_OPENSSL)
   target_include_directories(srtp2 PRIVATE ${OPENSSL_INCLUDE_DIR})
   target_link_libraries(srtp2 OpenSSL::Crypto)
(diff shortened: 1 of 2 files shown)


NancyLi1013:
Hi @pabuhler 

I'm the maintainer of vcpkg. Could you please help confirm when this PR will be accepted and merged?

Thanks

LilyWangLL:
Ping @pabuhler for response.

pabuhler:
@LilyWangLL I did a small change that was needed for me to get this to work with a test project. Did the change make sense to you?

Also I am no cmake guru but this looks like alot more than I would have expected in order to make it work with find_package. I am guess you know what you are doing ?

CMakeLists.txt
```diff
@@ -256,7 +260,10 @@ if(WIN32)
   target_compile_definitions(srtp2 PUBLIC _CRT_SECURE_NO_WARNINGS)
 endif()
 
-install(TARGETS srtp2 DESTINATION lib)
+install(TARGETS srtp2 DESTINATION lib
```
LilyWangLL, line 263: Here should be:
```suggestion
install(TARGETS srtp2 EXPORT libSRTPTargets
        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
)
```

CMakeLists.txt
```diff
@@ -352,3 +359,40 @@ if(TEST_APPS)
     endif()
   endif()
 endif()
+
+# Export targets
+install(
+  EXPORT libSRTPTargets
+  FILE libSRTPTargets.cmake
+	NAMESPACE libSRTP::
+	DESTINATION lib/cmake/libSRTP
```
LilyWangLL, line 368: This could make target files install to correct path.
```suggestion
	DESTINATION share/libSRTP
```

CMakeLists.txt
```diff
@@ -352,3 +359,40 @@ if(TEST_APPS)
...
+  COMPATIBILITY AnyNewerVersion
+)
+
+#--------------------------------------------------------------------
+# Install CMake config files
+#--------------------------------------------------------------------
+install(FILES
+  ${CMAKE_CURRENT_BINARY_DIR}/libSRTPConfig.cmake
+  ${CMAKE_CURRENT_BINARY_DIR}/libSRTPConfigVersion.cmake
+  DESTINATION lib/cmake/libSRTP
```
LilyWangLL, line 397: This could make config files install to correct path.
```suggestion
  DESTINATION share/libSRTP
```

LilyWangLL (changes requested):

LilyWangLL:
> @LilyWangLL I did a small change that was needed for me to get this to work with a test project. Did the change make sense to you?
> 
> Also I am no cmake guru but this looks like alot more than I would have expected in order to make it work with find_package. I am guess you know what you are doing ?

Thanks for your reply. Thanks for @SE2Dev's work. I test this PR locally and suggest some changes. If this PR merged, we can use `find_package` to use libsrtp susseed in CMake project.

pabuhler:
I will try out the suggestions locally.

pabuhler:
@LilyWangLL  Is there a way to test suggestions locally without first committing or having to manually edit ? this is more a GitHub question but since you added them I wounder if you know how to use them?  I do not like the idea of committing things I cant tests.

SE2Dev:
> @LilyWangLL Is there a way to test suggestions locally without first committing or having to manually edit ? this is more a GitHub question but since you added them I wounder if you know how to use them? I do not like the idea of committing things I cant tests.

Personally, I usually just use VSCode and set up a devcontainer locally, and ran the CMake "install" target inside the devcontainer so that the files wouldn't be littered all over my system when testing.

CMakeLists.txt
pabuhler: This is a different change, current libsrtp installs to lib regardless if it is shared or static.

CMakeLists.txt
pabuhler: Why is this correct, the share dest folder is not used as far as I see at the moment.

pabuhler:
I have merged the changes as is, it seams to work for me in most cases I have tested, @LilyWangLL   If you have a good argument for using the "share"  folder for installing then please make a new PR.

Issue #567, opened 2021-09-17
call srtp_unprotect return srtp_err_status_auth_fail when rtp seqnum around 655xx but roll back to small value(like 20,30) the api return is ok.


ricelmm:
Hi dear all
i meet one issue whic make me have confused for few days.
i step up one webrtc media server (like licode, mediasoup) and use chrome to publish stream. The server is built by openssl 1.0.2 and libsrtp 2.3.0.
And i found one issue happened at 10% lost network which simulate at PC using clumsy2.0
![image](https://user-images.githubusercontent.com/10380844/133717750-20288c24-b89d-4dad-967b-9812ed2a44b9.png)
when the rtp seqnum at around 655xx and the rtp packet is nack from chrome. i found call srtp_unprotect return srtp_err_status_auth_fail  and when the seqnum roll back to small number this API return is ok. And i enable the libsrtp debug log and i found that log like this:
srtp: function srtp_unprotect
srtp: estimated u_packet index: 000000000000ffb2
srtp: estimated u_packet index: 000000000000ffb2
aes icm ossl: setting iv: 000000008ee3dd6b00000000ffb20000
aes icm ossl: set_counter: 195c023f26efb0a4b2b37224433a0000
hmac sha-1 openssl: input: 9066ffb2d3d287058ee3dd6bbede0001e2016044dba62d99e8dc4f2abbbd1e92
hmac sha-1 openssl: output: ab66673e9e1b429a6830
srtp: computed auth tag:    ab66673e9e1b429a6830
srtp: packet auth tag:      cdb78cfc764fe807f4ab
hmac sha-1 openssl: input: 9066ffb2d3d287058ee3dd6bbede0001e2016044dba62d99e8dc4f2abbbd1e92
hmac sha-1 openssl: output: ab66673e9e1b429a6830
srtp: computed auth tag:    ab66673e9e1b429a6830
srtp: packet auth tag:      cdb78cfc764fe807f4ab

and input rtp_unprotect the packet(hex content) is the same with the packet is captured from wireshark. And we use srtp_profile_aes128_cm_sha1_80, and when we upgrade openssl 1.1.1 and libsrtp 2.3.0 this issue still exist.

so why return fail when rtp seqnum is 655xx and rtp packet is discorder while roll back is ok?

ricelmm:
hello.... anybody can help me?

pabuhler:
[...]

PR #583, opened 2021-10-28
test for out of order after rollover

pabuhler:
This is a test case to check if issue #567 is reproducible.


paulej (approved):
Would prefer the GCM vs ICM test to not depend on `#ifdef`, but otherwise fine.  We pass the cipher as a parameter to this function, then having the calling function test both sequentially.  But it probably doesn't matter much for this test.

PR #593, opened 2022-01-15
Remove compatibility code for legacy OpenSSL to fix LibreSSL build

klemensn:
In current LibreSSL, `HMAC_CTX` aka. `struct hmac_ctx_st` is an opaque
structure as of LibreSSL hmac.h revision 1.15 (14.01.2022) [0], thus
`sizeof(HMAC_CTX)` fails to compile.

The non-legacy code path should compile with LibreSSL versions as old
as 2.7.0 (21.03.2018).

Found while building https://github.com/desktop-app/tg_owt which bundles
libsrtp 2.2.0 [1] on OpenBSD 7.0 -CURRENT/with latest LibreSSL.

Suggestion to remove the legacy code from Theo Buehler, thanks.

0: http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libcrypto/hmac/hmac.h?rev=1.15&content-type=text/x-cvsweb-markup
1: https://github.com/desktop-app/tg_owt/blob/6708e0d31a73e64fe12f54829bf4060c41b2658e/src/third_party/libsrtp/crypto/hash/hmac_ossl.c#L85

### DIFF
--- a/crypto/hash/hmac_ossl.c
+++ b/crypto/hash/hmac_ossl.c
@@ -78,26 +78,6 @@ static srtp_err_status_t srtp_hmac_alloc(srtp_auth_t **a,
         return srtp_err_status_bad_param;
     }
 
-/* OpenSSL 1.1.0 made HMAC_CTX an opaque structure, which must be allocated
-   using HMAC_CTX_new.  But this function doesn't exist in OpenSSL 1.0.x. */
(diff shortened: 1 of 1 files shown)


pabuhler:
If I understand this effectively makes OpenSSL 1.1.0 the minimal supported version ? If that is true then maybe the build files need to be updated to reflect that.
@paulej  what do you think about requiring OpenSSL 1.1.0.

PR #594, opened 2022-01-15
test: Use pkill(1) not killall(1)

klemensn:
Not all operating systems, e.g. OpenBSD, have killall;  pkill(1) however
should be available on all OSes shipping killall(1).

This makes tests pass on OpenBSD without further changes.


pabuhler:
@klemensn seams to fail on Ubuntu, I guess because pkill also kills the rtpw_test.sh script.

klemensn:
> @klemensn seams to fail on Ubuntu, I guess because pkill also kills the rtpw_test.sh script.
I'm using `-x` now to require a full process name match, let's see if that makes Ubuntu happy.

PR #596, opened 2022-01-31
Speed up SRTP driver tests

bifurcation:
This PR removes a bunch of unnecessary code to make the SRTP test driver 10x-100x faster.  Two major changes:

1. *Make receiver ROC testing faster* - In `test_set_receiver_roc()`, there's a loop protecting a bunch of packets to advance the sender's ROC (since we don't want to rely on setting the sender ROC directly).  Instead of advancing the sequence number by each time, we can encrypt an exponentially smaller number of packets (`O(log N)` instead of `O(n)`) by starting with a bigger step.

2. *Remove test redundancy* - `srtp_test_set_receiver_roc_then_rollover()` has essentially the same content a call to `test_set_receiver_roc()`.  The only difference is that the sequence number in the priming sequence starts at `0xffff` instead of `0x0000`, but given the first change, it's now cheap just to do the comparable call to `test_set_receiver_roc()`.  So we make that substitution.

As a result, on my MacBook, measuring with `time`:

```
Before        internal   1.46s user 0.03s system 84% cpu 1.751 total
Before        OpenSSL    0.33s user 0.00s system 75% cpu 0.439 total
After (1)     internal   0.21s user 0.00s system 73% cpu 0.283 total
After (1)     OpenSSL    0.05s user 0.00s system 37% cpu 0.142 total
After (1+2)   internal   0.01s user 0.00s system 71% cpu 0.011 total
After (1+2)   OpenSSL    0.01s user 0.00s system  5% cpu 0.147 total
```

### DIFF
--- a/test/srtp_driver.c
+++ b/test/srtp_driver.c
@@ -3517,7 +3517,7 @@ static srtp_err_status_t test_set_receiver_roc(uint32_t packets,
         }
 
         while (stride > (packets - i) && stride > 1) {
-          stride >>= 1;
+            stride >>= 1;
         }
 
         i += stride;
--- a/test/srtp_driver.c
+++ b/test/srtp_driver.c
@@ -99,8 +99,6 @@ srtp_err_status_t srtp_test_get_roc(void);
 
 srtp_err_status_t srtp_test_set_receiver_roc(void);
 
-srtp_err_status_t srtp_test_set_receiver_roc_then_rollover(void);
-
 srtp_err_status_t srtp_test_set_sender_roc(void);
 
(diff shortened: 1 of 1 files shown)


test/srtp_driver.c
```diff
@@ -3523,7 +3516,12 @@ static srtp_err_status_t test_set_receiver_roc(uint32_t packets,
             return status;
         }
 
-        seq++;
+        while (stride > (packets - i) && stride > 1) {
+          stride >>= 1;
```
pabuhler, line 3520:
```suggestion
            stride >>= 1;
```

Issue #599, opened 2022-04-03
Support for OpenSSL 3.0


nazar-pc:
In 3.0 some APIs were deprecated that results in this warnings:
```
[1025/1321] Compiling C object subprojects/libsrtp-2.4.2/libsrtp2.a.p/crypto_hash_hmac_ossl.c.o
../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c: In function ‘srtp_hmac_alloc’:
../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c:106:5: warning: ‘HMAC_CTX_new’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  106 |     (*a)->state = HMAC_CTX_new();
      |     ^
In file included from ../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c:54:
../../../subprojects/openssl-3.0.2/include/openssl/hmac.h:33:33: note: declared here
   33 | OSSL_DEPRECATEDIN_3_0 HMAC_CTX *HMAC_CTX_new(void);
      |                                 ^~~~~~~~~~~~
../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c: In function ‘srtp_hmac_dealloc’:
../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c:136:5: warning: ‘HMAC_CTX_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  136 |     HMAC_CTX_free(hmac_ctx);
      |     ^~~~~~~~~~~~~
In file included from ../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c:54:
../../../subprojects/openssl-3.0.2/include/openssl/hmac.h:35:28: note: declared here
   35 | OSSL_DEPRECATEDIN_3_0 void HMAC_CTX_free(HMAC_CTX *ctx);
      |                            ^~~~~~~~~~~~~
../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c: In function ‘srtp_hmac_start’:
../../../subprojects/libsrtp-2.4.2/crypto/hash/hmac_ossl.c:152:5: warning: ‘HMAC_Init_ex’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
[...]

PR #602, opened 2022-05-02
Openssl EVP MAC

pabuhler:
This is a fix for issue #599 .

With OpenSSL 3 the HAMC CTX api has been deprecated and
EVP MAC should be used instead. This change adds a compile
time check on OpeenSSL version and will use the new api if the
version is greater than or equal to 3 .

As part of this change cmake CI build now has warnings as errors
in order to catch the deprecated api warning. There will be a follow
up PR that also adds warnings as errors to windows as well as enabling
more warnings.

### DIFF
--- a/crypto/hash/hmac_ossl.c
+++ b/crypto/hash/hmac_ossl.c
@@ -54,6 +54,9 @@
 
 #if defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
 #define SRTP_OSSL_USE_EVP_MAC
+/* before this version reinit of EVP_MAC_CTX was not supported so need to
+ * duplicate the CTX each time */
+#define SRTP_OSSL_MIN_REINIT_VERSION 0x30000030L
 #endif
 
 #ifndef SRTP_OSSL_USE_EVP_MAC
@@ -73,6 +76,8 @@ typedef struct {
 #ifdef SRTP_OSSL_USE_EVP_MAC
     EVP_MAC *mac;
     EVP_MAC_CTX *ctx;
+    int use_dup;
+    EVP_MAC_CTX *ctx_dup;
 #else
     HMAC_CTX *ctx;
 #endif
@@ -107,6 +112,7 @@ static srtp_err_status_t srtp_hmac_alloc(srtp_auth_t **a,
         *a = NULL;
         return srtp_err_status_alloc_fail;
     }
+    memset(hmac, 0, sizeof(srtp_hmac_ossl_ctx_t));
 
 #ifdef SRTP_OSSL_USE_EVP_MAC
     hmac->mac = EVP_MAC_fetch(NULL, "HMAC", NULL);
@@ -118,6 +124,10 @@ static srtp_err_status_t srtp_hmac_alloc(srtp_auth_t **a,
     }
 
     hmac->ctx = EVP_MAC_CTX_new(hmac->mac);
+
+    hmac->use_dup =
+        OpenSSL_version_num() < SRTP_OSSL_MIN_REINIT_VERSION ? 1 : 0;
+
 #else
     hmac->ctx = HMAC_CTX_new();
 #endif
@@ -147,7 +157,12 @@ static srtp_err_status_t srtp_hmac_dealloc(srtp_auth_t *a)
 
     if (hmac) {
 #ifdef SRTP_OSSL_USE_EVP_MAC
-        EVP_MAC_CTX_free(hmac->ctx);
+        if (hmac->ctx_dup) {
+            EVP_MAC_CTX_free(hmac->ctx_dup);
+        }
+        if (hmac->ctx) {
+            EVP_MAC_CTX_free(hmac->ctx);
+        }
         EVP_MAC_free(hmac->mac);
 #else
         HMAC_CTX_free(hmac->ctx);
@@ -172,8 +187,18 @@ static srtp_err_status_t srtp_hmac_start(void *statev)
     srtp_hmac_ossl_ctx_t *hmac = (srtp_hmac_ossl_ctx_t *)statev;
 
 #ifdef SRTP_OSSL_USE_EVP_MAC
-    if (EVP_MAC_init(hmac->ctx, NULL, 0, NULL) == 0)
-        return srtp_err_status_auth_fail;
+    if (hmac->use_dup) {
+        if (hmac->ctx_dup) {
+            EVP_MAC_CTX_free(hmac->ctx_dup);
+        }
+        hmac->ctx_dup = EVP_MAC_CTX_dup(hmac->ctx);
+        if (hmac->ctx_dup == NULL) {
+            return srtp_err_status_alloc_fail;
+        }
+    } else {
+        if (EVP_MAC_init(hmac->ctx, NULL, 0, NULL) == 0)
+            return srtp_err_status_auth_fail;
+    }
 #else
     if (HMAC_Init_ex(hmac->ctx, NULL, 0, NULL, NULL) == 0)
         return srtp_err_status_auth_fail;
@@ -212,7 +237,8 @@ static srtp_err_status_t srtp_hmac_update(void *statev,
                 srtp_octet_string_hex_string(message, msg_octets));
 
 #ifdef SRTP_OSSL_USE_EVP_MAC
-    if (EVP_MAC_update(hmac->ctx, message, msg_octets) == 0)
+    if (EVP_MAC_update(hmac->use_dup ? hmac->ctx_dup : hmac->ctx, message,
+                       msg_octets) == 0)
         return srtp_err_status_auth_fail;
 #else
     if (HMAC_Update(hmac->ctx, message, msg_octets) == 0)
@@ -246,10 +272,12 @@ static srtp_err_status_t srtp_hmac_compute(void *statev,
 
     /* hash message, copy output into H */
 #ifdef SRTP_OSSL_USE_EVP_MAC
-    if (EVP_MAC_update(hmac->ctx, message, msg_octets) == 0)
+    if (EVP_MAC_update(hmac->use_dup ? hmac->ctx_dup : hmac->ctx, message,
+                       msg_octets) == 0)
         return srtp_err_status_auth_fail;
 
-    if (EVP_MAC_final(hmac->ctx, hash_value, &len, sizeof hash_value) == 0)
+    if (EVP_MAC_final(hmac->use_dup ? hmac->ctx_dup : hmac->ctx, hash_value,
+                      &len, sizeof hash_value) == 0)
         return srtp_err_status_auth_fail;
 #else
     if (HMAC_Update(hmac->ctx, message, msg_octets) == 0)
--- a/crypto/hash/hmac_ossl.c
+++ b/crypto/hash/hmac_ossl.c
@@ -72,6 +72,19 @@ srtp_debug_module_t srtp_mod_hmac = {
     "hmac sha-1 openssl" /* printable name for module   */
 };
 
+/*
+ * There are three different behaviors of OpenSSL HMAC for different versions.
+ *
+ * 1. Pre-3.0 - Use HMAC API
+ * 2. 3.0.0 - 3.0.2 - EVP API is required, but doesn't support reinitialization,
+ *    so we have to use EVP_MAC_CTX_dup
+ * 3. 3.0.3 and later - EVP API is required and supports reinitialization
+ *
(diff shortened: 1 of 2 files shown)


pabuhler:
There seams to be a difference between OpenSSL 3.0.2 and OpenSSL 3.0.3-dev when it comes to EVP_MAC_CTX reuse. The changes here work with 3.0.3 but not 3.0.2, need to investigate why.

JonathanLennox:
> [openssl/openssl#17811](https://github.com/openssl/openssl/issues/17811) --- looks like need to either store the key or else only support 3.0.3 and greater :( . @bifurcation thoughts ?

Storing the key and re-applying it will likely be a noticeable performance regression vs. using the HMAC_CTX APIs, since re-applying the key re-computes the HMAC's i_ctx and o_ctx.

pabuhler:
@JonathanLennox  in that case I would think we need to wait for 3.0.3 release, and for now have the code behind a configure flag or something similar with a min version check.  Far more work than was planned.

JonathanLennox:
Couldn't the ifdef that sets `SRTP_OSSL_USE_EVP_MAC` be conditional on being OpenSSL 3.0.3 or later, and just use the HMAC-flavored functions for OpenSSL <= 3.0.2?

You'd have to remove the `-Werror` addition, but everything else should be fine.  (Alternately it may be possible to set OPENSSL_NO_DEPRECATED_3_0 for OpenSSL versions between 3 and 3.0.2, but I'm quite quite clear on how that works.)

Unless you're worried about building with 3.0.3+, and then running with 3.0.2-, which I guess could be possible.

pabuhler:
> Unless you're worried about building with 3.0.3+, and then running with 3.0.2-, which I guess could be possible.

I think this is the biggest problem,  do not wish to have runtime checks on a patch version.
I think for now will add an ENABLE_OPENSSL3  compile flag and if that is set verify the version is 3.0.3+ for compiling. The thought is if this is set then it is the responsibility of the developer to run against the right version.
Such behavior change in OpenSSL really should be with a minor version change so we could just require 3.1 and be happy.

JonathanLennox:
The other possibility - which I think should work on 3.0.2 without a performance penalty, though I haven't tried it - would be to have srtp_hmac_alloc store a "pristine" MAC ctx which has had its key initialized, and then in srtp_hmac_start construct a "working" MAC ctx that's used for the actual calculations, using `EVP_MAC_CTX_dup()`.

pabuhler:
@JonathanLennox  I also read that using dup was a possibility with little impact. I will try it out

pabuhler:
@JonathanLennox  I added the dup function but the performance hit is noticeable. It will only use this based on a runtime check of the version otherwise it will reinit the CTX. Not sure if all this was worth it to avoid compiler warnings.

JonathanLennox:
> @JonathanLennox I added the dup function but the performance hit is noticeable. It will only use this based on a runtime check of the version otherwise it will reinit the CTX.

I assume dup is still faster than rekey, though.

> Not sure if all this was worth it to avoid compiler warnings.

You'll notice I said that in https://github.com/cisco/libsrtp/issues/599. 🙂

pabuhler:
@JonathanLennox, I think the performance hit is great enough to not want to use dup and since 3.0.3 is not yet released there is no point in this commit yet. For now will just avoid the warning. Was still a good learning exercise.

pabuhler:
With git hub going to Ubuntu 22.04 as latest OpensSSL3 will become the default, so maybe time to get this in.
OpenSSL3 is now at 3.0.7 which includes the fix to reintialize a EVP MAC context so the chance that libsrtp will be running against 3.0.2 or less and getting the performance hit of doing a context dup is very small.
@bifurcation  & @JonathanLennox could you guys chime in on this and review the code. I have reset this branch to the tip of mina and reapplied the 3 commits needed.

crypto/hash/hmac_ossl.c
```diff
@@ -118,38 +184,66 @@ static srtp_err_status_t srtp_hmac_dealloc(srtp_auth_t *a)
...
+        if (hmac->ctx_dup) {
+            EVP_MAC_CTX_free(hmac->ctx_dup);
+        }
+        hmac->ctx_dup = EVP_MAC_CTX_dup(hmac->ctx);
+        if (hmac->ctx_dup == NULL) {
+            return srtp_err_status_alloc_fail;
+        }
+    } else {
+        if (EVP_MAC_init(hmac->ctx, NULL, 0, NULL) == 0)
+            return srtp_err_status_auth_fail;
```
bifurcation, line 200:
```suggestion
        if (EVP_MAC_init(hmac->ctx, NULL, 0, NULL) == 0) {
            return srtp_err_status_auth_fail;
        }
```

Always use curly braces around `if` statements, here and elsewhere.

crypto/hash/hmac_ossl.c
```diff
@@ -118,38 +184,66 @@ static srtp_err_status_t srtp_hmac_dealloc(srtp_auth_t *a)
...
-    HMAC_CTX *state = (HMAC_CTX *)statev;
+    srtp_hmac_ossl_ctx_t *hmac = (srtp_hmac_ossl_ctx_t *)statev;
 
     debug_print(srtp_mod_hmac, "input: %s",
                 srtp_octet_string_hex_string(message, msg_octets));
 
-    if (HMAC_Update(state, message, msg_octets) == 0)
+#ifdef SRTP_OSSL_USE_EVP_MAC
+    if (EVP_MAC_update(hmac->use_dup ? hmac->ctx_dup : hmac->ctx, message,
+                       msg_octets) == 0)
```
bifurcation, line 241:
```suggestion
    if (EVP_MAC_update(hmac->ctx, message, msg_octets) == 0)
```

(Goes with the suggestion above to swap the roles of `ctx` and `ctx_dup`.)

crypto/hash/hmac_ossl.c
```diff
@@ -173,12 +271,21 @@ static srtp_err_status_t srtp_hmac_compute(void *statev,
     }
 
     /* hash message, copy output into H */
-    if (HMAC_Update(state, message, msg_octets) == 0)
+#ifdef SRTP_OSSL_USE_EVP_MAC
+    if (EVP_MAC_update(hmac->use_dup ? hmac->ctx_dup : hmac->ctx, message,
+                       msg_octets) == 0)
```
bifurcation, line 276:
```suggestion
    if (EVP_MAC_update(hmac->ctx, message, msg_octets) == 0)
```

crypto/hash/hmac_ossl.c
```diff
@@ -173,12 +271,21 @@ static srtp_err_status_t srtp_hmac_compute(void *statev,
...
         return srtp_err_status_auth_fail;
 
-    if (HMAC_Final(state, hash_value, &len) == 0)
+    if (EVP_MAC_final(hmac->use_dup ? hmac->ctx_dup : hmac->ctx, hash_value,
+                      &len, sizeof hash_value) == 0)
```
bifurcation, line 280:
```suggestion
    if (EVP_MAC_final(hmac->ctx, hash_value, &len, sizeof hash_value) == 0)
```

crypto/hash/hmac_ossl.c
```diff
@@ -83,14 +105,44 @@ static srtp_err_status_t srtp_hmac_alloc(srtp_auth_t **a,
...
-    (*a)->state = HMAC_CTX_new();
-    if ((*a)->state == NULL) {
+    hmac =
+        (srtp_hmac_ossl_ctx_t *)srtp_crypto_alloc(sizeof(srtp_hmac_ossl_ctx_t));
+    if (hmac == NULL) {
+        srtp_crypto_free(*a);
+        *a = NULL;
+        return srtp_err_status_alloc_fail;
+    }
+    memset(hmac, 0, sizeof(srtp_hmac_ossl_ctx_t));
```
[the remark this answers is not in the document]
pabuhler: looking at srtp_crypto_alloc it already does :) , it use calloc.  Will remove this.

crypto/hash/hmac_ossl.c
```diff
@@ -118,38 +184,66 @@ static srtp_err_status_t srtp_hmac_dealloc(srtp_auth_t *a)
...
+
+#ifdef SRTP_OSSL_USE_EVP_MAC
+    if (hmac->use_dup) {
+        if (hmac->ctx_dup) {
+            EVP_MAC_CTX_free(hmac->ctx_dup);
+        }
+        hmac->ctx_dup = EVP_MAC_CTX_dup(hmac->ctx);
+        if (hmac->ctx_dup == NULL) {
+            return srtp_err_status_alloc_fail;
+        }
```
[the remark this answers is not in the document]
pabuhler: I did look at doing something like this before but it means changes are needed in the construction and init functions so in the end I thought this was the simpler way.  Could encapsulate the logic in a simple get_ctx() function or macro to make it ready better. What do you think?

crypto/hash/hmac_ossl.c
```diff
@@ -51,7 +51,17 @@
 #include "err.h" /* for srtp_debug */
 #include "auth_test_cases.h"
 #include <openssl/evp.h>
+
+#if defined(OPENSSL_VERSION_MAJOR) && (OPENSSL_VERSION_MAJOR >= 3)
+#define SRTP_OSSL_USE_EVP_MAC
+/* before this version reinit of EVP_MAC_CTX was not supported so need to
+ * duplicate the CTX each time */
```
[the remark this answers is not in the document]
pabuhler: Done

crypto/hash/hmac_ossl.c
```diff
@@ -101,11 +174,25 @@ static srtp_err_status_t srtp_hmac_alloc(srtp_auth_t **a,
...
-    hmac_ctx = (HMAC_CTX *)a->state;
+    srtp_hmac_ossl_ctx_t *hmac = (srtp_hmac_ossl_ctx_t *)a->state;
+
+    if (hmac) {
+#ifdef SRTP_OSSL_USE_EVP_MAC
+        if (hmac->ctx != NULL) {
+            EVP_MAC_CTX_free(hmac->ctx);
+        }
+        if (hmac->ctx_dup != NULL) {
+            EVP_MAC_CTX_free(hmac->ctx_dup);
```
bifurcation, line 185: [The documentation](https://www.openssl.org/docs/man3.0/man3/EVP_MAC_CTX_free.html) says "NULL is a valid parameter, for which this function is a no-op.", so you don't need the `if` branch here.

crypto/hash/hmac_ossl.c
```diff
@@ -118,38 +205,69 @@ static srtp_err_status_t srtp_hmac_dealloc(srtp_auth_t *a)
...
 {
-    HMAC_CTX *state = (HMAC_CTX *)statev;
-
-    if (HMAC_Init_ex(state, NULL, 0, NULL, NULL) == 0)
+    srtp_hmac_ossl_ctx_t *hmac = (srtp_hmac_ossl_ctx_t *)statev;
+
+#ifdef SRTP_OSSL_USE_EVP_MAC
+    if (hmac->use_dup) {
+        if (hmac->ctx != NULL) {
+            EVP_MAC_CTX_free(hmac->ctx);
```
bifurcation, line 213: As above, this doesn't need to be conditional.

crypto/hash/hmac_ossl.c
```diff
@@ -118,38 +205,69 @@ static srtp_err_status_t srtp_hmac_dealloc(srtp_auth_t *a)
...
-    if (HMAC_Init_ex(state, key, key_len, EVP_sha1(), NULL) == 0)
-        return srtp_err_status_auth_fail;
+#ifdef SRTP_OSSL_USE_EVP_MAC
+    OSSL_PARAM params[2];
+
+    params[0] = OSSL_PARAM_construct_utf8_string("digest", "SHA1", 0);
+    params[1] = OSSL_PARAM_construct_end();
 
+    if (EVP_MAC_init(hmac->use_dup ? hmac->ctx_dup : hmac->ctx, key, key_len,
+                     params) == 0) {
```
bifurcation, line 244:
```suggestion
    if (EVP_MAC_init(hmac->ctx, key, key_len, params) == 0) {
```

bifurcation (approved):
Couple of small things, otherwise LGTM.

crypto/hash/hmac_ossl.c
pabuhler: when using dup, doesn't the key need to be set on the dup_ctx ?  the normal ctx will be null at this point ??

pabuhler: testing shows this to crash, so rejecting for now

PR #603, opened 2022-05-02
Warnings as errors

pabuhler:
This makes the cmake CI build stricter by add more warnings and forcing them all to be fixed.
For now warnings as errors enabled by default with cmake, could be optional in the future.

Follow up from change first introduced in #602

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,7 @@ set(ENABLE_MBEDTLS OFF CACHE BOOL "Enable MbedTLS crypto engine")
 set(ENABLE_NSS OFF CACHE BOOL "Enable NSS crypto engine")
 set(TEST_APPS ON CACHE BOOL "Build test applications")
 set(BUILD_WITH_SANITIZERS OFF CACHE BOOL "Adds sanitizer flags to compiler")
+set(BUILD_WITH_WARNINGS ON CACHE BOOL "Adds warnings and warning as error flags to compiler")
 option(BUILD_SHARED_LIBS "Build shared library" OFF)
 
 if(ENABLE_OPENSSL OR ENABLE_MBEDTLS OR ENABLE_NSS)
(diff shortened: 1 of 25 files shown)


test/srtp_driver.c
```diff
@@ -219,12 +220,13 @@ int main(int argc, char *argv[])
     unsigned do_list_mods = 0;
     unsigned do_log_stdout = 0;
     srtp_err_status_t status;
+    size_t hdr_size = 12;
```
JonathanLennox, line 223: Can be const?

pabuhler: I thought I had tried and it failed, will double check.

pabuhler: looks like it worked

crypto/cipher/aes_icm.c
```diff
@@ -98,6 +98,7 @@ static srtp_err_status_t srtp_aes_icm_alloc(srtp_cipher_t **c,
                                             int tlen)
 {
     srtp_aes_icm_ctx_t *icm;
+    (void)tlen;
```
paulej, line 101: If this is just to suppress warnings about unused variables, we could just not give the parameter a name, I think.

CMakeLists.txt
```diff
@@ -116,6 +116,12 @@ if(BUILD_WITH_SANITIZERS AND NOT WIN32)
   set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -fno-omit-frame-pointer -fsanitize=${SANITIZERS}")
 endif()
 
+if(CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_C_COMPILER_ID MATCHES "GNU")
```
paulej, line 119: Would be good if this were controlled by an option to avoid warnings when compilers change and start complaining about things that today's compilers do not complain about.  Or in environments where there are warnings one wants to ignore.

crypto/test/datatypes_driver.c
```diff
@@ -139,12 +139,12 @@ int main(void)
...
 void byte_order(void)
 {
-    int i;
+    size_t i;
     v128_t e;
 
     printf("byte ordering of crypto/math datatypes:\n");
     for (i = 0; i < sizeof(e); i++)
-        e.v8[i] = i;
+        e.v8[i] = (uint8_t)i;
```
paulej, line 147: Since `i` is defined right there in the loop, why not make it the needed type to avoid the cast?

CMakeLists.txt
pabuhler: Will add option but have it default on for now.

crypto/cipher/aes_icm.c
pabuhler: that gives a different warning ;)

PR #604, opened 2022-05-04
Fix compile error due to unused variable

rickwporter:
Without this change, I see the following error:
In file included from test/test_srtp.c:54:
./test/cutest.h:262:16: warning: variable 'n' set but not used [-Wunused-but-set-variable]
        size_t n = 0;
               ^


pabuhler (approved):
This is not really libsrtp code but it is ok to prevent warnings.

PR #605, opened 2022-05-06
Set OpenSSL api compat to 1.1.x for hamc

pabuhler:
An alternative to #602 as a solution to #599.
Since the EVP_MAC_xxx api has some reinitialization issues pre 3.0.3 and the workaround in #603
introduces a performance hit this is probably the best solution until OpenSSL 3.0.3 is released and widely adopted.

### DIFF
--- a/crypto/hash/hmac_ossl.c
+++ b/crypto/hash/hmac_ossl.c
@@ -50,6 +50,7 @@
 #include "alloc.h"
 #include "err.h" /* for srtp_debug */
 #include "auth_test_cases.h"
+#define OPENSSL_API_COMPAT 0x10100000L
 #include <openssl/evp.h>
 #include <openssl/hmac.h>
 
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -12,18 +12,24 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        crypto: [internal, openssl, nss, mbedtls]
+        crypto: [internal, openssl, openssl3, nss, mbedtls]
         exclude:
           - os: windows-latest
             crypto: openssl
           - os: windows-latest
+            crypto: openssl3
+          - os: windows-latest
             crypto: nss
           - os: windows-latest
             crypto: mbedtls
+          - os: ubuntu-latest
+            crypto: openssl3
         include:
           - crypto: internal
             cmake-crypto-enable: ""
           - crypto: openssl
+            cmake-crypto-enable: "-DENABLE_OPENSSL=ON"
+          - crypto: openssl3
             cmake-crypto-enable: "-DENABLE_OPENSSL=ON"
           - crypto: nss
             cmake-crypto-enable: "-DENABLE_NSS=ON"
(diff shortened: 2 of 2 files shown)


crypto/hash/hmac_ossl.c
```diff
@@ -50,6 +50,7 @@
 #include "alloc.h"
 #include "err.h" /* for srtp_debug */
 #include "auth_test_cases.h"
+#define OPENSSL_API_COMPAT 0x10100000L
```
nazar-pc, line 53: It isn't nice to define global things like this in the library and change behavior of the dependency :confused:

This should really only be done in the app.

paulej: This one is tough. I wouldn't personally want to keep track of every dependency of every external project and what version of some third-party's third-party stuff I need to use. I think so long as the expected behavior is achieved, that's what matters.

pabuhler: As I understand this will not have a bearing on the behavior of either the compile time or runtime OpeenSSL library in use. This value is intended to be set by developers using OpeenSSL.
See:
https://www.openssl.org/docs/manmaster/man7/OPENSSL_API_COMPAT.html
and comments at:
https://github.com/openssl/openssl/blob/a6680123643bc3289ecbcbd6bce844a814c1510a/include/openssl/macros.h#L84

This is an indication that libsrtp hmac requires an OpenSSL api compatible with 1.1.x
@nazar-pc   if you disagree then maybe we need to find a 3rd approach

nazar-pc: According to provided links it is not guaranteed that this hack will even work if headers were included by another code in the app already:
> NOTE: to be effective, a user defined macro must be defined before including any header file that depends on it, either in the compilation command (cc -DMACRO=value) or by defining the macro in source before including any headers.

Proper approach is to check OpenSSL version and use API recommended by that version.

pabuhler: @nazar-pc I dont disagree but have you followed #599 & #602 ?  #602  was an attempt to do exactly as you suggest but it results in a runtime version check and until the unreleased OpenSLL 3.0.3 version is widely adopted a substantial performance penalty.

nazar-pc: I did briefly.

I see it only checks for OpenSSL 3.0.0+. Replace with check for 3.0.3+ instead and let it generate warning for 3.0.0..3.0.2 as it did before to avoid performance penalty.

pabuhler: That is kind of OK for compile time and static linking, but I dont see how that would work at run time with shared library. If you build against 3.0.3 the new api would be used but if 3.0.2 was loaded dynamically then either hmac would not work or else there is a performance hit.

nazar-pc: Performance hit for those edge cases is acceptable IMHO. Also shared libraries typically shipped by Linux distros, so libsrtp shouldn't be built against the newer OpenSSL version comparing to one it'll be running against, so I'm not so worried about performance hit.

pabuhler:
Closing this and preferring the solution in #602, although this still would have been an option.

PR #607, opened 2022-06-30
Build without versioned soname on Android.

eliteraspberries:
I'm building for Android with CMake. This produces the shared library `libsrtp2.so.2.4.2` and a symbolic link to it `libsrtp2.so`.

I would like to just drop the library into the `jniLibs` directory, but Android Studio recognizes libraries by file name (ignores file names that don't match *.so), so I need to build libsrtp without a versioned soname, i.e. no linker flag `-Wl,-soname,...`.

This can be done by setting the CMake target property [NO_SONAME](https://cmake.org/cmake/help/latest/prop_tgt/NO_SONAME.html).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,6 +250,9 @@ add_library(srtp2
 )
 
 set_target_properties(srtp2 PROPERTIES VERSION ${CMAKE_PROJECT_VERSION})
+if(${CMAKE_C_COMPILER_TARGET} MATCHES ".*-linux-android.*")
+  set_target_properties(srtp2 PROPERTIES NO_SONAME ON)
+endif()
 
 target_include_directories(srtp2 PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/crypto/include>


pabuhler:
Hi @eliteraspberries  I just tried to build for Android using CMAKE and it did not create a version *.so for me. The output when building shared was just libsrtp2.so .
I have created a pull request with Android build that shows what I did, #628 .
I am going to close this for now, please reopen if you still have issue.

DoAnhKien:
Hi pablhler, can you show me the way that you build libsrtp2.so , i tried so many way but i not work

pabuhler:
Did you look at the steps in https://github.com/cisco/libsrtp/blob/main/.github/workflows/android.yml ? This was just an example and I am not sure how you try to build libsrtp for android. 

Remember you can also just add the libsrtp source files to your own build files.

If you provide a mini project that shows your issues then maybe I can take a look at what goes wrong but I am no Android build expert.

DoAnhKien:
thank pablhler, i did with the docs, then i success with it, thank you

PR #611, opened 2022-09-02
Create alias for srtp2 as libSRTP::srtp2

paulej:
This will add an alias for the srtp2 library so that when using CMake with either `find_library` or `FetchContent`, the use of `target_link_libraries()` can always follow the form `libSRTP::srtp2`.  Without this, using `FetchContent` will allow only the name `srtp2`.


pabuhler (approved):
LGTM,   I am not a cmake guru but from what I read this should be ok and acceptable behavior.

PR #612, opened 2022-09-07
Allow overriding stream list implementation

mildsunrise:
Hello! This PR proposes to refactor the logic managing the `stream_list` (which is currently an intrusive singly linked list) into its own functions:

``` c
srtp_err_status_t srtp_stream_list_create(srtp_stream_list_t *list);
void              srtp_stream_list_insert(srtp_stream_list_t *list, srtp_stream_t stream);
srtp_stream_t     srtp_stream_list_get(srtp_stream_list_t *list, uint32_t ssrc);
srtp_stream_t     srtp_stream_list_delete(srtp_stream_list_t *list, uint32_t ssrc);
void              srtp_stream_list_for_each(srtp_stream_list_t *list, int (*callback)(srtp_stream_t, void *), void *data);
srtp_err_status_t srtp_stream_list_dealloc(srtp_stream_list_t *list, srtp_stream_t template_);
```

These functions are then gated by a define, so that downstreams can remove the default linked list implementation and replace it with one suited to their needs.

More precisely, the linked list is generalized to an opaque API (`stream_list_priv.h`) with only the semantics I thought we'd need (retrieval, insertion, removal, and iteration that allows removing the currently iterating element) documented. It's still an internal API, we're not asking for stability on these semantics or signatures; the only requirement is that `stream_list` must be updated only through this API, and that this file is kept up to date with any semantics the rest of the code relies on.

It's an alternative to #508 that we believe should be more flexible for everybody and hopefully introduce less maintenance burden on your side, as well as avoid changing the current behavior. What are your thoughts on it?
/cc @murillo128


murillo128:
@fluffy @bifurcation could you take a look at this, please?

pabuhler (commented):
In general I think this is a better approach, did a quick review but will try to look in more detail in the next days.

With this API there is the assumption that inserting can never fail, maybe it should return a status.

mildsunrise:
Yes, inserting a duplicate key is defined as UB. This is because in the current code, when inserting a stream we don't check that another one with the same SSRC already exists. Since we make this assumption, in order to keep the current performance it's necessary to allow implementations to assume that as well... otherwise the default implementation for `srtp_stream_list_insert` would now need to iterate through the whole linked list to check. Should the code need a checked insertion, it can always call `srtp_stream_list_get` first to ensure the SSRC doesn't already exist.

This was the reasoning but of course it's easy to change if you prefer a checked insertion, or an advisory returned status

pabuhler:
I was more thinking if it failed due to some capacity issues or something similar in what ever list implementation was used. If the item did not actually get put in to a list it would cause a memory leak.

mildsunrise:
I see, that makes sense. I've amended `stream_list_priv.h` with the new semantics, and added fallback paths in all usages of `srtp_stream_list_insert`. The fallback path just deallocates the stream and aborts. It may sometimes leave the srtp context in a partially modified state, but that's better than a leak and we do it in other places too.

fluffy:
So first, I think refactoring along theses lines sounds like a good idea and makes sense. I would want to be very careful this did not break things and I have not carefully reviewed the code but I do like the basic idea.

fluffy:
I think it we had this, it would  make it much easier to also put the optimized version from 508 into the main repo and not just have it in a downstream.

mildsunrise:
> I would want to be very careful this did not break things and I have not carefully reviewed the code but I do like the basic idea.

Agree. Most of the changes just extract code into the functions, but there are 2 places (feac4ce and 6653a86) where I've had to change behavior a bit to generalize, which need thorough review.

dsdolzhenko:
@mildsunrise Hi! I've tried to implement a custom `stream_list` in a project that uses `libsrtp` with your changes included. However, `stream_list_priv.h` is not part of the public headers set, at least when built with meson as a dependency. Without `srtp_stream_t`'s definition being available in a downstream, it seems to be impossible to create a custom implementation of `stream_list`. Is it expected to be so?

pabuhler:
@dsdolzhenko  I will try the same,  I am not sure that this "feature" will just work using the build files provided with libsrtp out of the box, we would have to think if that is what we would want to support. I have the feeling that to start with this feature is more for custom builds of libsrtp where teams have their own build files for libsrtp.

mildsunrise:
This is the intent, yes. As mentioned this is an internal API and thus doesn't have the stability guarantees that public APIs have, it's targeted towards projects that do custom builds. We can always support more use cases later, but in any case I'd personally leave the changes sitting there for some time before stabilizing.

dsdolzhenko:
Ok, got it. Thank you for the clarifications and for the MR ;)

pabuhler:
Hi @mildsunrise, it has been awhile but I am now trying to get this in. I have made some local api changes, my suggestion is
```
srtp_stream_list_t srtp_stream_list_alloc();
srtp_err_status_t srtp_stream_list_dealloc(srtp_stream_list_t list);
srtp_err_status_t srtp_stream_list_insert(srtp_stream_list_t list, srtp_stream_t stream);
srtp_stream_t srtp_stream_list_get(srtp_stream_list_t list, uint32_t ssrc);
void srtp_stream_list_remove(srtp_stream_list_t list, srtp_stream_t stream);
void srtp_stream_list_for_each(srtp_stream_list_t list, int (*callback)(srtp_stream_t, void *), void *data);
```
and have changed the internal implementation to be a double linked list. I am not sure how you have replaced the api in your code so I am not sure if this will break anything for you.
My motivation was to make it a little more consistent with other LibSRTP api's and have the list have as little knowledge as possible about LibSRTP internals.
If this ok for you I will update this PR and you can have a look in more detail.

pabuhler:
Commit 914cb552fbcc1854f5c88a969eaa34d22377e598 in #632 contains some small proposed changes to this PR.  If they look ok in general then will merge them here for complete review.

@mildsunrise  do you think it is still compatible with want you initially wanted?

bifurcation:
I agree that this seems like a good approach.  Happy to do a detailed review once we make a decision on #632.

Also agree with @fluffy that it would be good to fold in #508 after we merge this.  Together with the fact that people seem to want to do other proprietary implementations, maybe we want three states for the implementation:

* Linked list (default) - the current implementation
* Static array
* Custom (no implementation built, so as not to conflict with custom)

This seems analogous to the crypto library selection, so probably good to do this switching in the same way -- put these in separate files and switch among them using the build system.

mildsunrise:
if I'm not missing anything, changes proposed by @pabuhler in #632 are:

 - `create` renamed to `alloc`, `delete` renamed to `remove`
 - default implementation is now a doubly linked list, so there's an additional intrusive field in streams
 - list implementation now allocates its own storage
 - `dealloc` no longer deallocates streams, instead it checks that list is empty
 - `remove` now takes the stream instead of the key

I'm happy with the changes and yes, this is compatible with what we had.

the only thing I want to point out is the last change, where the stream parameter is effectively an iterator for the internal (intrusive) implementation, but not other implementations. so your code shouldn't expect O(1) removes; if that may be needed, we should extend the API to expose iterators I think.

if not the case then I'm :+1: on merging #632 and closing this

pabuhler:
Thanks @mildsunrise  & @bifurcation ,

@mildsunrise  you summary is correct.
If needed we could also add a take()  function that combines both the get & remove in order to provide fast removal in both the lookup and iterate use cases.
I will merge the changes from #632  in to this PR and than we can try to get a complete review and get this in.

include/stream_list_priv.h
```diff
@@ -0,0 +1,123 @@
...
+ * if no such SSRC is found, NULL is returned.
+ */
+srtp_stream_t srtp_stream_list_get(srtp_stream_list_t list, uint32_t ssrc);
+
+/**
+ * remove the stream from the list.
+ *
+ * ownership is transferred to the caller.
+ *
+ * if the stream is not in the list the behavior is undefined.
```
bifurcation, line 106: Again, it seems like this should error, e.g., with `srtp_error_status_no_context`.

include/stream_list_priv.h
```diff
@@ -0,0 +1,123 @@
...
+ */
+void srtp_stream_list_remove(srtp_stream_list_t list, srtp_stream_t stream);
```
bifurcation, line 108: Considering this method on its own, I would have expected something more like:

```
srtp_stream_t srtp_stream_list_remove(srtp_stream_list_t list, unsigned int ssrc);
```

The API here only really makes sense in combination with `srtp_stream_list_get`, i.e., if the stream pointer came from that method. (It basically *is* that method except for the change of ownership.) We should either document that fact, or make the API more like the above.  My preference would be for the latter, maybe calling it `extract` in parallel with [the comparable C++ STL design pattern](https://en.cppreference.com/w/cpp/container/map/extract).

bifurcation (changes requested):
Partial review here, just covering the high-level design.

.github/workflows/cmake.yml
```diff
@@ -77,3 +77,11 @@ jobs:
       working-directory: ${{github.workspace}}/build
       shell: bash
       run: ctest
+
+    - name: Test Stream List Replacement
```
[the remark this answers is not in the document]
pabuhler: Yeah moving it to a new workflow is probably the best, this was in a way just proving that it could be done. I will move it and split the steps up.
Not sure why you would like it changed to "Statically Allocated"  there is no static allocations in the test. The test is simply to prove that it is possible to replace the internal stream list implementation with a different one by using the SRTP_NO_STREAM_LIST define.

include/stream_list_priv.h
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * srtp_stream_list_t holds a list of srtp_stream_t, each identified
+ * by their SSRC.
+ *
+ * the API was extracted to allow downstreams to override its
+ * implementation by defining the `SRTP_NO_STREAM_LIST` preprocessor
+ * directive, which removes the default implementation of these
+ * functions. if this is done, the `next` & `prev` fields are free for
+ * the implementation to use.
```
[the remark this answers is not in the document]
pabuhler: In my mind that would be a natural progression of this PR, not necessarily the purpose of this PR. We can consider going the full steep of making stream & stream list 100% independent of each other now if that is preferred. Will be a bit more work but will result in a clean change in the end.

include/stream_list_priv.h
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * insert a stream into the list
+ *
+ * returns srtp_err_status_alloc_fail if insertion failed due to unavailable
+ * capacity in the list. if operation succeeds, srtp_err_status_ok is returned
+ * and ownership is transferred from caller into the list
+ *
+ * if another stream with the same SSRC already exists in the list,
+ * behavior is undefined. if the SSRC field is mutated while the
+ * stream is inserted, further operations have undefined behavior
```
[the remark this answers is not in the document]
pabuhler: My guess here is that if the list is required to do this check then it will need to perform an iterate or lookup operation at a cost, and presumable that was already done by libsrtp before inserting. So for performance reason the list is a "dumb" data structure and the job of managing the contents is left up libsrtp.

include/stream_list_priv.h
pabuhler: I was a bit unsure here, since this should never happen unless there is a bug in libsrtp, but may be being defensive is better. I can add it.

include/stream_list_priv.h
```diff
@@ -0,0 +1,123 @@
...
+/*
+ * look up the stream corresponding to the specified SSRC and return it.
+ * if no such SSRC is found, NULL is returned.
+ */
+srtp_stream_t srtp_stream_list_get(srtp_stream_list_t list, uint32_t ssrc);
+
+/**
+ * remove the stream from the list.
+ *
+ * ownership is transferred to the caller.
```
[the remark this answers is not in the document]
pabuhler: same as above , I dont agree with the ownership statements so might just remove them

include/stream_list_priv.h
pabuhler: That is what it was before my change, but it is also used inside of a srtp_stream_list_for_each loop in which case the caller already has  the srtp_stream_t so this was to prevent a second lookup. This kind of goes back to what is this PR trying to achieve, minimal change or complete separation.

include/stream_list_priv.h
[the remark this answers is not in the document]
mildsunrise: removing `next` and `prev` from `srtp_stream_ctx_t` would mean the default impl should have to change from being an intrusive linked list to a non-intrusive one, with the performance implications that come with it. so I agree with @pabuhler that this can be done later. we can remove the "are free for the implementation to use" bit if you want

include/stream_list_priv.h
```diff
@@ -0,0 +1,123 @@
...
+ * the list must be empty or else an error is returned.
+ */
+srtp_err_status_t srtp_stream_list_dealloc(srtp_stream_list_t list);
+
+/**
+ * insert a stream into the list
+ *
+ * returns srtp_err_status_alloc_fail if insertion failed due to unavailable
+ * capacity in the list. if operation succeeds, srtp_err_status_ok is returned
+ * and ownership is transferred from caller into the list
```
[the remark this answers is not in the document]
mildsunrise: now that the list won't ever free streams, the comment can be removed yes

include/stream_list_priv.h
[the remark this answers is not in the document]
mildsunrise: > if another stream with the same SSRC already exists in the list, behavior is undefined

this is here solely because of the linked list default implementation, where checking for collisions would make insertions O(n). I also don't think you can specify it as "overwrite", because the caller would still expect that after removing the (duplicated) key, the key would no longer exist in the list. so I see no way of specifying collision behavior without making insertions or removals O(n)... but if you prefer that, or have some other solution, I'll happily accept it

include/stream_list_priv.h
mildsunrise: also in favor of returning an error rather than UB

include/stream_list_priv.h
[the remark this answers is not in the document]
bifurcation: Thanks, this makes sense.  I keep having to remember that this is not a real abstraction, just an internal boundary :)

include/stream_list_priv.h
bifurcation: No-op seems fine as well.

include/stream_list_priv.h
bifurcation: I'm fine with not changing the interface.  I'll suggest some documentation above.

include/stream_list_priv.h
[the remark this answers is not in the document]
bifurcation:
```suggestion
 * The stream to be removed is referenced "by value", i.e., by the pointer to be
 * removed from the list. This pointer is obtained using `srtp_stream_list_get`
 * or as callback parameter in `srtp_stream_list_for_each`.
```

srtp/srtp.c
```diff
@@ -279,6 +280,46 @@ static srtp_err_status_t srtp_stream_dealloc(
...
 
+/* try to insert stream in list or deallocate it */
+static srtp_err_status_t srtp_insert_or_dealloc_stream(srtp_stream_list_t list,
+                                                       srtp_stream_t stream,
+                                                       srtp_stream_t template)
+{
+    srtp_err_status_t status = srtp_stream_list_insert(list, stream);
+    /* on failure, ownership wasn't transferred and we need to deallocate */
+    if (status)
+        srtp_stream_dealloc(stream, template);
```
bifurcation, line 291:
```suggestion
    if (status) {
        srtp_stream_dealloc(stream, template);
    }
```

bifurcation: ... and please fix similar instances below.

srtp/srtp.c
```diff
@@ -2058,9 +2100,11 @@ static srtp_err_status_t srtp_unprotect_aead(srtp_ctx_t *ctx,
             return status;
         }
 
-        /* add new stream to the head of the stream_list */
-        new_stream->next = ctx->stream_list;
-        ctx->stream_list = new_stream;
+        /* add new stream to the list */
+        status = srtp_insert_or_dealloc_stream(ctx->stream_list, new_stream,
+                                               ctx->stream_template);
+        if (status)
```
bifurcation, line 2106: Missing brackets.

srtp/srtp.c
```diff
@@ -2149,9 +2193,11 @@ srtp_err_status_t srtp_protect_mki(srtp_ctx_t *ctx,
             if (status)
                 return status;
 
-            /* add new stream to the head of the stream_list */
-            new_stream->next = ctx->stream_list;
-            ctx->stream_list = new_stream;
+            /* add new stream to the list */
+            status = srtp_insert_or_dealloc_stream(ctx->stream_list, new_stream,
+                                                   ctx->stream_template);
+            if (status)
```
bifurcation, line 2199: Missing brackets.

srtp/srtp.c
```diff
@@ -2748,9 +2794,11 @@ srtp_err_status_t srtp_unprotect_mki(srtp_ctx_t *ctx,
         if (status)
             return status;
 
-        /* add new stream to the head of the stream_list */
-        new_stream->next = ctx->stream_list;
-        ctx->stream_list = new_stream;
+        /* add new stream to the list */
+        status = srtp_insert_or_dealloc_stream(ctx->stream_list, new_stream,
+                                               ctx->stream_template);
+        if (status)
```
bifurcation, line 2800: Missing brackets

srtp/srtp.c
```diff
@@ -2842,15 +2871,11 @@ srtp_err_status_t srtp_dealloc(srtp_t session)
...
-        srtp_stream_t next = stream->next;
-        status = srtp_stream_dealloc(stream, session->stream_template);
-        if (status)
-            return status;
-        stream = next;
-    }
+    /* deallocate streams */
+    status = srtp_remove_and_dealloc_streams(session->stream_list,
+                                             session->stream_template);
+    if (status)
```
bifurcation, line 2877: Missing brackets

srtp/srtp.c
```diff
@@ -2919,8 +2950,10 @@ srtp_err_status_t srtp_add_stream(srtp_t session, const srtp_policy_t *policy)
         session->stream_template->direction = dir_srtp_receiver;
         break;
     case (ssrc_specific):
-        tmp->next = session->stream_list;
-        session->stream_list = tmp;
+        status = srtp_insert_or_dealloc_stream(session->stream_list, tmp,
+                                               session->stream_template);
+        if (status)
```
bifurcation, line 2955: Missing brackets

srtp/srtp.c
```diff
@@ -2977,28 +3020,19 @@ srtp_err_status_t srtp_create(srtp_t *session, /* handle for session     */
...
         return srtp_err_status_no_ctx;
-
-    /* remove stream from the list */
-    if (last_stream == stream)
-        /* stream was first in list */
-        session->stream_list = stream->next;
-    else
-        last_stream->next = stream->next;
+    }
+    srtp_stream_list_remove(session->stream_list, stream);
```
bifurcation, line 3035:
```suggestion

    srtp_stream_list_remove(session->stream_list, stream);
```

srtp/srtp.c
```diff
@@ -3035,12 +3069,66 @@ srtp_err_status_t srtp_update(srtp_t session, const srtp_policy_t *policy)
...
+    srtp_xtd_seq_num_t old_index;
+    srtp_rdb_t old_rtcp_rdb;
+
+    /* old / non-template streams are copied unchanged */
+    if (stream->session_keys[0].rtp_auth !=
+        session->stream_template->session_keys[0].rtp_auth) {
+        srtp_stream_list_remove(session->stream_list, stream);
+        data->status = srtp_insert_or_dealloc_stream(
+            data->new_stream_list, stream, session->stream_template);
+        if (data->status)
```
bifurcation, line 3094: Missing brackets

srtp/srtp.c
```diff
@@ -3035,12 +3069,66 @@ srtp_err_status_t srtp_update(srtp_t session, const srtp_policy_t *policy)
...
+        return 0;
+    }
+
+    /* save old extendard seq */
+    old_index = stream->rtp_rdbx.index;
+    old_rtcp_rdb = stream->rtcp_rdb;
+
+    /* remove stream */
+    data->status = srtp_remove_stream(session, ssrc);
+    if (data->status)
```
bifurcation, line 3105: Missing brackets

srtp/srtp.c
```diff
@@ -3035,12 +3069,66 @@ srtp_err_status_t srtp_update(srtp_t session, const srtp_policy_t *policy)
...
+        return 1;
+
+    /* allocate and initialize a new stream */
+    data->status = srtp_stream_clone(data->new_stream_template, ssrc, &stream);
+    if (data->status)
```
bifurcation, line 3110: Missing brackets

srtp/srtp.c
```diff
@@ -3035,12 +3069,66 @@ srtp_err_status_t srtp_update(srtp_t session, const srtp_policy_t *policy)
...
+        return 1;
+
+    /* add new stream to the head of the new_stream_list */
+    data->status = srtp_insert_or_dealloc_stream(data->new_stream_list, stream,
+                                                 data->new_stream_template);
+    if (data->status)
```
bifurcation, line 3116: Missing brackets

srtp/srtp.c
```diff
@@ -3064,77 +3152,36 @@ static srtp_err_status_t update_template_streams(srtp_t session,
...
+        return data.status;
+    }
+
+    /* dealloc old list / template */
+    srtp_remove_and_dealloc_streams(session->stream_list,
+                                    session->stream_template);
+    srtp_stream_list_dealloc(session->stream_list);
     srtp_stream_dealloc(session->stream_template, NULL);
-    /* set new template */
+    /* set new list / template */
```
bifurcation, line 3181:
```suggestion

    /* set new list / template */
```

srtp/srtp.c
```diff
@@ -3902,9 +3949,11 @@ static srtp_err_status_t srtp_unprotect_rtcp_aead(
             return status;
         }
 
-        /* add new stream to the head of the stream_list */
-        new_stream->next = ctx->stream_list;
-        ctx->stream_list = new_stream;
+        /* add new stream to the list */
+        status = srtp_insert_or_dealloc_stream(ctx->stream_list, new_stream,
+                                               ctx->stream_template);
+        if (status)
```
bifurcation, line 3955: Missing brackets

srtp/srtp.c
```diff
@@ -3968,9 +4017,11 @@ srtp_err_status_t srtp_protect_rtcp_mki(srtp_t ctx,
             if (status)
                 return status;
 
-            /* add new stream to the head of the stream_list */
-            new_stream->next = ctx->stream_list;
-            ctx->stream_list = new_stream;
+            /* add new stream to the list */
+            status = srtp_insert_or_dealloc_stream(ctx->stream_list, new_stream,
+                                                   ctx->stream_template);
+            if (status)
```
bifurcation, line 4023: Missing brackets

srtp/srtp.c
```diff
@@ -4415,9 +4466,11 @@ srtp_err_status_t srtp_unprotect_rtcp_mki(srtp_t ctx,
         if (status)
             return status;
 
-        /* add new stream to the head of the stream_list */
-        new_stream->next = ctx->stream_list;
-        ctx->stream_list = new_stream;
+        /* add new stream to the list */
+        status = srtp_insert_or_dealloc_stream(ctx->stream_list, new_stream,
+                                               ctx->stream_template);
+        if (status)
```
bifurcation, line 4472: Missing brackets

srtp/srtp.c
```diff
@@ -4598,45 +4651,62 @@ srtp_err_status_t stream_get_protect_trailer_length(srtp_stream_ctx_t *stream,
...
+    uint32_t temp_length;
+
+    data->valid = 1;
+    if (stream_get_protect_trailer_length(stream, data->is_rtp, data->use_mki,
+                                          data->mki_index,
+                                          &temp_length) == srtp_err_status_ok) {
+        if (temp_length > data->length) {
+            data->length = temp_length;
+        }
+    }
```
bifurcation, line 4675:
```suggestion
    if (stream_get_protect_trailer_length(stream, data->is_rtp, data->use_mki,
                                          data->mki_index,
                                          &temp_length) == srtp_err_status_ok) {
        data->valid = 1;
        if (temp_length > data->length) {
            data->length = temp_length;
        }
    }
```

bifurcation: This just moves the assignment to `data->valid` within the `if`, so that `valid` indicates that you got data from at least one stream.

srtp/srtp.c
```diff
@@ -4598,45 +4651,62 @@ srtp_err_status_t stream_get_protect_trailer_length(srtp_stream_ctx_t *stream,
     return srtp_err_status_ok;
 }
 
+struct get_protect_trailer_length_data {
+    int valid;
+    uint32_t length;
```
bifurcation, line 4656:
```suggestion
    uint32_t found_stream; /* whether at least one matching stream was found */
    uint32_t length;       /* maximum trailer length found so far */
```

srtp/srtp.c
```diff
@@ -4750,3 +4820,86 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
 
     return srtp_err_status_ok;
 }
+
+#ifndef SRTP_NO_STREAM_LIST
+
+/* in the default implementation, we have an intrusive linked list */
```
bifurcation, line 4826:
```suggestion
/* in the default implementation, we have an intrusive doubly-linked list */
```

srtp/srtp.c
```diff
@@ -4750,3 +4820,86 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
...
+struct srtp_stream_list_ctx_t_ {
+    srtp_stream_ctx_t data;
```
bifurcation, line 4828:
```suggestion
    /* a stub stream that just holds pointers to the beginning and end of the list */
    srtp_stream_ctx_t data;
```

srtp/srtp.c
```diff
@@ -4750,3 +4820,86 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
...
+} srtp_stream_list_ctx_t_;
+
+srtp_err_status_t srtp_stream_list_alloc(srtp_stream_list_t *list_ptr)
+{
+    srtp_stream_list_t list =
+        srtp_crypto_alloc(sizeof(srtp_stream_list_ctx_t_));
+    if (list == NULL) {
+        return srtp_err_status_alloc_fail;
+    }
+    *list_ptr = list;
```
bifurcation, line 4838:
```suggestion
    list->prev = NULL;
    list->next = NULL;

    *list_ptr = list;
```

bifurcation: Actually, what is the right "empty list" state?  On the one hand, `srtp_stream_list_dealloc` assumes `NULL`.  But if we want this to actually be a proper doubly-linked list, traversable from either end, it seems like the empty list state should have `data.prev == data.next == &data`.  This also avoids the need for `stream->next == NULL` checks.

bifurcation: All that said, why do we need the double linkage here?  It seems like libsrtp only ever needs to traverse in one direction.

test/srtp_driver.c
```diff
@@ -4287,3 +4299,240 @@ const srtp_policy_t wildcard_policy = {
     0,    /* list of encrypted extension headers is empty */
     NULL
 };
+
+static srtp_stream_t stream_list_test_create_stream(uint32_t ssrc)
+{
+    srtp_stream_t stream = calloc(1, sizeof(srtp_stream_ctx_t));
```
bifurcation, line 4305:
```suggestion
    srtp_stream_t stream = malloc(sizeof(srtp_stream_ctx_t));
```

test/srtp_driver.c
```diff
@@ -4287,3 +4299,240 @@ const srtp_policy_t wildcard_policy = {
...
+        return srtp_err_status_fail;
+    }
+    if (srtp_stream_list_insert(list, stream_list_test_create_stream(3))) {
+        return srtp_err_status_fail;
+    }
+    if (srtp_stream_list_insert(list, stream_list_test_create_stream(4))) {
+        return srtp_err_status_fail;
+    }
+
+    /* find */
```
bifurcation, line 4369: Would be good to test for a false positive as well, say `srtp_stream_list_get(list, 5) != NULL`

include/srtp_priv.h
```diff
@@ -96,12 +97,6 @@ srtp_err_status_t srtp_steam_init_all_master_keys(
     srtp_master_key_t **keys,
     const unsigned int max_master_keys);
 
-/*
- * srtp_stream_init(s, p) initializes the srtp_stream_t s to
- * use the policy at the location p
- */
-srtp_err_status_t srtp_stream_init(srtp_stream_t srtp, const srtp_policy_t *p);
-
```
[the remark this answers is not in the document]
pabuhler: Original PR added srtp_stream_dealloc() here to make it semi public. Since it is no longer needed I removed and noticed that srtp_stream_init is also semi public but never referenced so thought to remove it from here as well. If you prefer I revert and make a new PR for this then let me know.

srtp/srtp.c
pabuhler: we should really do a sweep of all the code and fix this every where. But agree that it is good add them as we go.

test/srtp_driver.c
pabuhler: see the comment a bit further down in the file:  /* find not in list */
that is I guess what you meant ?

srtp/srtp.c
pabuhler: srtp_crypto_alloc  will have set the struct to zero, so setting NULL here is not necessary but is maybe ok to be explicit.

The double linkage is only to facilitate fast removal, ie to get hold of the previous item. There is no intention to traverse in revers.

test/srtp_driver.c
bifurcation: Yep, thanks.

bifurcation (approved):
Thanks for all the work on this @mildsunrise and @pabuhler.  Looks good to me with the latest changes.

mildsunrise:
Thank you everyone :) @murillo128 I'll integrate this into the media server when I find a spot

PR #614, opened 2022-11-07
Call `set_aad` and `get_tag` in AEAD performance tests

bifurcation:
Right now, the cipher speed test in `srtp_cipher_bits_per_second` only calls `set_iv` and `encrypt`.  In reality, when an AEAD cipher is used, there are also calls to `set_aad` and `get_tag`.  This PR adds those calls to `srtp_cipher_bits_per_second` so that it more accurately reflects AEAD performance.

### DIFF
--- a/crypto/cipher/cipher.c
+++ b/crypto/cipher/cipher.c
@@ -632,8 +632,9 @@ uint64_t srtp_cipher_bits_per_second(srtp_cipher_t *c,
     clock_t timer;
     unsigned char *enc_buf;
     unsigned int len = octets_in_buffer;
+    uint32_t tag_len = 32;
 
-    enc_buf = (unsigned char *)srtp_crypto_alloc(octets_in_buffer);
+    enc_buf = (unsigned char *)srtp_crypto_alloc(octets_in_buffer + tag_len);
     if (enc_buf == NULL) {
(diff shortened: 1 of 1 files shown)


crypto/cipher/cipher.c
```diff
@@ -632,24 +632,47 @@ uint64_t srtp_cipher_bits_per_second(srtp_cipher_t *c,
...
+
+        // Encrypt the buffer
         if (srtp_cipher_encrypt(c, enc_buf, &len) != srtp_err_status_ok) {
             srtp_crypto_free(enc_buf);
             return 0;
         }
+
+        // Get tag if supported by the cipher
+        if (c->type->get_tag) {
+           if (srtp_cipher_get_tag(c, (uint8_t*)(enc_buf + octets_in_buffer), &tag_len) !=
```
pabuhler, line 670: shouldn't this be (enc_buf + len) .... not really important i guess as the content of the buffer is not interesting.

pabuhler (approved):
LGTM  once the format check is fixed

crypto/cipher/cipher.c
bifurcation: Good point.  I also changed `tag_len = 32` to use `SRTP_MAX_TAG_LEN` instead, to be more accurate.

PR #616, opened 2022-11-21
add CIFuzz Github action to workflows

DavidKorczynski:
Add [CIFuzz](https://google.github.io/oss-fuzz/getting-started/continuous-integration/) workflow action to have fuzzers build and run on each PR. This is a service offered by OSS-Fuzz where libsrtp already runs. In the current PR the fuzzers gets build on a pull request and will run for 300 seconds.

I noticed the CI already has a fuzz action so this is an alternative suggestion in case you're interested -- CIFuzz has a set of features that are useful e.g. only highlighting issues only if they are introduced by the given PR, using of OSS-Fuzz code corpus and more. Let me know what you think!

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,11 @@
 name: CIFuzz
-on: [ push, pull_request ]
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pabuhler:
Hi @DavidKorczynski , I would be ok with this, I think the existing fuzzer task is just to verify the fuzzer builds.
If no one objects and it works we can merge it soon.

pabuhler:
The other work flows also run when pushing to main, should that be added here as well, as fuzzy the result of the merge is also important ?

DavidKorczynski:
> The other work flows also run when pushing to main, should that be added here as well, as fuzzy the result of the merge is also important ?

Yep, this sounds good to me! Done

PR #618, opened 2022-11-21
Respect the direction set in a template stream

bifurcation:
This PR attempts to fix the logical error noted in #617, that the directions set on a template stream are not checked when the template stream is used to create new (directed) streams.  It simply adds checks in `srtp_protect_mki` and `srtp_unprotect_mki` that verify that the template stream's direction is compatible with the context, either the same direction or `dir_srtp_unknown`.  Marking as WIP because even if we agree on approach, we probably need to cover the analgous RTCP methods.

Note that there is compatibility risk in merging this PR: If there are consumers that are relying the ambiguity noted here, then they will fail.

### DIFF
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -219,6 +219,8 @@ typedef enum {
                                         /**< needed                          */
     srtp_err_status_buffer_small = 28,  /**< out buffer is too small         */
     srtp_err_status_cryptex_err = 29,   /**< unsupported cryptex operation   */
+    srtp_err_status_direction_mismatch =
+        30, /**< operation does not match stream direction */
 } srtp_err_status_t;
 
 typedef struct srtp_ctx_t_ srtp_ctx_t;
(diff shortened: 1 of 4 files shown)


pabuhler:
This code apears to be 17 years old and kind of unchanged, I agree that it should be fixed, but it might break things for exisitng clients if they have been doing it wrong. Maybe this is something for 3.0 ? 
The update_xxx() functions also need to preserver the directions flags.

pabuhler:
@bifurcation , I took your change and applied some more direction checking, can you do a review ?

Summary of changes:

Tightens SRTP stream direction handling so sender and receiver streams cannot be used for the wrong operation.

It adds a new `srtp_err_status_direction_mismatch` status for cases where an existing stream is used with the opposite direction, and applies that check consistently across RTP and RTCP protect/unprotect paths. Wildcard template streams now also reject incompatible inbound/outbound operations before creating a stream.

The change also preserves stream direction across `srtp_update()`, including both specific streams and wildcard template streams, and rejects template updates that would flip an established template direction.

PR #623, opened 2022-12-07
Use ubuntu 20 04 in workflow

pabuhler:
This is just until things pass with ubuntu-latest agina


pabuhler:
close as accidentally included extra commit

PR #627, opened 2022-12-12
Fix more warnings to compile with clang 15

pabuhler:
Clang 15 enables -Wstrict-prototypes by default in C .
see https://discourse.llvm.org/t/rfc-enabling-wstrict-prototypes-by-default-in-c/60521

### DIFF
--- a/test/replay_driver.c
+++ b/test/replay_driver.c
@@ -251,7 +251,6 @@ double rdb_check_adds_per_second(void)
     uint32_t i;
     srtp_rdb_t rdb;
     clock_t timer;
-    int failures = 0; /* count number of failures        */
 
     if (srtp_rdb_init(&rdb) != srtp_err_status_ok) {
         printf("rdb_init failed\n");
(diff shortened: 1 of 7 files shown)


test/replay_driver.c
```diff
@@ -260,18 +259,12 @@ double rdb_check_adds_per_second(void)
...
-        if (srtp_rdb_check(&rdb, i) != srtp_err_status_ok)
-            ++failures;
-        if (srtp_rdb_add_index(&rdb, i) != srtp_err_status_ok)
-            ++failures;
+        srtp_rdb_check(&rdb, i + 2);
+        srtp_rdb_add_index(&rdb, i + 2);
+        srtp_rdb_check(&rdb, i + 1);
+        srtp_rdb_add_index(&rdb, i + 1);
+        srtp_rdb_check(&rdb, i);
+        srtp_rdb_add_index(&rdb, i);
```
bifurcation, line 267: This seems unrelated to the main topic of this PR, but still a good thing to fix.

bifurcation (approved):

test/replay_driver.c
pabuhler: It only showed up with clang 15, so some what related

Issue #597, opened 2022-02-11
update openssl version check in configure / cmake


pabuhler:
After #593  was merged the minimal openssl version is now 1.1.0, this should be reflected in documentation and configure / cmake scripts.

PR #629, opened 2022-12-15
update min openssl version to 1.1.0

pabuhler:
As of commit #593 the minimal supported version is 1.1.0, this updates build files to check for that version. This also means the check for a broken openssl cleans function is obsolete so it is removed.

#597


JonathanLennox:
It looks like there are still `OPENSSL_VERSION_NUMBER < 0x10100000L` checks in aes_gcm_ossl.c and aes_icm_ossl.c ... presumably with this change those can be removed too?

JonathanLennox:
For the autotools, it'd be good to have autoconf check for some openssl 1.1.0-only symbol that libsrtp uses...presumably `HMAC_CTX_new`.

pabuhler:
@JonathanLennox  HMAC_CTX_new is leaving us with OpenSSL 3 so I went with EVP_CIPHER_CTX_reset .

PR #632, opened 2023-01-08
stream list alternative

pabuhler:
An alternative / modification to #612 .

This PR is just for preliminary review, will merge in to #612 if ok.


pabuhler:
relevant changes have been added to #612

Issue #645, opened 2023-03-28
Performance issue with OpennSSL 3 and libsrtp


murillo128:
I think it is an OpenSSL issue, but crossposting it here:
 

https://github.com/openssl/openssl/issues/20625

murillo128:
I am building libsrtp using nodejs bundled OpenSSL and I have seen a performance performance issue with node 18 & 19 which bundles OpenSSL 3.0 both on AES GCM and AES CTR modes compared to node 16 which bundles OpenSSL 1.1 using the following benchmark:

https://gist.github.com/Sean-Der/7a42bd70edfe1324ccc6ab399d653c0e

The performance hit is quite significative, with AES GCM being more than 2x slower and AES CTR being 1.5x slower (depending on the server types I have tested with):

Server 1:
```
node v16.19.0 - openssl 1.1.1s+quic
gcm test done in 847746ms
ctr test done in 2209247ms

node v18.15.0 - openssl 3.0.8+quic
gcm test done in 2042822
ctr test done in 2837856
Server 2:

node v16.19.0  - openssl 1.1.1s+quic
gcm test done in 2210976
ctr test done in 3035863

node v18.15.0  - openssl 3.0.8+quic
gcm test done in 4259098
ctr test done in 4081153
```

murillo128:
i can try later with bigger payloads

murillo128:
Tried with different payload sizes and still seeing the performance regression on OpenSSL 3

```
Node 16
0    payload size gcm done in 886714
0    payload size ctr done in 2421151
400  payload size gcm done in 1494584
400  payload size ctr done in 4381984
1200 payload size gcm done in 2326866
1200 payload size ctr done in 8568198

Node 18
0    payload size gcm done in 2007877
0    payload size ctr done in 3287084
400  payload size gcm done in 2705423
400  payload size ctr done in 5485942
1200 payload size gcm done in 3549461
1200 payload size ctr done in 9669171
```

murillo128:
I agree that maintaining the code is a burden, that's why i preferred to be in libsrtp itself hehe..

the srtp_replace_cipher_type seems very helpful and would allow me to implement the hw optimized gcm version for the chipsets i care to optimize, but keeping openssl backend for everything else. Thank you for the tip!

mildsunrise:
[...]

PR #646, opened 2023-03-30
openssl 3 performance

pabuhler:
Investigate and fix reported performance issues with openssl 3 reported in #645.

Try explicit fetching once per stream, should be more efficient than implicit featuring on demand.

This is openssl 3 only so with fail to build with older version for now.


pabuhler:
This change did not make the expected difference in performance as the cipher is already on fetched once when the stream is created.

PR #649, opened 2023-04-03
iv length is constant so set only once

pabuhler:
The iv length is preserved inside the EVP_CIPHER_CTX so no need to set more than once.
This is especially important with OpenSSL 3 where setting the iv len is a expensive operation due to param lookup code inside of OpenSSL.

This should help performance issue with OpenSSL 3 and libsrtp with #645 in the case of GCM but does not fix all of the performance issues.


murillo128:
I have been testing this PR and it marginally helps performance, so good to have it in anyway.

pabuhler:
@murillo128  I agree that it should just go in. @bifurcation  could you review ?

PR #658, opened 2023-11-08
remove comma from srtp_profile_t enum value.

melchi45:
remove comma from srtp_profile_t enum value.

comma have error old compiler on end of enum.


paulej (approved):
Syntax is valid either way, but I have no objection to removing it.

PR #660, opened 2023-12-05
meson.build: implement mbedtls support

iameli:
This was pretty easy; just porting over mbedtls support that was already added from https://github.com/cisco/libsrtp/pull/512. Now it works with Meson too!

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -1,0 +1,110 @@
+name: Meson CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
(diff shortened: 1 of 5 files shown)


pabuhler:
@iameli   I have no problem with merging these changes, although I do not know much about meson build. Since we moved to github actions for CI meson builds are no longer run regularly. Would you be interested in adding a git hub action for this ?

pabuhler (approved):

pabuhler:
@iameli awesome,  I have merged this now. In general I dont think we will block merging changes when a meson build fails, but at least we will now know something needs to be done.

PR #665, opened 2023-12-16
start using const on internal arguments

pabuhler:
Add const cast warnings.
Add const when the arguments do not need to be modified.

### DIFF
--- a/crypto/cipher/aes_gcm_nss.c
+++ b/crypto/cipher/aes_gcm_nss.c
@@ -212,7 +212,9 @@ static srtp_err_status_t srtp_aes_gcm_nss_context_init(void *cv,
         return (srtp_err_status_cipher_fail);
     }
 
-    SECItem key_item = { siBuffer, (unsigned char *)key, c->key_size };
+    /* explicitly cast away const of key */
+    SECItem key_item = { siBuffer, (unsigned char *)(uintptr_t)key,
+                         c->key_size };
     c->key = PK11_ImportSymKey(slot, CKM_AES_GCM, PK11_OriginUnwrap,
                                CKA_ENCRYPT, &key_item, NULL);
     PK11_FreeSlot(slot);
--- a/crypto/cipher/aes_icm_nss.c
+++ b/crypto/cipher/aes_icm_nss.c
@@ -254,7 +254,9 @@ static srtp_err_status_t srtp_aes_icm_nss_context_init(void *cv,
         return srtp_err_status_bad_param;
     }
 
-    SECItem keyItem = { siBuffer, (unsigned char *)key, c->key_size };
+    /* explicitly cast away const of key */
+    SECItem keyItem = { siBuffer, (unsigned char *)(uintptr_t)key,
+                        c->key_size };
     c->key = PK11_ImportSymKey(slot, CKM_AES_CTR, PK11_OriginUnwrap,
                                CKA_ENCRYPT, &keyItem, NULL);
     PK11_FreeSlot(slot);
--- a/crypto/hash/hmac_nss.c
+++ b/crypto/hash/hmac_nss.c
@@ -187,7 +187,8 @@ static srtp_err_status_t srtp_hmac_init(void *statev,
         return srtp_err_status_bad_param;
     }
 
-    SECItem key_item = { siBuffer, (unsigned char *)key, key_len };
+    /* explicitly cast away const of key */
+    SECItem key_item = { siBuffer, (unsigned char *)(uintptr_t)key, key_len };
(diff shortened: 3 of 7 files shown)


pabuhler:
This lays the ground work for using const in more places

crypto/cipher/aes_gcm_nss.c
```diff
@@ -212,7 +212,9 @@ static srtp_err_status_t srtp_aes_gcm_nss_context_init(void *cv,
         return (srtp_err_status_cipher_fail);
     }
 
-    SECItem key_item = { siBuffer, (unsigned char *)key, c->key_size };
+    /* explicitly cast away const of key */
+    SECItem key_item = { siBuffer, (unsigned char *)(uintptr_t)key,
```
paulej, line 216: I think the compiler is whing too much if this is an issue.

crypto/cipher/aes_icm_nss.c
```diff
@@ -254,7 +254,9 @@ static srtp_err_status_t srtp_aes_icm_nss_context_init(void *cv,
         return srtp_err_status_bad_param;
     }
 
-    SECItem keyItem = { siBuffer, (unsigned char *)key, c->key_size };
+    /* explicitly cast away const of key */
+    SECItem keyItem = { siBuffer, (unsigned char *)(uintptr_t)key,
```
paulej, line 258: Ditto here

crypto/hash/hmac_nss.c
```diff
@@ -187,7 +187,8 @@ static srtp_err_status_t srtp_hmac_init(void *statev,
         return srtp_err_status_bad_param;
     }
 
-    SECItem key_item = { siBuffer, (unsigned char *)key, key_len };
+    /* explicitly cast away const of key */
+    SECItem key_item = { siBuffer, (unsigned char *)(uintptr_t)key, key_len };
```
paulej, line 191: And here

srtp/srtp.c
```diff
@@ -79,32 +79,32 @@ srtp_debug_module_t mod_srtp = {
...
+    if (pkt_octet_len < rtp_header_len)
         return srtp_err_status_bad_param;
 
     /* Verifing profile length. */
     if (hdr->x == 1) {
-        srtp_hdr_xtnd_t *xtn_hdr =
-            (srtp_hdr_xtnd_t *)((uint32_t *)hdr + uint32s_in_rtp_header +
-                                hdr->cc);
+        const srtp_hdr_xtnd_t *xtn_hdr =
+            (const srtp_hdr_xtnd_t *)((const uint32_t *)hdr +
```
paulej, line 102: Pre-existing, but the double cast here is also a little surprising that it might be necessary. `const` additions good.

srtp/srtp.c
```diff
@@ -79,32 +79,32 @@ srtp_debug_module_t mod_srtp = {
...
+                                      uint32s_in_rtp_header + hdr->cc);
         int profile_len = ntohs(xtn_hdr->length);
         rtp_header_len += profile_len * 4;
         /* profile length counts the number of 32-bit words */
-        if (*pkt_octet_len < rtp_header_len)
+        if (pkt_octet_len < rtp_header_len)
```
paulej, line 107: Wow. I'm surprised the compiler did not complain about this one.

srtp/srtp.c
```diff
@@ -2568,9 +2568,9 @@ srtp_err_status_t srtp_unprotect_mki(srtp_ctx_t *ctx,
 
     /* Determine if MKI is being used and what session keys should be used */
     if (use_mki) {
-        session_keys = srtp_get_session_keys(
-            stream, (uint8_t *)hdr, (const unsigned int *)pkt_octet_len,
-            &mki_size);
+        session_keys =
+            srtp_get_session_keys(stream, (const uint8_t *)hdr,
+                                  (unsigned int)*pkt_octet_len, &mki_size);
```
paulej, line 2573: While this cast might be necessary, why do we have a mix of signed and unsigned integer use? That might be a good thing to harmonize. I personally prefer unsigned integers for length values.

paulej (approved):

srtp/srtp.c
pabuhler: This is only because of the usage of uint32s_in_rtp_header, so needs needs the first cast to do the right pointer arithmetic. I have have another change coming that provides some helper functions for the hdr ext stuff as these cast are ugly and murmurous.

srtp/srtp.c
pabuhler: this was due to api change from int * to int

srtp/srtp.c
pabuhler: This is something we have been looking at and yes we need to harmonize on this and other types, it is coming :)
For length I would even go as far as using size_t every where including in the public api in v3!

PR #670, opened 2023-12-27
fix srtp_remove_stream to use SSRC in host byte order

pabuhler:
This fixes the inconsistency in SSRC byte order usage in the public api. Now all usage should be in host byte order.

#220

### DIFF
--- a/fuzzer/fuzzer.c
+++ b/fuzzer/fuzzer.c
@@ -849,7 +849,7 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
             /* Keep removing streams until the set of SSRCs extracted from the
              * fuzzer input is exhausted */
             if (i < num_remove_stream) {
-                if (srtp_remove_stream(srtp_ctx, remove_stream_ssrc[i]) !=
+                if (srtp_stream_remove(srtp_ctx, remove_stream_ssrc[i]) !=
                     srtp_err_status_ok) {
                     goto end;
                 }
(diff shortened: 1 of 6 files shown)


JonathanLennox:
See @jesup 's comment on #220 at https://github.com/cisco/libsrtp/issues/220#issuecomment-274152974: I don't think changing this while leaving the function name and signature the same is a good idea.  Even at an API-breaking change like 3.0.0, this won't have any compiler warning to tell someone porting to the new API that they need to change something.

pabuhler:
@JonathanLennox fair point,  so what about the new name, srtp_delete_stream() ?  
Other options could be to change from add/remove to create/destroy ...
Just changing the name though will not ensure people take notice of the change in parameters.

paulej (changes requested):
I agree with changes like this for consistency. The app developer should never have to call `htons()` or `htonl()` to use the libsrtp APIs, IMO.  I also agree with Jonathan's comment from Randell that we shouldn't make a change that might blindside developers. Perhaps just rename the API slightly? (We could keep both the old and a new one, but I think that's less desirable.  Upgrading SRTP and having a clean interface should not be hard, especially if these kinds of breaking changes are documented in one obvious place.)

paulej:
Maybe `srtp_stream_add` and `srtp_stream_remove`?  Just changing the order, but I prefer to have the noun before the action in APIs.  APIs are mixed in this regard, too.

pabuhler:
> Maybe `srtp_stream_add` and `srtp_stream_remove`? Just changing the order, but I prefer to have the noun before the action in APIs. APIs are mixed in this regard, too.

I think I like that, how about, I rename to srtp_stream_remove / add / update and add a note to the documentation of srtp_stream_remove about the change in byte order. It will also end up in the CHANGES file.

Can make a new task to ensure all API's are reviewed for naming consistency.

paulej:
> I think I like that, how about, I rename to srtp_stream_remove / add / update and add a note to the documentation of srtp_stream_remove about the change in byte order. It will also end up in the CHANGES file.
> 
> Can make a new task to ensure all API's are reviewed for naming consistency.

Sounds good to me.

paulej (approved):

PR #674, opened 2024-01-04
stream_index

jmillan:
Enable stream index for fast stream retrieval

* Maintains the current stream list.
* Creates a new steam index for fast stream retrieval.

### Rationale

A linked list:
- Fast adding and removing elements on the edge.
- Slow traversal
  -  Each node resides in a different place in memory
  -  Each node points to the next one, requiring pointer dereferencing on each node.

libsrtp does orders of magnitude more list traversal than additions or removals, hence a linked list is not a proper way to store them.

The solution in this PR is keeping a stream index in an array such as:
- Array elements are contiguous in memory.
- No pointer dereferencing needed for node traversal.
- Great cache locality.

### Details
- Inserting a new stream index item may require reallocating memory in order to accomodate the new item and keep all them in contiguous memory.
- Deleting a stream index requires moving one position in memory the following index entries.
- Traversing the stream index is as fast as it can be due to the previously given rationale.

### Performance

In my tests I find this solution being >6 times more performant than current master branch.

#### master
Screenshot 2024-01-04 at 15 32 29

#### stream-index
Screenshot 2024-01-04 at 15 32 45

### Others

IMHO the linked list may perfectly be removed in favour of the steam index, but as in this PR, they can perfectly coexist.

I'm letting libsrtp maintainers decide to adapt and apply this code to their will.

NOTE: This is currently enabled via a flag. I believe `config` files will require some work for the option to be properly added.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -4841,7 +4841,7 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
 
 #ifndef SRTP_NO_STREAM_LIST
 
-#ifdef SRTP_STREAM_INDEX
+#ifdef ENABLE_STREAM_INDEX
 
 #define INITIAL_STREAM_INDEX_SIZE 2
 
@@ -4970,7 +4970,7 @@ srtp_stream_t srtp_stream_index_get(srtp_stream_index stream_index,
 
 /* in the default implementation, we have an intrusive doubly-linked list */
 typedef struct srtp_stream_list_ctx_t_ {
-#ifdef SRTP_STREAM_INDEX
+#ifdef ENABLE_STREAM_INDEX
     srtp_stream_index index;
 #endif
     /* a stub stream that just holds pointers to the beginning and end of the
@@ -4986,7 +4986,7 @@ srtp_err_status_t srtp_stream_list_alloc(srtp_stream_list_t *list_ptr)
         return srtp_err_status_alloc_fail;
     }
 
-#ifdef SRTP_STREAM_INDEX
+#ifdef ENABLE_STREAM_INDEX
     srtp_err_status_t stat = srtp_stream_index_alloc(&list->index);
     if (stat) {
         return stat;
@@ -5009,7 +5009,7 @@ srtp_err_status_t srtp_stream_list_dealloc(srtp_stream_list_t list)
     if (list->data.next) {
         return srtp_err_status_fail;
     }
-#ifdef SRTP_STREAM_INDEX
+#ifdef ENABLE_STREAM_INDEX
     srtp_stream_index_dealloc(list->index);
 #endif
     srtp_crypto_free(list);
@@ -5027,7 +5027,7 @@ srtp_err_status_t srtp_stream_list_insert(srtp_stream_list_t list,
     list->data.next = stream;
     stream->prev = &(list->data);
 
-#ifdef SRTP_STREAM_INDEX
+#ifdef ENABLE_STREAM_INDEX
     srtp_err_status_t stat = srtp_stream_index_insert(list->index, stream);
     if (stat) {
         return stat;
@@ -5039,7 +5039,7 @@ srtp_err_status_t srtp_stream_list_insert(srtp_stream_list_t list,
 
 srtp_stream_t srtp_stream_list_get(srtp_stream_list_t list, uint32_t ssrc)
 {
-#ifdef SRTP_STREAM_INDEX
+#ifdef ENABLE_STREAM_INDEX
     return srtp_stream_index_get(list->index, ssrc);
 #else
     /* walk down list until ssrc is found */
@@ -5065,7 +5065,7 @@ void srtp_stream_list_remove(srtp_stream_list_t list,
     if (stream_to_remove->next != NULL) {
         stream_to_remove->next->prev = stream_to_remove->prev;
     }
-#ifdef SRTP_STREAM_INDEX
+#ifdef ENABLE_STREAM_INDEX
     srtp_stream_index_remove(list->index, stream_to_remove->ssrc);
 #endif
 }
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -4867,6 +4867,7 @@ srtp_err_status_t srtp_stream_index_alloc(srtp_stream_index *stream_index_ptr)
     stream_index->entries = srtp_crypto_alloc(sizeof(stream_index_entry) *
                                               INITIAL_STREAM_INDEX_SIZE);
     if (stream_index->entries == NULL) {
+        srtp_crypto_free(stream_index);
         return srtp_err_status_alloc_fail;
     }
 
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -4895,14 +4895,14 @@ srtp_err_status_t srtp_stream_index_insert(srtp_stream_index stream_index,
             return srtp_err_status_alloc_fail;
         }
 
-        // copy previous entries into the new ones.
+        // copy previous entries into the new ones
         memcpy(new_entries, stream_index->entries,
                sizeof(stream_index_entry) * stream_index->size);
-        // release previous entries.
+        // release previous entries
         srtp_crypto_free(stream_index->entries);
-        // assign new entries to the index.
+        // assign new entries to the index
         stream_index->entries = new_entries;
-        // update index info.
+        // update index info
         stream_index->size = new_size;
         stream_index->available = new_size / 2;
     }
@@ -4912,7 +4912,7 @@ srtp_err_status_t srtp_stream_index_insert(srtp_stream_index stream_index,
     stream_index->entries[next_index].ssrc = stream->ssrc;
     stream_index->entries[next_index].stream = stream;
 
-    // update available value.
+    // update available value
     stream_index->available--;
 
     return srtp_err_status_ok;
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -4841,8 +4841,138 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
 
 #ifndef SRTP_NO_STREAM_LIST
 
+#ifdef SRTP_STREAM_INDEX
+
+#define INITIAL_STREAM_INDEX_SIZE 2
+
+typedef struct stream_index_entry {
+    uint32_t ssrc;
+    srtp_stream_t stream;
+} stream_index_entry;
+
+typedef struct srtp_stream_index_ {
+    stream_index_entry *entries;
+    size_t size;
+    size_t available;
+} srtp_stream_index_;
+
+void srtp_stream_index_dump(srtp_stream_index stream_index)
+{
+    printf("stream index size: %zu\n", stream_index->size);
+    printf("stream index available: %zu\n", stream_index->available);
+
+    unsigned int end = stream_index->size - stream_index->available;
+    unsigned int i = 0;
+
+    for (; i < end; i++) {
+        printf("ssrc: %d, stream:%p\n", stream_index->entries[i].ssrc,
+               stream_index->entries[i].stream);
+    }
+}
+srtp_err_status_t srtp_stream_index_alloc(srtp_stream_index *stream_index_ptr)
+{
+    srtp_stream_index stream_index =
+        srtp_crypto_alloc(sizeof(srtp_stream_index_));
+    if (stream_index == NULL) {
+        return srtp_err_status_alloc_fail;
+    }
+
+    stream_index->entries = srtp_crypto_alloc(sizeof(stream_index_entry) *
+                                              INITIAL_STREAM_INDEX_SIZE);
+    if (stream_index->entries == NULL) {
+        return srtp_err_status_alloc_fail;
+    }
+
+    *stream_index_ptr = stream_index;
+
+    return srtp_err_status_ok;
+}
+
+srtp_err_status_t srtp_stream_index_dealloc(srtp_stream_index stream_index)
+{
+    srtp_crypto_free(stream_index->entries);
+    srtp_crypto_free(stream_index);
+
+    return srtp_err_status_ok;
+}
+
+srtp_err_status_t srtp_stream_index_insert(srtp_stream_index stream_index,
+                                           srtp_stream_t stream)
+{
+    // there is no available index entry, duplicate number of entries
+    if (stream_index->available == 0) {
+        size_t new_size = stream_index->size * 2;
+        stream_index_entry *new_entries =
+            srtp_crypto_alloc(sizeof(stream_index_entry) * new_size);
+        if (new_entries == NULL) {
+            return srtp_err_status_alloc_fail;
+        }
+
+        // copy previous entries into the new ones.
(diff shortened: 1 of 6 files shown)


srtp/srtp.c
```diff
@@ -4841,8 +4841,138 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
...
+    srtp_stream_t stream;
+} stream_index_entry;
+
+typedef struct srtp_stream_index_ {
+    stream_index_entry *entries;
+    size_t size;
+    size_t available;
+} srtp_stream_index_;
+
+void srtp_stream_index_dump(srtp_stream_index stream_index)
```
jmillan, line 4859: Not used, was written for debugging purposes.

srtp/srtp.c
```diff
@@ -4841,8 +4841,125 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
...
+    srtp_stream_index stream_index =
+        srtp_crypto_alloc(sizeof(srtp_stream_index_));
+    if (stream_index == NULL) {
+        return srtp_err_status_alloc_fail;
+    }
+
+    stream_index->entries = srtp_crypto_alloc(sizeof(stream_index_entry) *
+                                              INITIAL_STREAM_INDEX_SIZE);
+    if (stream_index->entries == NULL) {
+        return srtp_err_status_alloc_fail;
```
saghul, line 4870: If you exit here you'd leak `stream_index`, wouldn't you?

jmillan: correct :+1:

jmillan: I'll fix it.

paulej (changes requested):
Generally, I'm OK with this given the performance improvements.

I do think we should discuss whether we should remove the linked lists. If this does, indeed, perform well, I would prefer it and keeping the existing code seems like it would just create maintenance issues.

This will still iterate over elements serially. Previously, you had proposed a hashing method that I think would have subdivided the search space. Do you think that idea might still be useful along with this? Several people have expressed concern over the use of libsrtp in large-scale conferences due to these searches. If there are 2000 streams, I assume the linear search is still a little slow. What if we use this array approach here but preface it with a 256-member bucket of pointers. We do something like `stream_lists[SSRC & 0xff]` to find the right bucket and then use the stream list logic you have here once in the right bucket.  We could delay allocating memory for those 256 lists until there is a need to use a given bucket. Thus, for small conferences the memory use would still be relatively small.

srtp/srtp.c
jmillan: done, good catch @saghul ❤️

srtp/srtp.c
```diff
@@ -4841,8 +4841,125 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
...
+
+    // update available value
+    stream_index->available--;
+
+    return srtp_err_status_ok;
+}
+
+void srtp_stream_index_remove(srtp_stream_index stream_index, uint32_t ssrc)
+{
+    unsigned int i = 0;
```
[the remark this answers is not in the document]
jmillan: I'm following the conventions I see all over the code. Should I do it in a different way?

srtp/srtp.c
```diff
@@ -4841,8 +4841,125 @@ srtp_err_status_t srtp_get_stream_roc(srtp_t session,
...
+
+            break;
+        }
+    }
+}
+
+srtp_stream_t srtp_stream_index_get(srtp_stream_index stream_index,
+                                    uint32_t ssrc)
+{
+    unsigned int i = 0;
```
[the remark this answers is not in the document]
jmillan: Same, I can perfectly modify to whatever you tell me. I'm just following the conventions I'm seeing in the code.

```txt
srtp/srtp.c
157:    unsigned int i = 0;
338:    unsigned int i = 0;
505:    unsigned int i = 0;
904:    unsigned int i = 0;
1626:    unsigned int i = 0;
4924:    unsigned int i = 0;
4943:    unsigned int i = 0;
```

jmillan:
> This will still iterate over elements serially.

But has nothing to do in terms of performance for the reasons given in the description.

> Previously, you had proposed a hashing method that I think would have subdivided the search space. Do you think that idea might still be useful along with this?

Yes, the hashing would go further in terms of speed for large environments. I did not even think of implementing it here as it was rejected in the first place in my previous PR.

Anyway the hashing could be added in a different PR to make things more clear.

 > What if we use this array approach here but preface it with a 256-member bucket of pointers. 

This was a memory concern in my hashing PR, even though I created a hash with 32 buckets https://github.com/cisco/libsrtp/pull/659#issuecomment-1838649171

My opinion is that everything commented here can be done sequentially in the corresponding PRs:

- Add `stream_index` keeping `stream_list`. (this PR)
- Remove `stream_list`? NOTE: `stream_list` related functions would need to bee kept since libsrtp offers an inteface to those so the user can override these already.
- Add a hash.

srtp/srtp.c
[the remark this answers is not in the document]
pabuhler: I think it is fine to move the iterator into the for loop, there is abit of a mix in the existing code but that is mostly historical and going forward it should be style for new code.

srtp/srtp.c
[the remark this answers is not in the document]
pabuhler: ditto

srtp/srtp.c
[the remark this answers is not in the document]
jmillan: done

pabuhler:
@jmillan, hi, I am not sure if you noticed but in the main branch we have now started ABI breaking changes and will try to get a libSRTPv3 out some time this year, so if this is intended for 2.x then the PR should be based of and towards the 2_x_dev branch. I dont mind which you prefer, the difference is the main branch will be API unstable for a while so can readily make changes to existing AIP's etc, where as the 2_x_dev branch can not break compatibility.

I kind of agree with your actions points, but I think I would have preferred this to already be a separate srtp_list implementation rather than a mix, you could use the same config to select if this new list or the old list is complied in, what do you think?

I also think that adding the hash bucket would be a good idea, especially if we support dynamically growing the number of buckets and the size of the arrays.

jmillan:
Hi,

> @jmillan, hi, I am not sure if you noticed but in the main branch we have now started ABI breaking changes and will try to get a libSRTPv3 out some time this year, so if this is intended for 2.x then the PR should be based of and towards the 2_x_dev branch.

I don't have a big preference, honestly.

>  I think I would have preferred this to already be a separate srtp_list implementation rather than a mix, you could use the same config to select if this new list or the old list is complied in, what do you think?

Sounds good. I've just pushed this.

> I also think that adding the hash bucket would be a good idea, especially if we support dynamically growing the number of buckets and the size of the arrays.

A hash has a fixed number of buckets unless we want to move entries around when incrementing them, which we don't.

jmillan:
@pabuhler 

This is the current state:
- stream index substitutes the `stream_list` if `ENABLE_STREAM_INDEX` option is given, otherwise it uses the current implementation.
  -  meaning if current `stream_list` is to be removed in the future, we only need to remove a contiguous code block.
- add a hash?
  -  as commented, a hash needs to have a fixed number of buckets. Let me know if this PR can be already merged, so the hashing can be done next, in another PR.

paulej (approved):
Looks good to me. I would prefer to not have both the old list and new array, but we can consider removing the configuration as a later step.

jmillan:
@paulej, we are on time to remove the old list. We can perfectly remove it now and get rid of the configuration option.

Otherwise I would like to ask anyone to check whether the new configuration option is properly added and applied for all needed cases.

srtp/srtp.c
```diff
@@ -4926,4 +4927,161 @@ void srtp_stream_list_for_each(srtp_stream_list_t list,
...
+    return NULL;
+}
+
+void srtp_stream_list_for_each(srtp_stream_list_t list,
+                               int (*callback)(srtp_stream_t, void *),
+                               void *data)
+{
+    list_entry *entries = list->entries;
+
+    unsigned int ssrc;
```
pabuhler, line 5065: please use uint32_t  for ssrc

pabuhler:
> @paulej, we are on time to remove the old list. We can perfectly remove it now and get rid of the configuration option.
> 
> Otherwise I would like to ask anyone to check whether the new configuration option is properly added and applied for all needed cases.

If we a merging straight in to main then maybe just drop the configuration and remove the old list code.
If we remove the old code then should also remove the next / prev pointers in the strp_stream_ctx_t_ struct ?

If you keep the config then we will need to add it to cmake as well and should have a ci build that uses the config with at least one of the build systems. Also the config for autotools, ie "configure" should have been added to configure.ac,  the configure script is generated from that. I would have also renamed the define to start with SRTP_  .

Just removing the old code is by far the least work :)

jmillan:
> Just removing the old code is by far the least work :)

Done :-)

jmillan:
Fuzzer fails. I'll check it.

pabuhler:
> Fuzzer fails. I'll check it.

ok, do you get any where ? been a while since I have used the fuzzer locally, but I know we might need to update it if we are changing API's etc.

jmillan:
> ok, do you get any where ? been a while since I have used the fuzzer locally, but I know we might need to update it if we are changing API's etc.

I can reproduce it. This PR does not change the API so it may be something introduced here. I'll check into it as soon as I can.

jmillan:
The first problem I'm seeing is that two stream list entries, and hence two streams, with the same SSRCs are created:

Screenshot 2024-01-11 at 16 30 02

Is this even legal? Should libsrtp have 2 streams with the same ssrc? Can anyone please confirm this?

master branch does nothing to avoid that and neither does this branch.

I would say that should never happen and hence `srtp_insert_or_dealloc_stream()` or `srtp_stream_list_insert()` should check if there is already an stream with such ssrc rather than blindly insert it.

jmillan:
I've enhanced `srtp_stream_list_for_each()` so it does not consider `ssrc` in order to increase the loop iterator but rely on the list availability instead. This is needed because this method is used to iterate and remove the list entries.

The fuzzer should not fail now (It does not fail in my local env at least). But the previous question still needs to be answered, whether having multiple entries with the same SSRC should be permitted (IMHO NOT, but I need confirmation). This is something that master allows now BTW.

pabuhler:
> The fuzzer should not fail now (It does not fail in my local env at least). But the previous question still needs to be answered, whether having multiple entries with the same SSRC should be permitted (IMHO NOT, but I need confirmation). This is something that master allows now BTW.

I thought it was documented as undefined behavior, otherwise a full traversal of the list would be required for each insert. So no it is not allowed. In which case the fuzzer might need updating.
see: https://github.com/cisco/libsrtp/blob/fcc2517479dcb162a24da83740c156de5920f6c9/include/stream_list_priv.h#L88

srtp/srtp.c
```diff
@@ -4856,73 +4859,137 @@ srtp_err_status_t srtp_stream_list_alloc(srtp_stream_list_t *list_ptr)
...
 srtp_err_status_t srtp_stream_list_insert(srtp_stream_list_t list,
                                           srtp_stream_t stream)
 {
-    /* insert at the head of the list */
-    stream->next = list->data.next;
-    if (stream->next != NULL) {
-        stream->next->prev = stream;
+    /*
+     * there is no space to hold the new entry in the entries buffer,
+     * duplicate the size of the buffer.
```
pabuhler, line 4899: don't you mean "double the size ...."

jmillan:
> > The fuzzer should not fail now (It does not fail in my local env at least). But the previous question still needs to be answered, whether having multiple entries with the same SSRC should be permitted (IMHO NOT, but I need confirmation). This is something that master allows now BTW.
> 
> I thought it was documented as undefined behavior, otherwise a full traversal of the list would be required for each insert. So no it is not allowed. In which case the fuzzer might need updating.
> see: https://github.com/cisco/libsrtp/blob/fcc2517479dcb162a24da83740c156de5920f6c9/include/stream_list_priv.h#L88
>  

OK , yes, makes sense. Then we are good 👍

pabuhler (approved):

pabuhler:
If you are ready then I will merge, and thanks for your effort and patience :)

jmillan:
> If you are ready then I will merge, and thanks for your effort and patience :)

👍let me just fix the comment and make a couple more manual tests tomorrow.

jmillan:
@pabuhler,

Ready to merge 👍

saghul:
🙌

jmillan:
> @jmillan, hi, I am not sure if you noticed but in the main branch we have now started ABI breaking changes and will try to get a libSRTPv3 out some time this year, so if this is intended for 2.x then the PR should be based of and towards the 2_x_dev branch.

Considering this branch has been merged into `main`, this will be part of v3, right? 
Is `main` still considered stable?

pabuhler:
Yes this will be part of v3, it is possible to merged into the 2_x_dev branch if required. There is the option of replacing the list in 2.5 so I do not think it is critical.

main is not API stable for the foreseeable future but in terms of functionality and performance it should remain stable.

In general I think it is up the the PR author to decide to rebase / squash before merging but no hard and fast rules around that. In this case I agree that squashing it down would have probably been a good idea and I can take some blame for not suggesting it. In the end the diff was quite small and I forgot to look at commit history.

jmillan:
I was about to use the current 'main'  branch in [our project ](https://github.com/versatica/mediasoup) in order to get this enhancement. We are using meson, and for that we need to point the github .zip file for the exact commit or release of the subproject (libsrtp in this case). The problem is that github does not guarantee that the .zip file hash will remain unchanged for non release commits.

Having said this, we should point to a release hash rather than to a non release one. The question is then, when are you planning to release 2.5? If this will happen sooner than later then it could be interesting having this into 2.5 if possible.

pabuhler:
@jmillan this was why I asked where you intended to push it :) . I think we will try to release 2.6 during February as it is a year since last time and hopefully the last 2_x release. This change can be cherry picked on to the 2_x_dev branch with a few tweaks,  the question might come up if it should be behind config or not as there is more risk in 2_x .

jmillan:
Update: We have forked libsrtp, [created a release from the current `main`](https://github.com/versatica/libsrtp/releases/tag/v3.0-alpha) and started using it in mediasoup. This way we can start using this enhancement now and consequently verify its stability.

pabuhler:
@jmillan that sounds like a good idea, look forward to hearing any feedback and hopefully we get to a v3 release soon.

jmillan:
> This change can be cherry picked on to the 2_x_dev branch with a few tweaks, the question might come up if it should be behind config or not as there is more risk in 2_x .>

[mediasoup](https://mediasoup.org/) has been using this enhancement for more than a month and it's been working flawlessly.

jmillan:
I wonder if we should remain in `main` or move to 2_x, once this enhancement is moved. Do you have any consideration on that regard?

pabuhler:
@jmillan the API in main is in a state of flux at the moment so that alone is a reason to not pull from there until a v3 is closer to release.

Personally I prefer to spend time on trying to get v3 out, rather than porting this change to the 2_x_dev branch, but if someone creates a PR I will help review and get it merged.

Issue #671, opened 2023-12-27
Use consistent types through out code base


pabuhler:
Use a consistent set of C99 types, suggestion:
- size_t for length / size of memory
- uint8_t * for data pointers
- uint32_t for ssrc
- bool for boolean
- unsigned int

pabuhler:
Closing for now as the many changes have been done.

PR #675, opened 2024-01-05
use size_t for all memory length variables

pabuhler:
This change effects both public API and internal API and variables. It is meant to provide a consistent interface that reduces the amount of casts and increases readability of the code.

Part of #671.

### DIFF
--- a/crypto/cipher/aes_icm.c
+++ b/crypto/cipher/aes_icm.c
@@ -94,13 +94,13 @@ srtp_debug_module_t srtp_mod_aes_icm = {
  */
 
 static srtp_err_status_t srtp_aes_icm_alloc(srtp_cipher_t **c,
-                                            int key_len,
-                                            int tlen)
+                                            size_t key_len,
+                                            size_t tlen)
 {
     srtp_aes_icm_ctx_t *icm;
     (void)tlen;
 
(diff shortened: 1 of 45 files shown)


crypto/cipher/aes_icm.c
```diff
@@ -296,10 +296,10 @@ static void srtp_aes_icm_advance(srtp_aes_icm_ctx_t *c)
 
 static srtp_err_status_t srtp_aes_icm_encrypt(void *cv,
                                               unsigned char *buf,
-                                              unsigned int *enc_len)
+                                              size_t *enc_len)
 {
     srtp_aes_icm_ctx_t *c = (srtp_aes_icm_ctx_t *)cv;
-    unsigned int bytes_to_encr = *enc_len;
+    unsigned int bytes_to_encr = (unsigned int)*enc_len;
```
paulej, line 302: Why is this one not also `size_t`?

paulej (approved):

crypto/cipher/aes_icm.c
pabuhler: Because changing it led to a lot of other changes in in the aes code that I think require more understanding. Should maybe do it as it's own commit.

PR #677, opened 2024-01-08
remove last ekt reference

pabuhler:
Preserve the invalid policy test so more can be added.
Merge srtp_valid_policy and srtp_validate_policy_master_keys functions.

### DIFF
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -337,8 +337,6 @@ typedef struct srtp_policy_t {
                                    /**< this stream.                        */
     srtp_master_key_t **keys;      /** Array of Master Key structures       */
     unsigned long num_master_keys; /** Number of master keys                */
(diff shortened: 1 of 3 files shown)


paulej (approved):
Maybe we should create an EKT implementation for the public outside of libsrtp. I've written one in C++, but it likely needs work to make it robust. I do like EKT, though. Sadly, the RFC does not include conveying keys in RTCP, which is needed for a complete solution, IMO.

pabuhler:
> Maybe we should create an EKT implementation for the public outside of libsrtp. I've written one in C++, but it likely needs work to make it robust. I do like EKT, though. Sadly, the RFC does not include conveying keys in RTCP, which is needed for a complete solution, IMO.

Been a while since I looked at the final EKT draft, but not against even adding it back again if it makes sense. This was just clean up from before first.

PR #681, opened 2024-01-12
use uint8_t for byte and byte pointer

pabuhler:
This replaces "void *" in the API and usage of "unsigned char" as a byte type.

#671

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -105,7 +105,7 @@ static size_t srtp_get_rtp_xtn_hdr_len(const srtp_hdr_xtnd_t *xtn_hdr)
     return (ntohs(xtn_hdr->length) + 1u) * 4u;
 }
 
-static srtp_err_status_t srtp_validate_rtp_header(const void *rtp_hdr,
+static srtp_err_status_t srtp_validate_rtp_header(const uint8_t *rtp_hdr,
                                                   size_t pkt_octet_len)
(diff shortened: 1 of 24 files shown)


srtp/srtp.c
```diff
@@ -2039,16 +2037,15 @@ static srtp_err_status_t srtp_unprotect_aead(srtp_ctx_t *ctx,
     /*
      * Set the AAD for AES-GCM, which is the RTP header
      */
-    aad_len = (size_t)(enc_start - (uint8_t *)hdr);
-    status =
-        srtp_cipher_set_aad(session_keys->rtp_cipher, (uint8_t *)hdr, aad_len);
+    aad_len = (size_t)(enc_start - srtp_hdr);
```
paulej, line 2040: Does this actually produce a warning? I would assume this will convert without any issue.

paulej (approved):

srtp/srtp.c
pabuhler: No and looks like it is just left over from refactoring, good catch. I searched and there is a few more, will update pr before merging.

paulej: There are several. Line 1827, 1898, 2008, 2040 (this one), 2283, and 2652.  That's all I saw.

pabuhler: done

PR #683, opened 2024-01-13
remove support for gcm_x_8

pabuhler:
RFC 7714 only defines 16 byte tags for GCM.

This removes the config options to use 8 byte, it is still possible to do it but it should not be so easy as a predefined policy.


paulej (approved):
Changes look fine, but do you think there might be people using the shorter tags? While it's not in the final spec, it was in the drafts for a while.

pabuhler:
> Changes look fine, but do you think there might be people using the shorter tags? While it's not in the final spec, it was in the drafts for a while.

they might be but maybe this is what they need to wake to the fact that it is not standard ...

PR #684, opened 2024-01-14
Type usage consistency, braces, etc.

paulej:
This PR makes a number of changes to further use bool, size_t, ssize_t, etc. consistently through the code. I also added braces around single-line if() and for() statements. (As an aside, I actually like one-line statements, but when the clang-format rules force the line to break, I think braces should be employed. I'd be OK if we changed the clang-format file so that one-line statements stayed on a single line.)


crypto/include/datatypes.h
```diff
@@ -157,13 +163,13 @@ void v128_left_shift(v128_t *x, int shift_index);
     ((((x)->v32[(bit) >> 5]) &= ~((uint32_t)1 << ((bit)&31))))
 
 /*
- * srtp_octet_string_is_eq(a, b, len) returns 1 if the length len strings
- * a and b are not equal. It returns 0 otherwise. The running time of the
+ * srtp_octet_string_is_eq(a, b, len) returns true if the length len strings
+ * a and b are NOT equal. It returns false otherwise. The running time of the
```
pabuhler, line 167: consider renaming, srtp_byte_array_is_equal ?? ,  and making it more logical. This can be done in a new PR maybe to make it cleaner.

crypto/include/rdbx.h
```diff
@@ -109,7 +109,7 @@ srtp_err_status_t srtp_rdbx_dealloc(srtp_rdbx_t *rdbx);
  * index to which s corresponds, and returns the difference between
  * *guess and the locally stored synch info
  */
-int32_t srtp_rdbx_estimate_index(const srtp_rdbx_t *rdbx,
+ssize_t srtp_rdbx_estimate_index(const srtp_rdbx_t *rdbx,
```
pabuhler, line 112: not sure this should be ssize_t ...

pabuhler (approved):

pabuhler:
This is the last big part of #671

crypto/include/rdbx.h
paulej: Yeah, I'm not sure what we should use here. The int32_t type used before was fine. However, I saw a mix of types used (e.g., int used in rdbx_check_add() when receiving the result of the call to srtp_index_guess() that returned an int32_t, too).  This type indicates that it's signed and it arguably bigger than it needs to be.  Perhaps we should revisit this and create a "SeqDelta" type or something that is sized appropriately?  I'd prefer to not add more to this right now, but that might be useful.

crypto/include/datatypes.h
paulej: Yeah. we should do that and make the name align with the result expected.

PR #688, opened 2024-01-17
srtp_stream_list_ctx_t_: standarize naming

jmillan:
use 'capacity' and 'size' in order to represent the allocated storage and the number of elements respectively.

As per @Lastique comments [here](https://github.com/cisco/libsrtp/pull/508#issuecomment-1894441834)

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -4857,8 +4857,8 @@ typedef struct list_entry {
 
 typedef struct srtp_stream_list_ctx_t_ {
     list_entry *entries;
+    size_t capacity;
     size_t size;
-    size_t available;
 } srtp_stream_list_ctx_t_;
 
 srtp_err_status_t srtp_stream_list_alloc(srtp_stream_list_t *list_ptr)
@@ -4876,8 +4876,8 @@ srtp_err_status_t srtp_stream_list_alloc(srtp_stream_list_t *list_ptr)
         return srtp_err_status_alloc_fail;
     }
 
-    list->size = INITIAL_STREAM_INDEX_SIZE;
-    list->available = INITIAL_STREAM_INDEX_SIZE;
+    list->capacity = INITIAL_STREAM_INDEX_SIZE;
+    list->size = 0;
 
     *list_ptr = list;
 
@@ -4887,7 +4887,7 @@ srtp_err_status_t srtp_stream_list_alloc(srtp_stream_list_t *list_ptr)
 srtp_err_status_t srtp_stream_list_dealloc(srtp_stream_list_t list)
 {
     /* list must be empty */
-    if (list->available != list->size) {
+    if (list->size != 0) {
         return srtp_err_status_fail;
     }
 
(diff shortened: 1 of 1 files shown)


Lastique:
Unrelated to this PR, but you may want to consider changing the grow factor of the array to 1.5 (i.e. to `size_t new_capacity = list->capacity + (list->capacity + 1u) / 2u;`). See [here](https://github.com/facebook/folly/blob/main/folly/docs/FBVector.md#memory-handling) for a discussion why.

Also, while at it, you may want to check for capacity overflow, i.e. `if (new_capacity <= list->capacity) return fail;`.

jmillan:
@Lastique, nice comments. I'll do some modifications in this PR.

Lastique:
> Also, while at it, you may want to check for capacity overflow, i.e. `if (new_capacity <= list->capacity) return srtp_err_status_alloc_fail;`.

Actually, this condition is not quite correct. It should be something like this: `if ((new_capacity * sizeof(list_entry)) <= (list->capacity * sizeof(list_entry))) return srtp_err_status_alloc_fail;`.

jmillan:
The PR is now ready for review.

jmillan:
I've rebased to a single commit in order to have a better git history.

pabuhler (approved):

paulej:
> Unrelated to this PR, but you may want to consider changing the grow factor of the array to 1.5 (i.e. to `size_t new_capacity = list->capacity + (list->capacity + 1u) / 2u;`). See [here](https://github.com/facebook/folly/blob/main/folly/docs/FBVector.md#memory-handling) for a discussion why.

I read the material at the referenced link, but I fail to see the problem. If a vector has 100 elements and it grows by even 1, it will have to allocate a new buffer. If calling `realloc()`, there is a chance the buffer will remain in the same position, but just have additional memory available to the right of the existing buffer. In the worst case, it calls `malloc()` and moves all of the current data. What I fail to see is how this changes, regardless of C + 1, C * 1.5 ,or C * 2.

What am I missing?

jmillan:
> What am I missing?

`This means that any new chunk requested will be larger than all previously used chunks combined, so the vector must crawl forward in memory; it can't move back to its deallocated chunks. But any number smaller than 2 guarantees that you'll at some point be able to reuse the previous chunks
`

A factor of 2 will never let you to reuse previously used chunks. That will also contribute to memory fragmentation.

Lastique:
> Sure, but those chunks will be returned, and the OS will (ideally coalesce those into a larger contiguous chunk and) give those chunks to other application logic that requests memory.

This will at best only happen when the allocator has a whole unused page to free to the OS, which may not happen soon, depending on the vector sizes and the number of allocations. And that is assuming the allocator actually does release the page and not e.g. cache it for potential reuse in the future. Either way, reusing the memory within the application, without round tripping to the OS, is more efficient. The point of the 1.5 grow factor is that we *allow* for such reuse to happen, whereas with the factor of 2 we would not. It doesn't mean such reuse *will* happen, though.

paulej:
I have no strong objection either way. I was just confused by the claim. Now that I understand it better, I do think that claim, while not wrong on a pure sense, is wrong in reality since it doesn't take into account the fact that the system is doing lots of other things. It's more likely than not that previously freed memory is doing to be consumed by something else.  However, either works and I cannot prescribe what the best growth factor is. I suspect the best answer here is what works best for a typical, growing multi-party conference. I don't think there is a good answer, unless we know _a priori_ what to expect.

jmillan:
I have removed the 1.5 factor and applied back the previous 2.

This PR does two things:
* Use standard names in two struct members: (size, capacity). This is a readability & maintainability enhancement.
* Verify capacity overflow prior to requesting memory. This is a good practice and ultimately a bug fix.

As I get from the comments, changing the increase factor requires it's own space (PR) and corresponding rationale, data, etc.

paulej (approved):

PR #690, opened 2024-02-01
make MKI a stream level setting

pabuhler:
Move "use_mki" from a per packet parameter to a stream wide value. When using MKI all master keys need to have a valid MKI id.

Merge the MKI protect & unprotect functions. The MKI index is currently still passed per packet, this could also be moved up to stream level with a srtp_set_mki(ssrc) api as this is not likely to change per packet.

### DIFF
--- a/include/srtp.h
+++ b/include/srtp.h
@@ -338,6 +338,7 @@ typedef struct srtp_policy_t {
                                 /**< this stream.                        */
     srtp_master_key_t **keys;   /** Array of Master Key structures       */
     size_t num_master_keys;     /** Number of master keys                */
+    bool use_mki;               /** Whether MKI is in use                */
     size_t window_size;         /**< The window size to use for replay   */
                                 /**< protection.                         */
     bool allow_repeat_tx;       /**< Whether retransmissions of          */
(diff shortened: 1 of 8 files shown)


pabuhler:
This started as an attempt to merge the protect & unprotect api's but became a more fundamental change to have MKI as a mode rather than a per packet thing.

The test vector was generated from the 2.5 branch.

include/srtp.h
```diff
@@ -338,6 +338,7 @@ typedef struct srtp_policy_t {
                                 /**< this stream.                        */
     srtp_master_key_t **keys;   /** Array of Master Key structures       */
     size_t num_master_keys;     /** Number of master keys                */
+    bool use_mki;               /** Whether MKI is in use                */
```
paulej, line 341: Per https://www.rfc-editor.org/rfc/rfc3711#section-3.2.1, if MKI is used the length should remain unchanged for the duration of the cryptographic context. Should the `mki_size` be defined here instead of `srtp_master_key_t`?

pabuhler: I agree based on the RFC, but that is definitely not what was implemented before. I will change it has it does state "MUST" for the length being the same.

include/srtp_priv.h
```diff
@@ -135,6 +116,7 @@ typedef struct srtp_stream_ctx_t_ {
     uint32_t ssrc;
     srtp_session_keys_t *session_keys;
     size_t num_master_keys;
+    bool use_mki;
```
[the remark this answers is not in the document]
pabuhler: yip

srtp/srtp.c
```diff
@@ -1323,6 +1295,46 @@ srtp_err_status_t srtp_stream_init_keys(srtp_stream_ctx_t *srtp,
     return srtp_err_status_ok;
 }
 
+srtp_err_status_t srtp_stream_init_all_master_keys(srtp_stream_ctx_t *srtp,
+                                                   const srtp_policy_t *p)
+{
+    srtp_err_status_t status = srtp_err_status_ok;
+    if (p->key != NULL) {
+        if (p->use_mki) {
```
paulej, line 1303: Would a single key be a reason to not use MKI? Perhaps pointless, but illegal?

paulej (approved):

srtp/srtp.c
pabuhler: I think we will end up removing the single key option in the policy, it was only left there for ABI compatibility when MKI was first added.
It is "illegal" today as the single key option does not contain a mki id.

PR #692, opened 2024-03-20
Enables the use of wolfSSL for crypto primitives

SparkiDev:
To configure:
    ./configure --enable-wolfssl --with-wolfssl-dir=<wolfssl dir>

Add implementations of SRTP KDF, HMAC, AES-GCM and AES-CTR using wolfSSL.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,27 +82,38 @@ set(ENABLE_DEBUG_LOGGING OFF CACHE BOOL "Enable debug logging in all modules")
 set(ERR_REPORTING_STDOUT OFF CACHE BOOL "Enable logging to stdout")
 set(ERR_REPORTING_FILE "" CACHE FILEPATH "Use file for logging")
 set(ENABLE_OPENSSL OFF CACHE BOOL "Enable OpenSSL crypto engine")
+set(ENABLE_WOLFSSL OFF CACHE BOOL "Enable wolfSSL crypto engine")
 set(ENABLE_MBEDTLS OFF CACHE BOOL "Enable MbedTLS crypto engine")
 set(ENABLE_NSS OFF CACHE BOOL "Enable NSS crypto engine")
 
-if(ENABLE_OPENSSL OR ENABLE_MBEDTLS OR ENABLE_NSS)
+if(ENABLE_OPENSSL OR ENABLE_WOLFSSL OR ENABLE_MBEDTLS OR ENABLE_NSS)
   set(USE_EXTERNAL_CRYPTO TRUE)
 else()
   set(USE_EXTERNAL_CRYPTO FALSE)
 endif()
 
 if(ENABLE_OPENSSL)
-  if(ENABLE_NSS OR ENABLE_MBEDTLS)
-    message(FATAL_ERROR "ssl conflict. can not enable openssl and mbedtls or nss simultaneously.")
+  if(ENABLE_WOLFSSL OR ENABLE_NSS OR ENABLE_MBEDTLS)
+    message(FATAL_ERROR "ssl conflict. can not enable openssl and wolfssl, mbedtls or nss simultaneously.")
   endif()
   find_package(OpenSSL 1.1.0 REQUIRED)
   set(OPENSSL ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
   set(GCM ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
 endif()
 
+if(ENABLE_WOLFSSL)
+  if(ENABLE_OPENSSL OR ENABLE_NSS OR ENABLE_MBEDTLS)
+    message(FATAL_ERROR "ssl conflict. can not enable wolfssl and openssl, mbedtls or nss simultaneously.")
+  endif()
+  find_package(wolfSSL REQUIRED)
+  set(WOLFSSL ${ENABLE_WOLFSSL} CACHE BOOL INTERNAL)
+  set(WOLFSSL_KDF ${ENABLE_WOLFSSL} CACHE BOOL INTERNAL)
+  set(GCM ${ENABLE_WOLFSSL} CACHE BOOL INTERNAL)
+endif()
+
 if(ENABLE_MBEDTLS)
-  if(ENABLE_OPENSSL OR ENABLE_NSS)
-    message(FATAL_ERROR "ssl conflict. can not enable mbedtls and openssl or nss simultaneously.")
+  if(ENABLE_OPENSSL OR ENABLE_WOLFSSL OR ENABLE_NSS)
+    message(FATAL_ERROR "ssl conflict. can not enable mbedtls and openssl, wolfssl or nss simultaneously.")
   endif()
   find_package(MbedTLS REQUIRED)
   set(MBEDTLS ${ENABLE_MBEDTLS} CACHE BOOL INTERNAL)
@@ -110,8 +121,8 @@ if(ENABLE_MBEDTLS)
 endif()
 
 if(ENABLE_NSS)
-  if(ENABLE_OPENSSL OR ENABLE_MBEDTLS)
-    message(FATAL_ERROR "ssl conflict. can not enable nss and openssl or mbedtls simultaneously.")
+  if(ENABLE_OPENSSL OR ENABLE_WOLFSSL OR ENABLE_MBEDTLS)
+    message(FATAL_ERROR "ssl conflict. can not enable nss and openssl, wolfssl or mbedtls simultaneously.")
   endif()
   find_package(NSS REQUIRED)
   set(NSS ${ENABLE_NSS} CACHE BOOL INTERNAL)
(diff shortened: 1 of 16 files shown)


SparkiDev:
Can I get a review of this code?

pabuhler:
> Can I get a review of this code?

Yes it is on my list of things todo in the next week, sorry if that seams like a long time.
If possible can you add support for building with cmake as it is our main builld system these days and having a CI build for at least one build system and on at least one platform would be very helpful. see .github/workflows/cmake.yml or .github/workflows/autotools.yml

JonathanLennox:
Wrong things will happen if you enable two or more of --enable-openssl, --enable-wolfssl, and --enable-nss, and I think we have nothing that guards against this.  This isn't a new problem with wolfssl (it was already a problem with just openssl and nss), but should we have something to protect against this?

Or possibly even change the configure option syntax to --enable-crypto=[openssl,wolfssl,nss] with the 3.0 release, since that's where we're putting other breaking changes.

pabuhler:
> Wrong things will happen if you enable two or more of --enable-openssl, --enable-wolfssl, and --enable-nss, and I think we have nothing that guards against this. This isn't a new problem with wolfssl (it was already a problem with just openssl and nss), but should we have something to protect against this?
> 
> Or possibly even change the configure option syntax to --enable-crypto=[openssl,wolfssl,nss] with the 3.0 release, since that's where we're putting other breaking changes.

It think there is some protection in the cmake build files at least,  but I agree at this point when there are now ~5 options the config should be changed to a switch as you suggested

srtp/srtp.c
```diff
@@ -774,7 +787,87 @@ static srtp_err_status_t srtp_kdf_clear(srtp_kdf_t *kdf)
     return srtp_err_status_ok;
 }
 
-#else  /* if OPENSSL_KDF */
+#elif defined(WOLFSSL) && defined(WOLFSSL_KDF)
```
pabuhler, line 790: The OPENSSL_KDF is in practice never used ( at least not that I know off) and there has been some talk of removing it, are you sure you want to add wolfssl specific kdf functions?

crypto/include/aes_gcm.h
```diff
@@ -64,6 +64,35 @@ typedef struct {
...
+#include <wolfssl/options.h>
+#endif
+#include <wolfssl/wolfcrypt/settings.h>
+#include <wolfssl/wolfcrypt/aes.h>
+
+typedef struct {
+    int key_size;
+    int tag_len;
+    int aad_size;
+#ifndef WOLFSSL_AESGCM_STREAM
```
pabuhler, line 82: Which of the two api's is newer/preferred and do you really need to support both ?

SparkiDev: Prefer streaming but wolfSSL can be compiled without.
Some platforms don't have assembly code for streaming.

srtp/srtp.c
SparkiDev: Customers needing FIPS 140 validated cryptography for SRTP need to use an implementation of SRTP KDF in the boundary.
OpenSSL doesn't support SRTP KDF unless you get a fork as I understand it and has therefore that code has less value.

pabuhler: ok, have you verified that the tests pass with wolfssl kdf ? The one other time I have seen some one use an openssl kdf then it was broken.

SparkiDev: Yes, I have verified that the SRTP KDF in wolfSSL works in libsrtp.

SparkiDev:
Changed CI testing to use GitHub to clone wolfssl always using https.

srtp/srtp.c
pabuhler: I had just quick thought that it was why the test fails, but I see now it was build issues not test failures

pabuhler:
@SparkiDev ,
Hi I will look in to the failing Meson build, seams to also fail in the main branch at the moment. As to the other failures, maybe try to install "brew install automake" on macos and building wolfssl as a static library .... They are are just quick thoughts I have looked to deeply into the failures..

SparkiDev:
Changes:
Installing autoconf,, automake and libtool on make to build wolfSSL from source.
Invloking ldconfig on Ubuntu so that wolfSSL library is seen.

SparkiDev:
Attempted to fix cmake on macOS by exporting include path /usr/local/include

CMakeLists.txt
```diff
@@ -443,6 +465,8 @@ if(LIBSRTP_TEST_APPS)
             ${ENABLE_WARNINGS_AS_ERRORS})
     if(ENABLE_OPENSSL)
       target_include_directories(test_srtp PRIVATE ${OPENSSL_INCLUDE_DIR})
+    elseif(ENABLE_OPENSSL)
```
pabuhler, line 468: Shouldn't this be "elseif(ENABLE_WOLFSS)" ?

SparkiDev: Yes, it should have been!
Fixed and removed the change to to macOS builds to force /usr/local/include.

SparkiDev:
Changed WOLFSSL_INCLUDE_DIRS -> WOLFSSL_INCLUDE_DIR in FindwolfSSL.cmake to match CMakeLists.txt.

SparkiDev:
Changed when the Aes object is allocated to minimize memory usage.

SparkiDev:
Improved memory usage around AES-GCM.

crypto/cipher/aes_icm_wssl.c
```diff
@@ -0,0 +1,387 @@
...
+    if (icm->ctx == NULL) {
+        srtp_crypto_free(icm);
+        srtp_crypto_free(*c);
+        *c = NULL;
+        return srtp_err_status_alloc_fail;
+    }
+
+    err = wc_AesInit(icm->ctx, NULL, INVALID_DEVID);
+    if (err < 0) {
+        debug_print(srtp_mod_aes_icm, "wolfSSL error code: %d", err);
```
pabuhler, line 163: if this init fails then wc_AesFree will be called on the cxt in the dealloc function, is that the desired behavior?

crypto/hash/hmac_wssl.c
```diff
@@ -0,0 +1,225 @@
...
+    // allocate the buffer of wolfssl context.
+    (*a)->state = srtp_crypto_alloc(sizeof(Hmac));
+    if ((*a)->state == NULL) {
+        srtp_crypto_free(*a);
+        *a = NULL;
+        return srtp_err_status_alloc_fail;
+    }
+    err = wc_HmacInit((Hmac *)(*a)->state, NULL, INVALID_DEVID);
+    if (err < 0) {
+        debug_print(srtp_mod_hmac, "wolfSSL error code: %d", err);
```
pabuhler, line 96: this time wc_HmacFree will be called

pabuhler (commented):
In general it is fine, will approve once you have responded to the comments.

CMakeLists.txt
```diff
@@ -82,36 +82,46 @@ set(ENABLE_DEBUG_LOGGING OFF CACHE BOOL "Enable debug logging in all modules")
...
   set(GCM ${ENABLE_OPENSSL} CACHE BOOL INTERNAL)
 endif()
 
+if(ENABLE_WOLFSSL)
+  if(ENABLE_OPENSSL OR ENABLE_NSS OR ENABLE_MBEDTLS)
+    message(FATAL_ERROR "ssl conflict. can not enable wolfssl and openssl, mbedtls or nss simultaneously.")
+  endif()
+  find_package(wolfSSL REQUIRED)
+  set(WOLFSSL ${ENABLE_WOLFSSL} CACHE BOOL INTERNAL)
+  set(GCM ${ENABLE_WOLFSSL} CACHE BOOL INTERNAL)
```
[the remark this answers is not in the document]
SparkiDev: Was meant to be defined.
Fixed

config_in_cmake.h
```diff
@@ -18,6 +18,9 @@
 /* Define this to use OpenSSL crypto. */
 #cmakedefine OPENSSL 1
 
+/* Define this to use wolfSSL crypto. */
+#cmakedefine WOLFSSL 1
+
```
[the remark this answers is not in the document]
SparkiDev: Was meant to be defined.
Fixed

crypto/cipher/aes_icm_wssl.c
SparkiDev: Moved allocation and initialization down into srtp_aes_icm_wolfssl_context_init(().
Aes freed on init failure.

crypto/hash/hmac_wssl.c
SparkiDev: Fixed to free all allocated objects when init fails.

pabuhler:
@SparkiDev , it is merged but now the ci builds fail to build wolfssl ... https://github.com/cisco/libsrtp/actions/runs/8861716469/job/24333886680
Can you take a look? Thanks

SparkiDev:
https://github.com/cisco/libsrtp/pull/704 should fix this.

Issue #569, opened 2021-09-27
Please add a "non inplace" protect and unprotect API


ydroneaud:
In my use case, due to the inplace protect/unprotect API, RTP/RTCP packets have to be duplicated before being protected/unprotected.

I believe some memory bandwidth could be saved if libsrtp exposed an API that doesn't process data inplace.

I would imagine the API would be used this way

    int send(const void *pkt, size_t pkt)
    {
    ...
        srtp_protect2(ctx, rtp, rtplen, srtp, &srtpsize)
    ...
    }

    int receive(const void *pkt, size_t pkt)
    {
    ...
        srtp_unprotect2(ctx, srtp, srtplen, rtp, &rtpsize)
    ...
    }

bifurcation:
If protect/unprotect were not in-place, wouldn't there just be a copy going on internal to the protect/unprotect call instead of at the application layer?

ydroneaud:
> If protect/unprotect were not in-place, wouldn't there just be a copy going on internal to the protect/unprotect call instead of at the application layer?

It depends on the underliyng crypto primitive implementation.

bifurcation:
Fair point.  You would at least have to copy the header over, since that's not touched by the crypto.

paulej:
You want libsrtp to allocate memory and copy the packet data?  That strikes me as less desirable than doing that in your code since you could, at the very least, create a static buffer to avoid reallocating memory repeatedly.

bifurcation:
No, I think the request is to allow the application to provide separate input and output buffers.  So the application would continue to own the memory.

paulej:
OK. Then how is that different from:

```c
memcpy(srtp_buffer, input_buffer, sizeof(input_buffer));
srtp_protect(srtp_buffer);
```

Just the desire of libsrtp to do the memcpy?

bifurcation:
No, the point is that you can do encryption without any memcpy.  Consider the following cases:

```
void xor(uint8_t *dst, const uint8_t *srcA, const uint8_t *srcB, size_t len) {
    for (size_t i = 0; i < len; i++) {
        dst[i] ^= srcA[i] ^ srcB[i];
    }
}

uint8_t plaintext[BUFFER_SIZE];
uint8_t ciphertext[BUFFER_SIZE];
uint8_t key[BUFFER_SIZE];

// CASE 1: today
xor(plaintext, plaintext, key, BUFFER_SIZE);

// CASE 2: today, with memcpy
[...]

Issue #322, opened 2017-07-26
provide an new version of protect function that does not corrupt memory


fluffy:
I think we should also have a protect that passes in the size of the buffer. It's just better practice.

pabuhler:
This is now supported

PR #693, opened 2024-03-25
Support a not in-place api's for protect/unprtect

pabuhler:
The protect/unprotect api's can now operate both in-place or not in-place, depending on the requirements of the caller.
The length of the out buffer can now be checked to ensure there is sufficient space.
This addresses #322 "provide an new version of protect function that does not corrupt memory" and #569 "Please add a "non inplace" protect and unprotect API"

srtp_driver has been updated so all existing test can be run either in-place or not in-place.

- [x] Document new api and error codes
- [x] Update examples to use new api
- [x] Create separate PR's for all clean up / refactoring commits
- [ ] Follow up with test cleanup and reuse of CHECK macros

### DIFF
--- a/crypto/cipher/aes_gcm_nss.c
+++ b/crypto/cipher/aes_gcm_nss.c
@@ -337,8 +337,8 @@ static srtp_err_status_t srtp_aes_gcm_nss_encrypt(void *cv,
 {
     srtp_aes_gcm_ctx_t *c = (srtp_aes_gcm_ctx_t *)cv;
 
-    // When we get a non-NULL buffer, we know that the caller is
-    // prepared to also take the tag.  When we get a NULL buffer,
+    // When we get a non-NULL src buffer, we know that the caller is
+    // prepared to also take the tag.  When we get a NULL src buffer,
     // even though there's no data, we need to give NSS a buffer
     // where it can write the tag.  We can't just use c->tag because
     // memcpy has undefined behavior on overlapping ranges.
@@ -357,6 +357,10 @@ static srtp_err_status_t srtp_aes_gcm_nss_encrypt(void *cv,
         cv, true, non_null_buf, src_len, non_null_dst_buf, dst_len);
     if (status != srtp_err_status_ok) {
         return status;
+    }
+
+    if (*dst_len < c->tag_size) {
+        return srtp_err_status_bad_param;
     }
 
     memcpy(c->tag, non_null_dst_buf + (*dst_len - c->tag_size), c->tag_size);
--- a/crypto/cipher/aes_gcm_nss.c
+++ b/crypto/cipher/aes_gcm_nss.c
@@ -337,12 +337,6 @@ static srtp_err_status_t srtp_aes_gcm_nss_encrypt(void *cv,
 {
     srtp_aes_gcm_ctx_t *c = (srtp_aes_gcm_ctx_t *)cv;
 
-    //#todo, this might need som looking at
-    // nss requires space for tag, currently we assume that ther is space, this
-    // should change, the best would be to merge the cipher encrypt and get_tag
-    // api
-    *dst_len += 16;
-
     // When we get a non-NULL buffer, we know that the caller is
     // prepared to also take the tag.  When we get a NULL buffer,
     // even though there's no data, we need to give NSS a buffer
--- a/crypto/cipher/aes_gcm_nss.c
+++ b/crypto/cipher/aes_gcm_nss.c
@@ -281,8 +281,10 @@ static srtp_err_status_t srtp_aes_gcm_nss_set_aad(void *cv,
 
 static srtp_err_status_t srtp_aes_gcm_nss_do_crypto(void *cv,
                                                     bool encrypt,
-                                                    uint8_t *buf,
-                                                    size_t *enc_len)
+                                                    const uint8_t *src,
+                                                    size_t src_len,
+                                                    uint8_t *dst,
+                                                    size_t *dst_len)
 {
     srtp_aes_gcm_ctx_t *c = (srtp_aes_gcm_ctx_t *)cv;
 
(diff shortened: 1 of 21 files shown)


pabuhler:
This PR is set to draft state as it is still a work in progress and can potentially be force updated with breaking changes.

pabuhler:
@paulej

test/srtp_driver.c
```diff
@@ -160,19 +188,151 @@ srtp_master_key_t *test_keys[2] = {
...
+        status = srtp_unprotect(ctx, srtp, srtp_len, srtp, rtp_len);
+    }
+    return status;
+}
+
+srtp_err_status_t call_srtp_unprotect(srtp_ctx_t *ctx,
+                                      uint8_t *srtp,
+                                      size_t *srtp_len)
+{
+    return call_srtp_unprotect2(ctx, srtp, *srtp_len, srtp_len);
```
paulej, line 256: Why not just have `call_srtp_unprotect()` do the work of `call_srtp_unprotect2()`?

crypto/cipher/aes_gcm_nss.c
```diff
@@ -328,32 +330,43 @@ static srtp_err_status_t srtp_aes_gcm_nss_do_crypto(void *cv,
...
     // where it can write the tag.  We can't just use c->tag because
     // memcpy has undefined behavior on overlapping ranges.
     uint8_t tagbuf[16];
-    uint8_t *non_null_buf = buf;
-    if (!non_null_buf && (*enc_len == 0)) {
+    const uint8_t *non_null_buf = src;
+    uint8_t *non_null_dst_buf = dst;
+    if (!non_null_buf && (src_len == 0)) {
         non_null_buf = tagbuf;
+        non_null_dst_buf = tagbuf;
```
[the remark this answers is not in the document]
pabuhler: Yes, if the src buffer is null, it is an indication that this is being used for authentication only and no data will be encrypted.
The api today does not expect the tag to be written to dst at this point, see #714

crypto/cipher/aes_gcm_nss.c
```diff
@@ -328,32 +330,43 @@ static srtp_err_status_t srtp_aes_gcm_nss_do_crypto(void *cv,
...
+                                                  uint8_t *dst,
+                                                  size_t *dst_len)
 {
     srtp_aes_gcm_ctx_t *c = (srtp_aes_gcm_ctx_t *)cv;
 
+    //#todo, this might need som looking at
+    // nss requires space for tag, currently we assume that ther is space, this
+    // should change, the best would be to merge the cipher encrypt and get_tag
+    // api
+    *dst_len += 16;
```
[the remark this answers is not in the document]
pabuhler: Removing the entire section as it is no longer needed now that dst_len contains space for the tag.

crypto/cipher/aes_icm.c
```diff
@@ -295,12 +295,20 @@ static void srtp_aes_icm_advance(srtp_aes_icm_ctx_t *c)
...
+                                              uint8_t *dst,
+                                              size_t *dst_len)
 {
     srtp_aes_icm_ctx_t *c = (srtp_aes_icm_ctx_t *)cv;
-    size_t bytes_to_encr = *enc_len;
+    size_t bytes_to_encr = src_len;
     uint32_t *b;
+    const uint32_t *s;
+
+    // check out length if not equal or greater bail!
```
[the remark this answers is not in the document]
pabuhler: added now

crypto/cipher/null_cipher.c
```diff
@@ -116,13 +116,19 @@ static srtp_err_status_t srtp_null_cipher_set_iv(void *cv,
...
 static srtp_err_status_t srtp_null_cipher_encrypt(void *cv,
-                                                  uint8_t *buf,
-                                                  size_t *bytes_to_encr)
+                                                  const uint8_t *src,
+                                                  size_t src_len,
+                                                  uint8_t *dst,
+                                                  size_t *dst_len)
 {
-    /* srtp_null_cipher_ctx_t *c = (srtp_null_cipher_ctx_t *)cv; */
     (void)cv;
```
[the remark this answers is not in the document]
pabuhler: yip

test/srtp_driver.c
pabuhler: Wanted to keep the existing call_srtp_unprotect() signature and try to remove the xxx2() functions in a followup cleanup commit, but i see now this one could have easily been merge already now.

crypto/cipher/aes_gcm_nss.c
```diff
@@ -339,21 +343,24 @@ static srtp_err_status_t srtp_aes_gcm_nss_encrypt(void *cv,
...
+    srtp_err_status_t status = srtp_aes_gcm_nss_do_crypto(
+        cv, true, non_null_buf, src_len, non_null_dst_buf, dst_len);
     if (status != srtp_err_status_ok) {
         return status;
     }
 
-    memcpy(c->tag, non_null_buf + (*enc_len - c->tag_size), c->tag_size);
-    *enc_len -= c->tag_size;
+    memcpy(c->tag, non_null_dst_buf + (*dst_len - c->tag_size), c->tag_size);
+    *dst_len -= c->tag_size;
```
paulej, line 363: This sets the length in the formal param `dst_len` to be whatever it was coming in minus the tag size.  What if `dst_size` was 0 or 1, for example?  Since I cannot see any data coming out when the `src` buffer is NULL, is this just indicating how much buffer would remain if a tag had been produced?  (I understand a tag is somehow produced for something and stored in `c->tag`, but I don't see the relationship with that `dst` parameter when the `src` is NULL.)

crypto/cipher/aes_gcm_nss.c
```diff
@@ -387,11 +394,22 @@ static srtp_err_status_t srtp_aes_gcm_nss_get_tag(void *cv,
...
-                                                  uint8_t *buf,
-                                                  size_t *enc_len)
+                                                  const uint8_t *src,
+                                                  size_t src_len,
+                                                  uint8_t *dst,
+                                                  size_t *dst_len)
 {
-    srtp_err_status_t status =
-        srtp_aes_gcm_nss_do_crypto(cv, false, buf, enc_len);
+    uint8_t tagbuf[16];
```
paulej, line 402: While no less confusing to me than the `encrypt()` function, it might be good to replicate much of the comments. That said, it would be good if the comments on `encrypt()` would speak about which "NULL buffer" (src or dst). The `When we get a non-NULL buffer` doesn't indicate which buffer.

paulej (approved):

crypto/cipher/aes_gcm_nss.c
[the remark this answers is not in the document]
pabuhler: in this case the entire packet was pass as aad data previously with srtp_aes_gcm_nss_set_aad()

crypto/cipher/aes_gcm_nss.c
pabuhler: you are right that this is confusing and if we do #714 "consider merging the gcm encrypt and get tag functions of cipher api" once this is merged it will clean all this up alot.

As it is now though, if the call to srtp_aes_gcm_nss_do_crypto() succeeds then dst_len was large enough for the tag and any encrypted data, it then would have been update to be that size. Since currently the tag is not going to be returned from this function it is truncated by tag size. But I agree that a sanity check against tag size would have be good here.

crypto/cipher/aes_gcm_nss.c
pabuhler: I guess that is an old comment, will update.

PR #698, opened 2024-04-08
allow cmake to use imported `mbedtls` library

dmllr:
`find_library` in CMake always tries to find the library installed, but in the case when the same library was build from source as a CMake dependency, `find_library` would fail.

To fix the issue, CMakeLists.txt is changed to call `find_library` for mbedtls only if needed.


pabuhler:
"ld: symbol(s) not found for architecture x86_64",  i guess it needs to check if the architecture mathes as well ?

pabuhler:
Did not mean to close this :(

pabuhler:
@dmllr i have tried to find if this is the correct thing to be doing or not but I have not found any good documentation. Do you have any references? If it is correct than I guess it should be don't for all dependencies. I will continue to try and find some answers.

PR #704, opened 2024-04-29
Fix CI so that wolfSSL builds with appropriate ASM for macOS chip

SparkiDev:


.github/workflows/autotools.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       if: matrix.os == 'ubuntu-latest' && matrix.crypto == 'wolfssl'
       run: |
         git clone https://github.com/wolfSSL/wolfssl
+        git checkout v5.7.0-stable
         cd wolfssl
```
pabuhler, line 40: I guess the order matter here :)

pabuhler (commented):
Thanks for looking at this so quickly!

Issue #710, opened 2024-05-02
Failure to update key after mki failure


pbodilis:
As a receiver, I fail to update srtp keys with MKI using the srtp_stream_update
This is new to master, this works fine with version 2.6.0

scenario:
1. init srtp with ssrc_any_inbound, use_mki is true
2. try to unprotect a packet with ssrc0 and associated mki0 => failure
3. add srtp key with matching mki0 => success
4. try to unprotect  a packet with ssrc0 and associated mki0 => success
5. try to unprotect a packet with ssrc1 and associated mki1 => failure
6. add srtp key with matching mki1 => failure
7. update srtp with srtp key with matching mki1 => **failure**

I suspect the bug to be around `update_template_stream_cb` or `srtp_stream_remove`.

`srtp_unprotect` creates a new stream with `srtp_stream_clone`, ssrc is stored in network order

```

static bool update_template_stream_cb(srtp_stream_t stream, void *raw_data)
{
    struct update_template_stream_data *data =
        (struct update_template_stream_data *)raw_data;
    srtp_t session = data->session;
    uint32_t ssrc = stream->ssrc; // ssrc is in network order
...


    /* remove stream */
    data->status = srtp_stream_remove(session, ssrc);
    if (data->status) {
        return false;
    }



```

but then, `srtp_strp_remove` expect ssrc to be given in host order

so possible fix can be to replace the call in `update_template_stream_cb` to be consistent with the changes made in the API:
`
    data->status = srtp_stream_remove(session, ntohl(ssrc));
`

PR: https://github.com/cisco/libsrtp/pull/711

pbodilis:
Scenario logs:

```


           sw_srtp_log_handler|[srtp] alloc: (location: 0x6030000297d0) allocated
           sw_srtp_log_handler|[srtp] alloc: (location: 0x603000029800) allocated
           sw_srtp_log_handler|[srtp] alloc: (location: 0x603000029830) allocated
           sw_srtp_log_handler|[srtp] alloc: (location: 0x603000029860) allocated
           sw_srtp_log_handler|[srtp] alloc: (location: 0x603000029890) allocated
           sw_srtp_log_handler|[srtp] alloc: (location: 0x6030000298c0) allocated
           sw_srtp_log_handler|[srtp] alloc: (location: 0x603000029920) allocated
           sw_srtp_log_handler|[srtp] alloc: (location: 0x603000029950) allocated
[...]

PR #711, opened 2024-05-02
[code] fix endianness issue while updating srtp keys

pbodilis:
This is bound to the issue described here:

https://github.com/cisco/libsrtp/issues/710


pabuhler (approved):
srtp_stream_remove is also wrong in stream_update function, if you like you can change that as well or else I will do it after. Will look to add a unit test to catch this, since it is oblivious missing.

pbodilis:
Done

Issue #714, opened 2024-05-26
consider merging the gcm encrypt and get tag functions of cipher api


pabuhler:
The gcm encryption is a one shot function, it is always only called only once. The tag is often generated at this point by the underling crypto engine  (see aes_gcm_nss.c), it would be clear to append the tag to the out immediately instead of caching it. This would match the decrypt function.

PR #718, opened 2024-06-26
change srtp_cipher_encrypt to append the tag generated

pabuhler:
Change for #714.
This makes it symmetric with the srtp_cipher_decrypt function that will remove the tag.
Currently most of the backends would have cached the tag internally and returned it in the srtp_cipher_get_tag function, this removes that extra complexity.

### DIFF
--- a/crypto/cipher/aes_gcm_wssl.c
+++ b/crypto/cipher/aes_gcm_wssl.c
@@ -330,7 +330,7 @@ static srtp_err_status_t srtp_aes_gcm_wolfssl_encrypt(void *cv,
     srtp_aes_gcm_ctx_t *c = (srtp_aes_gcm_ctx_t *)cv;
     int err;
 
-    if (c->dir != srtp_direction_encrypt && c->dir != srtp_direction_decrypt) {
+    if (c->dir != srtp_direction_encrypt) {
         return srtp_err_status_bad_param;
     }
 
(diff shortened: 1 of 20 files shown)


crypto/cipher/aes_gcm_wssl.c
```diff
@@ -405,7 +386,11 @@ static srtp_err_status_t srtp_aes_gcm_wolfssl_decrypt(void *cv,
     int err;
 
     if (c->dir != srtp_direction_encrypt && c->dir != srtp_direction_decrypt) {
-        return (srtp_err_status_bad_param);
+        return srtp_err_status_bad_param;
+    }
+
+    if (*dst_len < src_len - c->tag_len) {
```
paulej, line 392: Is the subtraction correct?

paulej (approved):
I had one question, and it might just be my misunderstanding.

crypto/cipher/aes_gcm_wssl.c
pabuhler: This is a decrypt function so the tag len will be removed from the src, so the dst_len must be greater then or equal to the src_len minus the tag_len .
There should probably be a pre check that src_len >= tag_len, will add that

paulej: Yeah, that made sense this morning as I looked at it.

paulej (approved):

PR #725, opened 2024-09-30
add test for ROC mіsmatch

fippo:
illustrating the reason why one should carefully pick the initial sequence number low enough to avoid a rollover.

See
  https://webrtc-review.googlesource.com/c/src/+/358360

### DIFF
--- a/format.sh
+++ b/format.sh
@@ -20,7 +20,7 @@ done
 m=`git ls-files -m`
 if [ -n "$m" ]; then
     v=`$CLANG_FORMAT -version`
-    echo "Fromatting required when checking with $v"
+    echo "Formatting required when checking with $v"
     echo
     echo "The following files required formatting:"
     for f in $m; do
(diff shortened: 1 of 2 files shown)


fippo:
Quite interesting that not everything fails... it fails when using OpenSSL but not in "normal" / builtin mode?!
The reason for the failure is that the failed attempt to unprotect modifies pkt1 and pkt2 which is quite surprising.

Avoided by modifying the test *slightly*, see second commit.

pabuhler:
@fippo Hi, I was curious why it failed with openssl, in the end it was not so exciting. When built with openssl the test uses the GCM cipher which does the decryption & authentication at the same time so it does not fail until it is complete.
Will merge now and then work on adding support for trying ROC -1 when check new streams.

pabuhler (approved):

fippo:
ah, that makes sense, thanks for figuring that out!

PR #733, opened 2024-12-06
Fix srtp_unprotect_rtcp_mki when RTP auth != RTCP

vopatek:
srtp_get_session_keys, which is used by both srtp_unprotect_mki and srtp_unprotect_rtcp_mki, determines the tag len from rtp_auth.

This fails when rtp_auth differ from rtcp_auth. E.g. when SRTP is used with weak authentication but SRTCP must not (RFC 3711).

This commit splits the function in two:
srtp_get_session_keys_rtp
srtp_get_session_keys_rtcp

And adds a short auth policy test to test/srtp_driver.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -1638,23 +1638,15 @@ static void srtp_calc_aead_iv(srtp_session_keys_t *session_keys,
     v128_xor(iv, &in, &salt);
 }
 
-srtp_session_keys_t *srtp_get_session_keys(srtp_stream_ctx_t *stream,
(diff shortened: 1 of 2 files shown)


pabuhler:
Thanks @vopatek  this does seam to be an issue, I will fist fix up the CI build problem and then we can try to merge.

pabuhler:
Hi @vopatek ,

I have now merged this into the 2_x_dev branch and created a pr #737 that cherry picks this into main. Great if you could take a look and leave a comment if it looks ok.

Thanks!

PR #737, opened 2024-12-11
Fix srtp_unprotect_rtcp_mki when RTP auth != RTCP

pabuhler:
srtp_get_session_keys, which is used by both srtp_unprotect_mki and srtp_unprotect_rtcp_mki, determines the tag len from rtp_auth.

This fails when rtp_auth differ from rtcp_auth. E.g. when SRTP is used with weak authentication but SRTCP must not (RFC 3711).

This commit splits the function in two:
srtp_get_session_keys_rtp
srtp_get_session_keys_rtcp

And adds a short auth policy test to test/srtp_driver.

(cherry picked from commit 63a19f4f28b01789023ef528460feb586e352656)

Cherry picked from #733

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -1744,10 +1744,11 @@ static void srtp_calc_aead_iv(srtp_session_keys_t *session_keys,
     v128_xor(iv, &in, &salt);
 }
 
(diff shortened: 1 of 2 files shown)


vopatek (approved):
Looks good. Thanks!

PR #740, opened 2025-01-14
WolfSSL aes icm multi encrypt fix

pabuhler:
The current wolfssl integration fails to encrypt multiple packets correctly on Apple silicon with "--enable-armasm".
A multi packet test is added to cipher_driver to verify this.

libSRTP will allocate and initialize a cipher once and then use it to encrypt or decrypt multiple packets. Between each operation a new IV will be set, the IV is unique for each packet. This test will ensure that cipher implementations support this.

The test vectors are based on examples in RFC 9335 .


pabuhler:
@SparkiDev could you please review this PR since you added the wolfssl integration, thanks!

SparkiDev (approved):

Issue #744, opened 2025-02-08
Building against BoringSSL fails, easy fix


kmansoft:
Hello,

I'm trying to build libSRTP against BoringSSL using CMake.

Running into an issue in that CMakeLists.txt requires a specific OpenSSL version - 1.1.0 or newer - and BoringSSL doesn't define a version number, so the `FindPackage` here fails:

https://github.com/cisco/libsrtp/blob/main/CMakeLists.txt#L103

https://github.com/Kitware/CMake/blob/master/Modules/FindOpenSSL.cmake#L576

https://boringssl.googlesource.com/boringssl/+/HEAD/include/openssl/opensslv.h

Since OpenSSL 1.1.0 is from 25 Aug 2016, and was EOL'd in September 2019, I think it's safe to assume that any OpenSSL version would be compliant.

Would it be possible to remove the version requirement from CMakeLists.txt like this:

Current:

```
find_package(OpenSSL 1.1.0 REQUIRED)
```

Proposed fix:

```
find_package(OpenSSL REQUIRED)
```

Thank you for your consideration!

kmansoft:
Submitted a PR - https://github.com/cisco/libsrtp/pull/745

PR #745, opened 2025-02-11
Don't require OpenSSL version number as BoringSSL doesn't define it

kmansoft:
Fixes https://github.com/cisco/libsrtp/issues/744


pabuhler:
This makes some sense to me, if we are going to remove that then should probable remove the checks from the other build files as well.

pabuhler:
Have fixed the CiFuzz build upstream.

pabuhler (approved):
This makes sense to me.

kmansoft:
Thank you for merging!

PR #746, opened 2025-02-11
Fix function signature mismatch for `srtp_remove_stream`

vikramdattu:
- The definition for the function has the signature: `srtp_err_status_t srtp_remove_stream(srtp_t session, uint32_t ssrc)`
 - This is not always same as the signature present in srtp.h: `srtp_err_status_t srtp_remove_stream(srtp_t session, unsigned int ssrc)`

This causes function signature mismatch.
Aligned the signature from `srtp.h` with the definition using `uint32_t`


pabuhler (approved):
Have fixed the two build failures upstream, so they can be ignored.

Issue #631, opened 2022-12-16
check if libdl and libz are really needed


pabuhler:
In this case |I found that it was mileading mesage. If you will look closet on the bottom openssl detection
https://github.com/cisco/libsrtp/blob/cc362ae5b3d30c89d8dc00c22f84f615fb02d956/configure.ac#L208-L256
you can fiund libdl and libz detections.
I had no installed libz devel resources and that message was printed by incorrecvt logic in tjhat section
Looks like libz detection can be removed because nothing in source tree is usimg libz.
```console
[tkloczko@devel-g2v libsrtp-2.4.2]$ grep zlib.h -r
[tkloczko@devel-g2v libsrtp-2.4.2]$
```
Whole section could be dramatically reduced if `PKG_CHECK_MODULES()` aclocal macro would be used.

_Originally posted by @kloczek in https://github.com/cisco/libsrtp/issues/599#issuecomment-1338268618_

pabuhler:
Follow up on @kloczek 's comments in #599

PR #750, opened 2025-02-14
remove openssl version check

pabuhler:
This is a follow on from #744 / #745 but for other build systems.

Now that OpenSSL 1.1 is EOL there is no need to
enforce version checks.

This does not remove support for building against
OpenSSL 1.1 .

### DIFF
--- a/configure
+++ b/configure
@@ -4789,6 +4789,118 @@ done
 
 CFLAGS="$supported_cflags"
 
+{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for ANSI C header files" >&5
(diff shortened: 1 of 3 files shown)


paulej (approved):
This builds fine, but I noted that the zlib development libraries must be installed for the configure script to work, though it doesn't appear to really be necessary. It's unclear why this requirement exists.  CMake does not impose the same requirement. Perhaps it's something intrinsic to autotools? Not sure.

pabuhler:
Thanks @paulej, I guess you comments are related to #631 , will try to resolve before v3.0 is released.

PR #757, opened 2025-05-30
Properly support null crypto and null auth scenario

freemangordon:
Currently, it is silently assumed that we always have master key provided, even when the policy says we don't use encryption and authentication. This results in streams with such policy to fail.

Fix that by properly calculating the expected master key length by taking in account authentication type too. Correctly set expected key length for null crypto policy to 0. Do not assume we have HMAC if the policy is set to not have.

Add a test case for null/null policy while at it.


freemangordon:
This shall be back-ported to 2.5 (which is in debian stable) and newer releases, as they are currently broken in null/null scenario, IIUC. Do I have to make pull requests for those versions?

pabuhler:
> This shall be back-ported to 2.5 (which is in debian stable) and newer releases, as they are currently broken in null/null scenario, IIUC. Do I have to make pull requests for those versions?

Once this PR is landed then either you or I can back port to 2_x_dev in order to create what ever 2.x releases are needed.

pabuhler:
@freemangordon I will attempt to review the PR in depth over the weekend. I do agree that it should be fixed mostly as it is a regression. But going forward we may consider making it "harder" or requiring more explicit configure to allow libsrtp to be configured in a totally insecure manor, I think as a security library we have some responsibility to prevent users from accidentally configuring it in an insecure manor.

pabuhler (approved):

pabuhler:
@freemangordon  I made one small change but otherwise I think this is ok.  If there are no more comments then will merge shortly.

PR #768, opened 2025-09-22
Improve installation paths on windows

hmaarrfk:
This helps windows libraries get installed to

```
bin\srtp2.dll
lib\srtp2.lib
```

This is the split we use at conda-forge to help keep the dlls findable while keeking the lib files separate.

Let me know if you have any concerns.

Excited to see this upstreamed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -317,12 +317,7 @@ if(WIN32)
   target_compile_definitions(srtp3 PUBLIC _CRT_SECURE_NO_WARNINGS)
 endif()
 
-install(TARGETS srtp3
-  EXPORT libSRTPTargets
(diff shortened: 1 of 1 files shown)


pabuhler:
Hi @hmaarrfk, this should be ok. I wish these build systems where more standardized, and sometimes your wishes come true :)  ... according to https://cmake.org/cmake/help/latest/command/install.html the built-in defaults values should be as you suggested so maybe the best change is to just remove the "DESTINATION lib". I have checked that it works on linux, if it also works for you maybe you can change the PR?

hmaarrfk:
lets see what happens https://github.com/conda-forge/libsrtp2-feedstock/pull/2

PR #774, opened 2025-11-12
Make code work using NSS in FIPS-mode

msirringhaus:
This is part of an effort to make webRTC calls work under FIPS in Firefox. (See [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1797270) and more specifically [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1922760))

How to test:
```
meson setup -Dcrypto-library=nss build
export NSS_FIPS=1
meson test -C build
```
Half the tests fail without this patch and all work with it (tests also work with unset `NSS_FIPS`).

`PK11_ImportSymKey()` is not allowed to be used under FIPS, but there can be a workaroud to still be able to import raw key material, by encrypting it first and then unwrapping it (which is in the end a NO-OP, but satisfies the API under FIPS).

### DIFF
--- a/crypto/cipher/aes_gcm_nss.c
+++ b/crypto/cipher/aes_gcm_nss.c
@@ -222,7 +222,8 @@ static srtp_err_status_t srtp_aes_gcm_nss_context_init(void *cv,
          * Note: the caller is now responsible for the proper FIPS usage of the
          * key material!
          */
-        c->key = import_sym_key_in_FIPS(slot, CKM_AES_GCM, CKA_ENCRYPT, &key_item);
+        c->key =
+            import_sym_key_in_FIPS(slot, CKM_AES_GCM, CKA_ENCRYPT, &key_item);
     } else {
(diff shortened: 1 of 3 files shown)


crypto/include/nss_fips.h
```diff
@@ -0,0 +1,155 @@
...
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
+ * are met:
+ *
+ *   Redistributions of source code must retain the above copyright
+ *   notice, this list of conditions and the following disclaimer.
+ *
+ *   Redistributions in binary form must reproduce the above
```
pabuhler: This is not a public header in this project so will not be (and should not be) installed in binary installations, how to ensure this condition is meet?

msirringhaus: I got permission from the original author to remove the restrictive part of the header and make it public. See https://bugzilla.mozilla.org/show_bug.cgi?id=1922760#c3

PR #790, opened 2026-02-08
update policy config api

pabuhler:
As part of the API changes for 3.0, would like to simplify the policy API as well as making it easier to extend in the future. The current API has far to many details and can be cumbersome to set up and validate.

Main changes:
- move from a public struct to a opaque handle
- use profiles to initialize crypto policy
- remove detailed config options, can be added if needed
- use explicit key size's
- remove policy list
- add tests

Either a single policy can be used with srtp_create() or else multiple policies can be used with multiple calls to srtp_add_stream() .

A minimal example would be
```
srtp_policy_t policy;
srtp_policy_create(&policy);
srtp_policy_set_ssrc(policy, (srtp_ssrc_t){ ssrc_any_outbound, 0 });
srtp_policy_set_profile(policy, srtp_profile_aes128_cm_sha1_80);
srtp_policy_set_key(policy, master_key, sizeof(master_key), master_salt, sizeof(master_salt));
srtp_create(&srtp, policy); 
```

It could be nice to remove the set_ssrc() function if a suitable default behavior could be found.


pabuhler:
The implementation is currently just a new front end to the existing srtp_policy_t API, the intention would be to eventually remove the old structure and functions from the public API.

srtp/srtp_policy.c
```diff
@@ -0,0 +1,235 @@
...
+
+srtp_err_status_t srtp_policy2_create(srtp_policy2_t *policy)
+{
+    srtp_policy2_t p;
+
+    if (policy == NULL) {
+        return srtp_err_status_bad_param;
+    }
+
+    p = (srtp_policy2_t)srtp_crypto_alloc(sizeof(*p));
```
paulej: `p` is a pointer and not yet assigned, right? If so, `*p` is an invalid pointer dereference.

paulej: Here, `p` is an uninitialized pointer, right? If so, `*p` is an invalid pointer dereference.

paulej (commented):
Initial review

paulej (commented):
Initial review

paulej (commented):
Initial review

srtp/srtp_policy.c
pabuhler: compile time only ... but I get your point

srtp/srtp_policy.c
```diff
@@ -0,0 +1,235 @@
...
+        p->master_keys[i].key = p->master_key_store[i].key;
+        p->master_keys[i].mki_id = p->master_key_store[i].mki_id;
+        p->keys[i] = &p->master_keys[i];
+    }
+    p->legacy.keys = p->keys;
+
+    // setup hdr xtnd id's
+    p->legacy.enc_xtn_hdr = p->enc_hdr_xtnd_ids;
+
+    *policy = p;
```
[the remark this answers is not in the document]
pabuhler: pointer to a point

pabuhler:
@paulej , I have update with the major changes still some small things to sort out but progress can be reviewed, feedback appreciated.

fuzzer/fuzzer.c
```diff
@@ -254,113 +254,81 @@ static uint8_t *extract_key(const uint8_t **data,
...
 
-    if (simulate == true) {
-        *success = false;
+    if (srtp_policy_add_key(policy, key, key_size, key, 0, NULL, 0) ==
+        srtp_err_status_ok) {
+        added = true;
     }
 
+    fuzz_free(key);
+    return added;
```
paulej, line 283: Why not just return false and get rid of the "added" variable, given explicit return values are used before this one?

fuzzer/fuzzer.c
```diff
@@ -254,113 +254,81 @@ static uint8_t *extract_key(const uint8_t **data,
...
-        uint8_t do_extract_master_key;
-        EXTRACT_IF(&do_extract_master_key, *data, *size,
-                   sizeof(do_extract_master_key));
-
-        if ((do_extract_master_key % 2) == 0) {
-            break;
+    if (mki_id_size > 0) {
+        status = srtp_policy_use_mki(policy, mki_id_size);
+        if (status != srtp_err_status_ok) {
+            goto end;
```
paulej, line 326: `goto` statements were removed, but I see one added here. Any reason to not `return false`?

fuzzer/fuzzer.c
```diff
@@ -254,113 +254,81 @@ static uint8_t *extract_key(const uint8_t **data,
...
-        ret[i] = extract_master_key(data, size, key_size, false, NULL);
-
-        if (ret[i] == NULL) {
-            /* Shouldn't happen */
-            abort();
-        }
+    status =
+        srtp_policy_add_key(policy, key, key_size, key, 0, mki, mki_id_size);
+    if (status != srtp_err_status_ok) {
+        goto end;
```
paulej, line 333: same `goto` question

include/srtp.h
```diff
@@ -291,71 +291,267 @@ typedef struct {
...
+/*
+ * @brief identifies a particular SRTP profile
+ *
+ * An srtp_profile_t enumeration is used to identify a particular SRTP
+ * profile (that is, a set of algorithms and parameters).
+ */
+typedef enum {
+    srtp_profile_reserved = 0,
+    srtp_profile_null_null =
+        1, // TODO: for testing and troubleshooting only ??? check rfc3711
```
paulej: Regarding that TODO, it is either this or configuration to bypass SRTP entirely when debugging packet flows. I can see this useful for devs and customers who want to disable SRTP.

srtp/srtp_policy.c
```diff
@@ -0,0 +1,392 @@
...
+            return srtp_err_status_bad_param;
+        }
+    }
+
+    if (policy->window_size != 0 &&
+        (policy->window_size < 64 || policy->window_size >= 0x8000)) {
+        return srtp_err_status_bad_param;
+    }
+
+    // TODO: cryptex and encrypted hdr extensions can not both be enabled
```
paulej: Whatchagonnadoaboutit?

srtp/srtp_policy.c
```diff
@@ -0,0 +1,392 @@
...
+}
+
+srtp_err_status_t srtp_policy_set_window_size(srtp_policy_t policy,
+                                              size_t window_size)
+{
+    if (policy == NULL) {
+        return srtp_err_status_bad_param;
+    }
+
+    if (window_size != 0 && (window_size < 64 || window_size >= 0x8000)) {
```
paulej: This line is here and the srtp_policy_validate() function has the same. Does a window size of zero turn it off? I forgot. And is the policy passed into here considered valid? I didn't look to see what called this, but it made me wonder how many other checks are duplicated.

paulej (commented):
This thing is really long. My eyes glazed over. :) I didn't see anything that screamed of trouble.

fuzzer/fuzzer.c
pabuhler: This was an intermediate step, new code extracts a key and a salt so it is nice to have clean up, and it matches the rest of the code in this file

fuzzer/fuzzer.c
pabuhler: cleanup

fuzzer/fuzzer.c
pabuhler: clean up

include/srtp.h
pabuhler: yes I have keep it but there is a requirement that no key is set, so not sure if that breaks your use case

srtp/srtp_policy.c
pabuhler: add test and check in policy validation

srtp/srtp_policy.c
pabuhler: This one could probably be removed for the validate function, but not sure it hurt to be checked twice. Could maybe make helps in the code so the logic is ensured to be consistant.

In general now the policy as a whole is only validated once, so a lot of other checks in srtp code could be removed. But later maybe ?

pabuhler:
All major planned changes are completed to support the new config/policy API, the API can be polished later once the main part has been accepted.

srtp/srtp_policy.c
paulej: I didn't mean to suggest removing the check. I just meant to ask if we should have a function this and the other function calls that like `srtp_is_window_valid()` (This could be a static inline private to this file.)

paulej (approved):

srtp/srtp_policy.c
pabuhler: added helper func

pabuhler:
This change changes the config api completely, there will likely be follow changes as the API is used issues discovered. A review before 3.0 release will be done as part of #672 .

PR #805, opened 2026-03-26
Add API to require cryptex

### LINKED ISSUE #804: Allow cryptex to be mandatory-to-use
follow-up to #777

https://mailarchive.ietf.org/arch/msg/avt/pGnete_6IqExxTERNl44_9GqVRU/
made a good point: one might interpret (which the W3C spec currently does not) cryptex as mandatory to use.
In this mode srtp_unprotect would fail with an error.

The error returned *might* be `srtp_err_status_cryptex_err` which would use it in a different context than currently but that might be ok.

This would require
* adding `int require_cryptex` to the private `srtp_stream_ctx_t_`,
* adding `srtp_set_stream_require_cryptex` similar to `srtp_set_stream_use_cryptex`
* having `srtp_cryptex_unprotect_init` return the error if `require_cryptex` is true but `inuse` is false and there are header extensions.
* adding a test that srtp_unprotect fails with plaintext extensions

I **think** that plan works for both 2.8+ and 3.x

Happy to take a stab if that sounds like a plan, main question I have is whether it is ok to reuse the error.

fippo:
allowing applications to reject packets not using cryptex using
  srtp_set_stream_require_cryptex(...)

as described in
  https://datatracker.ietf.org/doc/html/rfc9335#section-5.2

Fixes #804 (for 2_x_dev)

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -5194,8 +5194,7 @@ srtp_err_status_t srtp_set_stream_require_cryptex(srtp_t session,
         }
         stream->require_cryptex = enable != 0;
         break;
-    case ssrc_any_inbound:
-    case ssrc_any_outbound: {
+    case ssrc_any_inbound: {
         struct set_cryptex_from_template_data data;
 
         if (session->stream_template == NULL) {
@@ -5207,6 +5206,8 @@ srtp_err_status_t srtp_set_stream_require_cryptex(srtp_t session,
                                   set_cryptex_from_template_cb, &data);
         break;
     }
+    case ssrc_any_outbound:
+        // Requiring cryptex is not possible for outbound SSRCs, fall through.
     default:
         return srtp_err_status_bad_param;
     }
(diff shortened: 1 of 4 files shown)


srtp/srtp.c
```diff
@@ -5171,6 +5176,44 @@ srtp_err_status_t srtp_set_stream_use_cryptex(srtp_t session,
...
+    switch (ssrc->type) {
+    case ssrc_specific:
+        stream = srtp_get_stream(session, htonl(ssrc->value));
+        if (stream == NULL) {
+            return srtp_err_status_bad_param;
+        }
+        stream->require_cryptex = enable != 0;
+        break;
+    case ssrc_any_inbound:
+    case ssrc_any_outbound: {
```
fippo: not clear what the behavior should be for ssrc_any_outbound, maybe return srtp_err_status_bad_param?

pabuhler: error sounds fine

pabuhler (approved):
I think this looks ok

fippo:
Updated, let me know if I should squash after review!

pabuhler (approved):

pabuhler:
No need to squash unless you want to. Will merge next week.

PR #813, opened 2026-06-01
mbedTLS 4 / PSA Crypto support on 2_x_dev

### LINKED ISSUE #812: mbedTLS 4 support on 2_x_dev — plans?
## Context

PR #782 landed mbedTLS 4 (PSA Crypto) support on `main`, which has since been renamed to libsrtp3 (v3.0.0). The 2.x line on `2_x_dev` stayed on the legacy `<mbedtls/aes.h>` / `<mbedtls/gcm.h>` path, so v2.8.0 and later 2.x consumers still cannot build against mbedTLS 4.

This is starting to bite for the embedded / IoT world: ESP-IDF v6 (currently in pre-release, GA imminent) ships mbedTLS 4 by default. Anything still on the libsrtp 2.x line — and that's the mainstream of WebRTC consumers today, including the AWS Kinesis Video Streams C SDK — can't compile against IDF v6.

## Question

What is the project's position on backporting PSA Crypto / mbedTLS 4 support to `2_x_dev`?

Specifically:

- Is `2_x_dev` considered feature-frozen (security fixes only), with 3.x expected as the migration path?
- If a backport gated on `MBEDTLS_VERSION_MAJOR >= 4` (or `>= 3` for the full PSA path) is acceptable in principle, we'd be happy to send a PR. A direct cherry-pick of #782 isn't clean — the 3.x release also changed the internal cipher signature (in-place → src/dst buffers), so the backport needs adaptation to keep the 2.x cipher API intact for existing consumers.

Either answer is fine — we just want to know whether to invest the work or accept the v6 ceiling on the 2.x line.

Happy to share the conflict list and a backport sketch if it helps the conversation.

vikramdattu:
Backport of mbedTLS 4 PSA Crypto. Gated under `#if MBEDTLS_VERSION_MAJOR >= 4`; legacy mbedTLS 3 path unchanged.

Local: both mbedTLS 3.6.4 and 4.0.0 → 10/10 ctest pass.

Closes #812. cc @pabuhler

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -62,14 +62,18 @@ jobs:
       if:  matrix.os == 'ubuntu-latest' && matrix.crypto == 'mbedtls4'
       run: |
         python3 -m pip install --break-system-packages --upgrade jsonschema jinja2
-        git clone https://github.com/Mbed-TLS/mbedtls.git mbedtls4-src
-        cd mbedtls4-src
+        git clone https://github.com/Mbed-TLS/mbedtls.git /tmp/mbedtls4-src
+        cd /tmp/mbedtls4-src
         git checkout mbedtls-4.0.0
         git submodule update --init --recursive
-        cmake -S . -B build -DENABLE_TESTING=OFF -DENABLE_PROGRAMS=OFF
+        # PIC is required to link the static libtfpsacrypto into libsrtp2's
+        # shared library; without it the final shared link fails with
+        # "relocation R_X86_64_PC32 ... can not be used when making a shared
+        # object; recompile with -fPIC".
+        cmake -S . -B build -DENABLE_TESTING=OFF -DENABLE_PROGRAMS=OFF \
+              -DCMAKE_POSITION_INDEPENDENT_CODE=ON
         cmake --build build -j
         sudo cmake --install build
-        cd ..
 
     - name: Setup macOS OpenSSL
       if: matrix.os == 'macos-latest' && matrix.crypto == 'openssl'
@@ -94,16 +98,20 @@ jobs:
     - name: Setup macOS MbedTLS 4
       if:  matrix.os == 'macos-latest' && matrix.crypto == 'mbedtls4'
       run: |
+        # Install under /tmp/ (outside $GITHUB_WORKSPACE) so the install
+        # survives the actions/checkout@v2 step that runs later in this job
+        # and clears the workspace.
         python3 -m pip install --break-system-packages --upgrade jsonschema jinja2
-        git clone https://github.com/Mbed-TLS/mbedtls.git mbedtls4-src
-        cd mbedtls4-src
+        git clone https://github.com/Mbed-TLS/mbedtls.git /tmp/mbedtls4-src
+        cd /tmp/mbedtls4-src
         git checkout mbedtls-4.0.0
         git submodule update --init --recursive
-        cmake -S . -B build -DENABLE_TESTING=OFF -DENABLE_PROGRAMS=OFF -DCMAKE_INSTALL_PREFIX="$GITHUB_WORKSPACE/mbedtls4-prefix"
+        cmake -S . -B build -DENABLE_TESTING=OFF -DENABLE_PROGRAMS=OFF \
+              -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+              -DCMAKE_INSTALL_PREFIX=/tmp/mbedtls4-prefix
         cmake --build build -j
         cmake --install build
-        echo "CMAKE_PREFIX_PATH=$GITHUB_WORKSPACE/mbedtls4-prefix" >> $GITHUB_ENV
-        cd ..
+        echo "CMAKE_PREFIX_PATH=/tmp/mbedtls4-prefix" >> $GITHUB_ENV
 
     - uses: actions/checkout@v2
 
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        crypto: [internal, openssl, openssl3, nss, mbedtls]
+        crypto: [internal, openssl, openssl3, nss, mbedtls, mbedtls4]
         exclude:
           - os: windows-latest
             crypto: openssl
@@ -24,6 +24,8 @@ jobs:
             crypto: nss
           - os: windows-latest
             crypto: mbedtls
+          - os: windows-latest
+            crypto: mbedtls4
           - os: ubuntu-latest
             crypto: openssl3
         include:
@@ -36,6 +38,8 @@ jobs:
           - crypto: nss
             cmake-crypto-enable: "-DENABLE_NSS=ON"
           - crypto: mbedtls
+            cmake-crypto-enable: "-DENABLE_MBEDTLS=ON"
+          - crypto: mbedtls4
             cmake-crypto-enable: "-DENABLE_MBEDTLS=ON"
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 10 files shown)


vikramdattu:
Dear @pabuhler can you please take a look?

pabuhler:
@vikramdattu , I started to look at this but I am no mbedTLS expert. So it is hard to say if the the usage is correct. I compare abit to main branch and use AI to review, it comes up with a few concerns.
How confident are you that the mbedTLS API usage is correct ?

vikramdattu:
Dear @pabuhler I am confident enough with the change. Also, I have added the mbedtls test path for the change. One thing that makes me comfortable is, I have kept the default to existing (mbedtls3) with the mbedtls4 as an opt-in. So, the change should be safe to go if it passes the test suite for mbedtls4.

I request you to do the top-level/API review. AI review will also do. I can then take this forward. 

PS: Adding @harshal5 for the in-depth review who is one of the security engineers from Espressif Systems.

Harshal5:
@Ashish285, could you please help by taking a look at this as well?

crypto/hash/hmac_mbedtls.c
```diff
@@ -80,6 +97,27 @@ static srtp_err_status_t srtp_hmac_mbedtls_alloc(srtp_auth_t **a,
     if (*a == NULL) {
         return srtp_err_status_alloc_fail;
     }
+#if MBEDTLS_VERSION_MAJOR >= 4
+    {
+        psa_hmac_ctx_t *state;
+        psa_status_t status = psa_crypto_init();
```
Harshal5, line 103: Can we move the `psa_crypto_init()` call to some srtp's crypto initialization API (if it exists)? So, that we could avoid adding calls to `psa_crypto_init()` at multiple places.
Multiple calls to `psa_crypto_init()` is functionally fine as the API exits early in such cases:
```
psa_status_t psa_crypto_init(void)
{
    psa_status_t status;

    /* Double initialization is explicitly allowed. Early out if everything is
     * done. */
    if (psa_get_initialized()) {
        return PSA_SUCCESS;
    }
    .
    .
}
```
So this is more of a code maintenance improvement than a functional requirement.

crypto/hash/hmac_mbedtls.c
```diff
@@ -177,12 +307,34 @@ static srtp_err_status_t srtp_hmac_mbedtls_compute(void *statev,
...
-    if (mbedtls_md_hmac_update(statev, message, msg_octets) != 0)
-        return srtp_err_status_auth_fail;
-
-    if (mbedtls_md_hmac_finish(state, hash_value) != 0)
-        return srtp_err_status_auth_fail;
+#if MBEDTLS_VERSION_MAJOR >= 4
+    {
+        psa_hmac_ctx_t *state = (psa_hmac_ctx_t *)statev;
+        size_t out_len = 0;
+
```
Harshal5, line 314: Could we add a check like we have:
```
if (msg_octets < 0) {
}
```
as done [above](https://github.com/cisco/libsrtp/pull/813/changes#diff-e4d09d245f9f832a6331fee230180fb08d8cc5aff4d79afa9fca1644ffc4925cR270) in the `srtp_hmac_mbedtls_update()` API?

Harshal5 (commented):
Unlike the old per-context `mbedtls_*` path, PSA stores keys in a shared global key store. If `libsrtp` processes streams from multiple threads, `mbedTLS` should be built with `MBEDTLS_THREADING_C` enabled.

This isn't a blocker, but it's probably worth documenting somewhere.

Other than that, LGTM!

crypto/hash/hmac_mbedtls.c
pabuhler: we are trying to move a way from having an explicit init function in libsrtp, not sure how that will end up. Maybe some thing that can be improved in the future.

.github/workflows/cmake.yml
```diff
@@ -54,6 +58,23 @@ jobs:
       if:  matrix.os == 'ubuntu-latest' && matrix.crypto == 'mbedtls'
       run: sudo apt-get install libmbedtls-dev
 
+    - name: Setup Ubuntu MbedTLS 4
+      if:  matrix.os == 'ubuntu-latest' && matrix.crypto == 'mbedtls4'
+      run: |
+        python3 -m pip install --break-system-packages --upgrade jsonschema jinja2
+        git clone https://github.com/Mbed-TLS/mbedtls.git /tmp/mbedtls4-src
+        cd /tmp/mbedtls4-src
+        git checkout mbedtls-4.0.0
```
Ashish285: Is there any specific reason to use `mbedtls-4.0.0`? `mbedtls-4.1.0` is available and is the LTS release

crypto/cipher/aes_icm_mbedtls.c
```diff
@@ -246,10 +289,46 @@ static srtp_err_status_t srtp_aes_icm_mbedtls_context_init(void *cv,
...
+             * and on PSA implementations that eagerly invalidate op key
+             * handles, destroy_key would leave the op in a bad state. */
+            psa_cipher_abort(&c->ctx->op);
+            c->ctx->op = psa_cipher_operation_init();
+            psa_destroy_key(c->ctx->key_id);
+            c->ctx->key_id = PSA_KEY_ID_NULL;
+        }
+
+        status =
+            psa_import_key(&attr, key, key_size_in_bits / 8, &(c->ctx->key_id));
```
Ashish285, line 320: After the key import, it is recommended to reset the key attributes with `psa_reset_key_attributes(attr)`

crypto/hash/hmac_mbedtls.c
vikramdattu: Agreed it's a maintenance improvement rather than a correctness issue — the repeated `psa_crypto_init()` calls are safe because of that early-out.

As @pabuhler notes, libsrtp is moving away from an explicit init entry point, so adding one now to centralize this would cut against that direction. I've kept the per-backend `psa_crypto_init()` calls for this reason. If/when libsrtp settles on a crypto-init hook, consolidating these is a straightforward follow-up.

vikramdattu:
Pushed one more fix (`3747a63`): **AES-ICM in-place `psa_cipher_update`**.

While integrating this branch as an ESP-IDF v6 component — which ships **TF-PSA-Crypto**, not mainline mbedTLS — `srtp_init()` failed the AES-ICM cipher self-test with `srtp_err_status_cipher_fail`.

Root cause: the CTR path calls `psa_cipher_update` **in place** (SRTP always encrypts with input buffer == output buffer), and TF-PSA-Crypto rejects an overlapping in/out buffer for AES-CTR on **non-block-aligned lengths** with `PSA_ERROR_INVALID_ARGUMENT`. Mainline mbedTLS 4.1.0 (what this PR's CI/ctest build against) *permits* the overlap, so the self-test is green there — this only surfaces on TF-PSA-Crypto.

The fix stages the CTR update through a bounded stack buffer so input and output don't overlap. AES-CTR is a stream cipher here (`out_len == in_len`, no deferred bytes), so processing in fixed-size block-aligned chunks preserves the keystream position across calls. The GCM path (`psa_aead_update`) is unaffected — I verified it handles overlapping buffers at non-aligned lengths on the same backend. The full libsrtp `host_test` now passes end-to-end on ESP-IDF v6.

crypto/hash/hmac_mbedtls.c
vikramdattu: This is done

.github/workflows/cmake.yml
vikramdattu: Updated to 4.1

crypto/cipher/aes_icm_mbedtls.c
vikramdattu: Added psa_reset_key_attributes call

Harshal5:
LGTM!

vikramdattu:
@pabuhler may I request your attention on the PR please? PTAL and comment anything obvious.

pabuhler:
> @pabuhler may I request your attention on the PR please? PTAL and comment anything obvious.

I will review tomorrow, sorry for delay.

pabuhler (approved):

PR #814, opened 2026-07-03
RFC 4771 ROC in-band support

volovolov:
### Summary

This PR implements the RFC 4771 *Roll-over Counter (ROC) Carrying* integrity transform for SRTP. It lets a sender periodically embed its 32-bit ROC in the SRTP authentication tag of selected packets, so a late-joining or re-synchronizing receiver can recover the correct ROC without having to guess it from the RTP sequence number. All three RFC 4771 modes are supported, with mode 3 layered on top of the AES-GCM AEAD transform per RFC 7714 §8.2.

RCC is **opt-in and fully backward compatible**: streams that do not enable it (`srtp_rcc_mode_none`, the default) behave exactly as before and use the standard RFC 3711 implicit-ROC transform.

### Motivation

Under RFC 3711 the ROC is never transmitted; a receiver infers it from the RTP sequence number. A receiver that joins mid-stream, loses synchronization, or misses a sequence-number wrap can end up with the wrong ROC and fail authentication on every subsequent packet. RFC 4771 solves this by carrying the ROC in-band on a configurable subset of packets ("ROC-carrying" packets, those whose RTP sequence number is `≡ 0 (mod R)`).

### What's implemented

- **Mode 1** — Only ROC-carrying packets are integrity-protected; other packets carry no authentication tag.
- **Mode 2** — ROC-carrying packets use the RCC tag; all other packets use the standard integrity transform. (Most common / robust choice.)
- **Mode 3** — NULL-MAC variant: the transform carries only the 4-octet ROC with no MAC of its own. Supported on top of AES-GCM (RFC 7714), where the AEAD tag authenticates the packet and the ROC feeds the GCM IV.

For a ROC-carrying packet in modes 1/2, the tag is built as:

```
TAG = ROC (4 octets) || MAC_tr
```

where `MAC_tr` is the `(auth_tag_len - 4)` most-significant octets of the HMAC-SHA1. This requires `auth_tag_len >= 5`.

### Packet layout

**Modes 1/2 (AES-CM + HMAC-SHA1), ROC-carrying packet:**

```
[ RTP header ][ encrypted payload ][ MKI (optional) ][ ROC(4) || MAC_tr ]
```

**Mode 3 (AES-GCM, RFC 7714 §8.2), ROC-carrying packet:**

```
[ RTP header ][ ciphertext incl. 16-octet GCM tag ][ MKI (optional) ][ ROC(4) ]
```

The 4-octet ROC is the SRTP authentication-tag field and is placed **after** the MKI. It is *not* part of the GCM AAD (AAD remains the RTP header only); instead the ROC feeds the GCM IV, so any tampering with the carried ROC produces a wrong IV and a GCM authentication failure — giving implicit integrity protection of the ROC.

### API changes

New public enum and function in `include/srtp.h`:

```c
typedef enum {
    srtp_rcc_mode_none = 0,   /* default RFC 3711 transform */
    srtp_rcc_mode_1    = 1,
    srtp_rcc_mode_2    = 2,
    srtp_rcc_mode_3    = 3
} srtp_rcc_mode_t;

srtp_err_status_t srtp_policy_set_rcc_mode_tx_rate(
    srtp_policy_t policy, srtp_rcc_mode_t rcc_mode, uint16_t roc_tx_rate);
```

- `roc_tx_rate` (R) is the transmission rate; `R == 1` carries the ROC in every packet. Must be `>= 1` when RCC is enabled.
- Both peers must derive the same mode and rate so they agree on the packet layout.
- Validation enforces cipher/mode consistency: modes 1/2 require an AES-CM profile and `auth_tag_len >= 5`; mode 3 requires an AES-GCM profile. Inconsistent combinations are rejected by `srtp_create()` / policy validation.

### Implementation notes

- Modes 1/2 are handled by dedicated `srtp_protect_rcc()` / `srtp_unprotect_rcc()` paths in `srtp/srtp.c`.
- Mode 3 is handled inside `srtp_protect_aead()` / `srtp_unprotect_aead()`, since GCM dispatch happens ahead of the RCC dispatch.
- On a ROC-carrying packet the receiver adopts the sender's ROC (`srtp_rdbx_set_roc_seq()` + re-anchoring the replay index); non-carrying packets use the normal estimated packet index and replay check.
- MKI lookup: because the ROC now follows the MKI in mode 3, the unprotect dispatch passes `srtp_len - 4` for ROC-carrying packets so the existing "MKI is last" lookup still locates the MKI correctly. Carry-packet detection uses only `seq % R == 0` and the configured mode (no session keys required).

### Testing

Added `test/rcc_test.c` covering:

- Modes 1 and 2 round-trip protect/unprotect (AES-CM + HMAC-SHA1).
- Mode 3 round-trip over AES-GCM.
- ROC recovery on a receiver starting from a mismatched ROC.
- Tamper detection for mode 3 (modifying the carried ROC yields `srtp_err_status_auth_fail`).

All tests pass. Mode 3 tests require a build with GCM enabled (`-DENABLE_OPENSSL=ON`).

### Backward compatibility

- Default behavior is unchanged: `srtp_rcc_mode_none` streams use the standard RFC 3711 transform.
- No changes to existing wire format for non-RCC streams.
- New API is purely additive.

### Limitations / open items

- Modes 1/2 are defined only for AES-CM ciphers; mode 3 only for AES-GCM (matching the RFCs).
- RCC applies to SRTP (RTP); SRTCP is unaffected.


pabuhler:
@volovolov , if you rebase on main then the wolfssl build failures should be fixed.

volovolov:
> @volovolov , if you rebase on main then the wolfssl build failures should be fixed.

@pabuhler Thank you, it's done now!

pabuhler:
@volovolov , looks like something else broke :( ... it is at least not related to your changes, I will follow it up

test/rcc_test.c
```diff
@@ -0,0 +1,605 @@
+/*
```
pabuhler, line 1: Great that you have added these tests, some feedback though.

This test is not run any where, it needs to be added as a test target for at least cmake builds, but ideally also autotools and mason.

The style of tests in this file does not match either srtp_driver.c or test_srtp_policy.c , it would be good to not add a new test style. My preference would be to do it in the same way as test_srtp_policy.c

srtp/srtp.c
```diff
@@ -1642,6 +1646,92 @@ static srtp_err_status_t srtp_stream_init(srtp_stream_ctx_t *srtp,
...
+            if (!is_gcm) {
+                return srtp_err_status_bad_param;
+            }
+            break;
+        default:
+            return srtp_err_status_bad_param;
+        }
+    }
+
+    /*
```
pabuhler: This is a duplicate block of code, one should be deleted.
This validation should also maybe be done as part of srtp_policy_validate().

srtp/srtp_policy.c
```diff
@@ -1022,6 +1022,20 @@ srtp_err_status_t srtp_policy_get_mki_length(srtp_policy_t policy,
...
 }
 
+srtp_err_status_t srtp_policy_set_rcc_mode_tx_rate(srtp_policy_t policy,
+                                                   srtp_rcc_mode_t rcc_mode,
+                                                   uint16_t roc_tx_rate)
+{
+    if (policy == NULL) {
+        return srtp_err_status_bad_param;
+    }
+
```
pabuhler, line 1100: should you check that roc_tx_rate >= 1 when mode != none ?

pabuhler (changes requested):
Just an initial review, have not fully reviewed the main change yet.

srtp/srtp.c
volovolov: Thanks, makes sense indeed to move the validation into srtp_policy_validate(), did that.

test/rcc_test.c
volovolov: I tried to do all of that, please review the latest changes in rcc_test.c

srtp/srtp_policy.c
volovolov: Agree, done now.

srtp/srtp.c
```diff
@@ -2905,6 +3588,16 @@ srtp_err_status_t srtp_unprotect(srtp_t ctx,
                                    rtp_len, session_keys, advance_packet_index);
     }
 
+    /*
+     * If RFC 4771 RCC is enabled for this stream, dispatch to the RCC handler
+     * which reads the sender's ROC from the authentication tag.
+     */
+    if (stream->rcc_mode != srtp_rcc_mode_none &&
+        stream != ctx->stream_template) {
```
pabuhler: Why this check?  Looking at the tests there are none for ssrc_any_inbound, is that meant to be supported?
If it is a template stream and the ROC is already greater than 1 think it will always fail authentication?
Maybe need to add the stream duplication code in  srtp_unprotect_rcc ?
I think it would work for the GCM case but not here.

pabuhler (changes requested):
Hi, 

Sorry for the delay in reviewing, I had some concerns regarding srtp_protect_rcc & srtp_unprotect_rcc and needed the time to look into it properly.
I was concerned that the new functions duplicates a lot of code and would  increase maintenance overhead, but I understand that it would be alot of work to refactor the original code to support both and that it is maybe better done afterwards in small steps.

It also looks like there is no replay protection of ROC caring packets in mode 1&2, this means that when a ROC caring packet is received it always succeeds and reset the replay index. The ROC caring packet should only be accepted if it advances the replay index, otherwise it is possible to replay an older sequence.
Also if the ROC caring packet is aa new packet but false with in the current tracked window it should just update the window and not reset it.
If need I can help provide test case that show these problems.

volovolov:
@pabuhler Thank you for your latest feedback! I tried to address all of it, please have a look. There a no longer dedicated srtp_protect_rcc / srtp_unprotect_rcc, and issues concerning cases of ssrc_any_inbound and replay protection in modes 1-3 should be addressed, with new tests adding respective coverage. Best regards, Volodymyr

include/srtp.h
```diff
@@ -387,6 +387,37 @@ srtp_err_status_t srtp_policy_get_profile(srtp_policy_t policy,
...
+ * is congruent to 0 modulo the transmission rate @c roc_tx_rate, "R").
+ * For a packet that carries the ROC the tag is built as
+ * @c TAG @c = @c ROC(4 @c octets) @c || @c MAC_tr, where @c MAC_tr is the
+ * @c (auth_tag_len @c - @c 4) most significant octets of the HMAC.
+ *
+ * Modes 1 and 2 use the HMAC-SHA1 integrity transform and are defined only
+ * for the AES-CM ciphers.  Mode 3 is the RFC 4771 NULL-MAC variant: it
+ * carries only the 4-octet ROC with no MAC of its own.  Mode 3 is supported
+ * here on top of AES-GCM (RFC 7714); the AEAD tag authenticates the packet
+ * and the ROC is appended immediately after the GCM tag.
```
pabuhler: if MKI is enabled then the ordering will be GCM tag | MKI | ROC , not sure how important it is to be correct here.

srtp/srtp.c
```diff
@@ -2777,19 +2888,31 @@ srtp_err_status_t srtp_protect(srtp_t ctx,
 
         /* run auth func over ROC, put result into auth_tag */
         debug_print(mod_srtp, "estimated packet index: %016" PRIx64, est);
-        status = srtp_auth_compute(session_keys->rtp_auth, (uint8_t *)&est, 4,
-                                   auth_tag);
+        if (rcc_carry) {
+            /* RFC 4771: TAG = ROC (4 octets, network order) || MAC_tr */
+            uint8_t mac[SRTP_MAX_TAG_LEN];
+            status = srtp_auth_compute(session_keys->rtp_auth, (uint8_t *)&est,
```
pabuhler, line 2898: not sure why this extra temp buffer was needed, couldn't the normal tag be written to the out buffer and then the just the first 4 bytes be overwriten with ROC ?

pabuhler (approved):
Couple of comments but otherwise looks ok

include/srtp.h
volovolov: Thanks for noticing! I've gone through the comments mentioning ROC placement in the AES-GCM case and made appropriate corrections.

srtp/srtp.c
volovolov: Got rid of the temp buffer.

PR #818, opened 2026-08-01
fix AEAD SRTCP out-of-place buffer overflow

pabuhler:
Authenticate E=0 packets before copying and exclude the GCM tag from the output. Add a regression test for contract-sized output buffers.

Reported by afldl, 2026-07


paulej (approved):
LGTM

PR #820, opened 2026-09-02
srtp: detect SSRC collisions on RCC unprotect

pabuhler:
Deferring index estimation for ROC-carrying packets also skipped the sender-direction check, so an RCC sender stream would accept its own packets instead of reporting a collision.

Small follow up to #814


pabuhler:
@volovolov this is a small follow up to you #814 PR, what do you think?

volovolov:
@pabuhler Makes total sense, thanks for the fix and the respective tests!

PR #821, opened 2026-09-06
srtp: locate the MKI correctly for RCC mode 1 packets without a tag

lenamonj:
With RFC 4771 RCC mode 1 and an MKI, every packet that does not carry the ROC failed `srtp_unprotect` with `bad_mki`: `srtp_protect` appends no authentication tag to those packets, but the key lookup still stepped a full tag length back from the packet end before reading the MKI.

`srtp_get_session_keys_for_rtp_packet` now takes the packet's `rcc_carry` state and derives what trails the MKI from it: the 4-octet ROC of a mode 3 ROC-carrying AES-GCM packet, nothing for a mode 1 packet without the ROC, the full tag otherwise. The ad-hoc mode 3 subtraction in `srtp_unprotect` folds into that one place. Non-RCC streams compute exactly what they did before.

The new test `rcc_mode1_rate4_mki_untagged` fails on main with `bad_mki` and passes here; `format.sh` is clean.

### DIFF
--- a/srtp/srtp.c
+++ b/srtp/srtp.c
@@ -1978,28 +1978,39 @@ static srtp_err_status_t srtp_get_session_keys_for_packet(
     return srtp_err_status_bad_mki;
 }
 
+/*
+ * Locating the MKI means knowing how many octets trail it, which RFC 4771
+ * makes depend on the packet: for AES-GCM only the 4-octet ROC of a mode 3
+ * ROC-carrying packet (RFC 7714 section 8.2), for a mode 1 packet that does
+ * not carry the ROC nothing at all (srtp_protect() appends no tag to those),
+ * otherwise the full authentication tag.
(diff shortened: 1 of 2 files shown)


pabuhler (approved):
@lenamonj , out of curiosity how was this discovered, through testing or AI code review?

lenamonj:
AI code review rather than testing. I build and run an automated review loop, so I am constantly pointing it at real repositories to evaluate and improve the method, and libsrtp came up because the RCC and MKI mechanics are exactly the kind of state dependent layout I have been working to make it better at. It flagged the key lookup stepping back a full tag length on a mode 1 packet that srtp_protect never tagged. I confirmed it, wrote the test that fails on main with bad_mki, and checked the derivation against RFC 4771 before filing.

